PR #37, opened 2022-02-14
Add meilisearch to the benchmarks

irevoire:
I ran the full test suite twice on a 13 inch macbook pro M1 2020


sunshowers:
Thanks for the contribution! Unfortunately, for benchmarks we'd want to use the same hardware, etc. Results across different operating systems and hardware can provide a general idea.

sunshowers:
I ended up running the tests on the same hardware/OS as the others: https://github.com/nextest-rs/nextest/commit/391f25d22ca78f2dcdfac1e9e0cf8aa478372230 so I've added a benchmark for this.

> Maybe indicating the OS/hardware used to run the benches (line by line) could solve this ? Benchmarks between cargo-test and cargo-nextest should always be run (for a single project) on the same machine so it would work even for external benchmarks

Good suggestion! I think we're good on benchmarks for now, but this would be potentially worth expanding on in the future.

PR #42, opened 2022-02-15
feat: print module path in cyan

bestgopher:
print module path in cyan instead of white.

image
image


sunshowers:
Thank you! I think this looks great.

Issue #57, opened 2022-02-17
`list` (maybe `run`?) doesn't respect `CARGO_TARGET_{TRIPLE}_RUNNER`


Jake-Shadle:
We build and test Windows binaries from Linux, and use [`CARGO_TARGET_{TRIPLE}_RUNNER`](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner) to specify (in our case) `wine` so that we can test the binaries, however nextest doesn't seem to read this configuration and attempts to directly execute the test binaries when performing the initial test list before running the tests.

I can add support for this tomorrow in a PR, just thought I would file this issue for visibility. Great work on this btw!

sunshowers:
To determine platform etc support, you'll probably want to just use target-spec, which depends on your own cfg-expr library!

sunshowers:
@Jake-Shadle I had a look at cargo's config and it feels like it'll be really hard to emulate all of it correctly. What do you think of adding this to nextest.toml instead?

sunshowers:
@Jake-Shadle this is now released as 0.9.6 -- could you confirm that https://nexte.st/book/target-runners.html makes sense to you? Thanks!

Jake-Shadle:
Thanks for the quick release! Yes, the docs make sense to me, I think the biggest thing is that people who are already using one of those methods for setting a target runner will be able to drop in nextest with no problems. :slightly_smiling_face:

PR #61, opened 2022-02-18
Add support for target runners

Jake-Shadle:
This adds support for [target runners](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner) which allows nextest to run tests for binaries that can't execute natively on the host such in cross-compilation situations targeting `x86_64-pc-windows-msvc` from a Linux host, or `wasm32-unknown-unknown`, etc.

This PR supports both the `CARGO_TARGET_{TRIPLE}_RUNNER` environment variable, as well as reading configured runners from `.cargo/config.toml`. This currently respects the same (IMO broken) [behavior](https://github.com/rust-lang/cargo/issues/2930) in cargo for looking up the hierarchy of configs.

Example output running windows binaries via wine from linux

image

Resolves: #57

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -470,6 +470,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "home"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2456aef2e6b6a9784192ae780c0f15bc57df0e918585282325e8c8ac27737654"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "humantime"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -633,6 +642,7 @@ dependencies = [
  "aho-corasick",
  "camino",
  "cargo_metadata",
+ "cfg-expr",
  "chrono",
  "color-eyre",
  "config",
@@ -641,6 +651,7 @@ dependencies = [
  "debug-ignore",
  "duct",
  "guppy",
+ "home",
  "humantime-serde",
  "indent_write",
  "indoc",
@@ -657,6 +668,7 @@ dependencies = [
  "serde",
  "serde_json",
  "strip-ansi-escapes",
+ "toml",
  "twox-hash",
 ]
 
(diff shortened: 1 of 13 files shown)


sunshowers (changes requested):
Thanks for doing this!

So in general I like the idea of supporting target runners, but I think trying to emulate Cargo's behavior is going to be really complicated, as this PR shows. Instead, I think we should support a `NEXTEST_<TRIPLE>_RUNNER` env var, and a config in `.config/nextest.toml` under the profiles section.

I also think this needs some tests to ensure that the correct runner's being run. An integration test could be done similar to the basic tests -- through a stub binary that just prints some output.

Jake-Shadle:
I started addressing your PR feedback but just to be sure...you're not saying to support a `NEXTEST_RUNNER` and nextest.toml configuration for the target runner _instead_ of cargo, but in addition to them?

Jake-Shadle:
So I'm looking at the config code right now, are you sure you want me to add runner support there? There's no target specific keys in there yet, and it seems like adding them to profiles could be really confusing since it would have to match both the target and the profile, and, at least how it is setup now, would mean you'd have to duplicate the runner for each profile...

sunshowers:
Thanks for workingo n this!

> I started addressing your PR feedback but just to be sure...you're not saying to support a NEXTEST_RUNNER and nextest.toml configuration for the target runner instead of cargo, but in addition to them?

I think it makes sense to support them instead of cargo, so that we don't have to deal with the complexity of emulating cargo's config logic. Nextest has several major differences from cargo test (such as per-test process isolation), and this can be documented as another.

> So I'm looking at the config code right now, are you sure you want me to add runner support there? There's no target specific keys in there yet, and it seems like adding them to profiles could be really confusing since it would have to match both the target and the profile, and, at least how it is setup now, would mean you'd have to duplicate the runner for each profile...

Most people I think will just set the default profile, which other profiles will inherit from. In general there aren't expected to be too many profiles -- the main use case I've seen is local vs CI builds.

sunshowers:
(So yeah I think `profile.<profile-name>.target.<target-name>.runner` would be the correct place for this.)

sunshowers:
OK, so after having thought about this more, I think the current solution is actually fine. I'll land this and iterate on it myself. Thank you so much for your patience dealing with this!!!

sunshowers (approved):

sunshowers:
The tests are absolutely *fantastic*, by the way. I love them so much.

Issue #44, opened 2022-02-15
Add support for running tests without building them


Guiguiprim:
This might be a weird request, but:

To optimize CI resources, improve build reproducibility and other things, I have the following setup:
- One CI step: cross compilation for windows from a linux CI runner (using nix-docker image)
- An other CI step: run the tests in a windows VM

Currently I build the tests with `cargo test --no-run` and then execute them without `cargo test` in the VMs (with an ugly script).

If it was possible to:
- `cargo nixtest --no-run` in the CI runner
- copy the `target` folder (or at least the relevant part) in the VM
- `cargo nixtest --no-build` in the VM

It would be really awesome.

As is, I don't know if it would be easily doable. If it is not to complex and an acceptable feature, I might be open in doing it under mentoring.

TriplEight:
Double this, it would be a good feature to share the compiled cache between i.e. 2 test executors.

poliorcetics:
> **The list phase.** `cargo-nextest` first builds all test binaries with `cargo test --no-run`, then queries those binaries to produce a list of all tests within them.

[source](https://nexte.st/book/how-it-works.html#the-nextest-model)

If running the binaries to get the test lisy is necessary, I don't think this will be easy

TriplEight:
It requires handling compile cache**s** carefully.

sunshowers:
Thanks for the feature request! I really like this idea. This is totally possible to do, though somewhat tricky. The main challenge here is how to communicate the list of binaries between the different machines.

The way it would probably have to be done is:
* Build and serialize the list of tests and binaries (the `list` [JSON serialization](https://nexte.st/book/machine-readable.html) would work for this)
* Add a `--test-list` option to `nextest run` that takes, as an argument, the serialized JSON representation. This would be an alternative means to compute the test list, and would be inserted [at this point](https://github.com/nextest-rs/nextest/blob/26356ae332504b8915ee4cca1f958b1e725e1c5d/cargo-nextest/src/dispatch.rs#L346-L347).
* For the situation where test binaries are at a different path, add a `--remap-path-prefix` option (dependent on `--test-list`) that replaces any matching prefixes in the test binaries with the specified one. This should be in the format `--remap-path-prefix FROM=TO`, matching rustc.

Some issues I can imagine you running into, and that I would like to see tests for:
[...]

PR #74, opened 2022-02-23
Adds support for separated build and run

Guiguiprim:
Implements https://github.com/nextest-rs/nextest/issues/44

This is still work in progress, but I just wanted to be sure I was on the right tracks from your point of view.

I have the basic feature working:
```sh
# builds and list binaries
> cargo nextest list-bins --message-format json > binaries.json
# runs without building
> cargo nextest run --test-list binaries.json
```

Still to do:
- [x] add path remapping
- [x] add tests

At the moment, I'm still creating the graph when running, I'm wondering if I should save it (or at least the needed part). This could allow me to use the runner without `cargo`  installed (which could be great for my use case, but not indispensable for the moment).

### DIFF
--- a/cargo-nextest/Cargo.toml
+++ b/cargo-nextest/Cargo.toml
@@ -29,3 +29,8 @@ owo-colors = { version = "3.2.0", features = ["supports-colors"] }
 shellwords = "1.1.0"
 supports-color = "1.3.0"
 serde_json = "1.0.79"
+
+[dev-dependencies]
+tempdir = "0.3.7"
+regex = "1.5.4"
+once_cell = "1.9.0"
--- a/cargo-nextest/Cargo.toml
+++ b/cargo-nextest/Cargo.toml
@@ -28,3 +28,4 @@ nextest-metadata = { version = "0.1.0", path = "../nextest-metadata" }
 owo-colors = { version = "3.2.0", features = ["supports-colors"] }
 shellwords = "1.1.0"
 supports-color = "1.3.0"
+serde_json = "1.0.79"
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -154,9 +154,9 @@ enum Command {
 }
 
 #[derive(Debug, Args)]
-#[clap(next_help_heading = "REUSE BUILD OPTIONS")]
+#[clap(next_help_heading = "REUSE BUILD OPTIONS (EXPERIMENTAL)")]
 struct ReuseBuildOpts {
-    /// Path to list of test binaries.
+    /// Path to list of test binaries
     #[clap(long, value_name = "PATH")]
     binaries_metadata: Option<Utf8PathBuf>,
 
@@ -394,6 +394,26 @@ impl TestReporterOpts {
     }
 }
 
+fn check_reuse_build_enabled_if_used(reuse_build: &ReuseBuildOpts) -> Result<()> {
+    let used = reuse_build.binaries_metadata.is_some()
+        || reuse_build.binaries_dir_remap.is_some()
+        || reuse_build.cargo_metadata.is_some()
+        || reuse_build.workspace_remap.is_some()
+        || reuse_build.platform_filter.is_some();
+
+    let var_name = "NEXTEST_EXPERIMENTAL_REUSE_BUILD";
+    let enabled = std::env::var(var_name).is_ok();
+
+    if used && !enabled {
+        Err(Report::new(ExpectedError::experimental_feature_error(
+            "build reuse",
+            var_name,
+        )))
+    } else {
+        Ok(())
+    }
+}
+
 impl AppOpts {
     /// Execute the command.
     fn exec(self) -> Result<()> {
@@ -404,6 +424,8 @@ impl AppOpts {
                 list_type,
                 reuse_build,
             } => {
+                check_reuse_build_enabled_if_used(&reuse_build)?;
+
                 let app = App::new(
                     self.output,
                     reuse_build,
@@ -421,6 +443,8 @@ impl AppOpts {
                 reporter_opts,
                 reuse_build,
             } => {
+                check_reuse_build_enabled_if_used(&reuse_build)?;
+
                 let app = App::new(
                     self.output,
                     reuse_build,
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -3,7 +3,7 @@
 
 use crate::{
     cargo_cli::{CargoCli, CargoOptions},
-    output::{OutputContext, OutputOpts},
+    output::{OutputContext, OutputOpts, OutputWriter},
     ExpectedError,
 };
 use camino::{Utf8Path, Utf8PathBuf};
@@ -28,7 +28,7 @@ use owo_colors::{OwoColorize, Style};
 use std::{
     error::Error,
     fmt::Write as _,
-    io::{BufWriter, Cursor, Write},
+    io::{Cursor, Write},
 };
 use supports_color::Stream;
 
@@ -45,9 +45,9 @@ pub struct CargoNextestApp {
 
 impl CargoNextestApp {
     /// Executes the app.
-    pub fn exec(self) -> Result<()> {
+    pub fn exec(self, output_writer: &mut OutputWriter) -> Result<()> {
         let NextestSubcommand::Nextest(app) = self.subcommand;
-        app.exec()
+        app.exec(output_writer)
     }
 }
 
@@ -416,7 +416,7 @@ fn check_reuse_build_enabled_if_used(reuse_build: &ReuseBuildOpts) -> Result<()>
 
 impl AppOpts {
     /// Execute the command.
-    fn exec(self) -> Result<()> {
+    fn exec(self, output_writer: &mut OutputWriter) -> Result<()> {
         match self.command {
             Command::List {
                 build_filter,
@@ -433,7 +433,7 @@ impl AppOpts {
                     self.config_opts,
                     self.manifest_path,
                 )?;
-                app.exec_list(message_format, list_type)
+                app.exec_list(message_format, list_type, output_writer)
             }
             Command::Run {
                 profile,
@@ -452,7 +452,13 @@ impl AppOpts {
                     self.config_opts,
                     self.manifest_path,
                 )?;
-                app.exec_run(profile.as_deref(), no_capture, &runner_opts, &reporter_opts)
+                app.exec_run(
+                    profile.as_deref(),
+                    no_capture,
+                    &runner_opts,
+                    &reporter_opts,
+                    output_writer,
+                )
             }
         }
     }
@@ -558,20 +564,23 @@ impl App {
         }
     }
 
-    fn exec_list(&self, message_format: MessageFormatOpts, list_type: ListType) -> Result<()> {
+    fn exec_list(
+        &self,
+        message_format: MessageFormatOpts,
+        list_type: ListType,
+        output_writer: &mut OutputWriter,
+    ) -> Result<()> {
         let binary_list = self.build_binary_list()?;
 
         match list_type {
             ListType::BinariesOnly => {
-                write_to_stdout(|writer| {
-                    binary_list
-                        .write(
-                            message_format.to_output_format(self.output.verbose),
-                            writer,
-                            self.output.color.should_colorize(Stream::Stdout),
-                        )
-                        .map_err(Into::into)
-                })?;
+                let mut writer = output_writer.stdout_writer();
+                binary_list.write(
+                    message_format.to_output_format(self.output.verbose),
+                    &mut writer,
+                    self.output.color.should_colorize(Stream::Stdout),
+                )?;
+                writer.flush()?;
             }
             ListType::Full => {
                 let target_runner = self.load_runner();
@@ -580,11 +589,12 @@ impl App {
                     test_list.colorize();
                 }
 
-                write_to_stdout(|writer| {
-                    test_list
-                        .write(message_format.to_output_format(self.output.verbose), writer)
-                        .map_err(Into::into)
-                })?;
+                let mut writer = output_writer.stdout_writer();
+                test_list.write(
+                    message_format.to_output_format(self.output.verbose),
+                    &mut writer,
+                )?;
+                writer.flush()?;
             }
         }
         Ok(())
@@ -596,6 +606,7 @@ impl App {
         no_capture: bool,
         runner_opts: &TestRunnerOpts,
         reporter_opts: &TestReporterOpts,
+        output_writer: &mut OutputWriter,
     ) -> Result<()> {
         let config = self
             .config_opts
@@ -622,8 +633,7 @@ impl App {
         }
 
         let runner = runner_builder.build(&test_list, &profile, handler);
-        let stderr = std::io::stderr();
-        let mut writer = BufWriter::new(stderr);
+        let mut writer = output_writer.stderr_writer();
         let run_stats = runner.try_execute(|event| {
             // Write and flush the event.
             reporter.report_event(event, &mut writer)?;
@@ -634,18 +644,6 @@ impl App {
         }
         Ok(())
     }
-}
-
-fn write_to_stdout<F: FnOnce(&mut BufWriter<std::io::StdoutLock>) -> Result<()>>(
-    write: F,
-) -> Result<()> {
-    let stdout = std::io::stdout();
-    let lock = stdout.lock();
-    // Buffer the output to minimize syscalls.
-    let mut writer = BufWriter::new(lock);
-    write(&mut writer)?;
-    writer.flush()?;
-    Ok(())
 }
 
 fn acquire_graph_data(manifest_path: Option<&Utf8Path>, output: OutputContext) -> Result<String> {
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -10,8 +10,9 @@ use camino::{Utf8Path, Utf8PathBuf};
 use clap::{ArgEnum, Args, Parser, Subcommand};
 use color_eyre::eyre::{Report, Result, WrapErr};
 use guppy::graph::PackageGraph;
+use nextest_metadata::{BinaryListSummary, Platform};
 use nextest_runner::{
-    config::NextestConfig,
+    config::{NextestConfig, NextestProfile},
     errors::{TargetRunnerError, WriteEventError},
     partition::PartitionerBuilder,
     reporter::{StatusLevel, TestOutputDisplay, TestReporterBuilder},
@@ -19,7 +20,9 @@ use nextest_runner::{
     signal::SignalHandler,
     target_runner::TargetRunner,
     test_filter::{RunIgnored, TestFilterBuilder},
-    test_list::{OutputFormat, RustTestArtifact, SerializableFormat, TestList},
+    test_list::{
+        BinaryList, OutputFormat, PathMapper, RustTestArtifact, SerializableFormat, TestList,
+    },
 };
 use owo_colors::{OwoColorize, Style};
 use std::{
@@ -51,12 +54,12 @@ impl CargoNextestApp {
 #[derive(Debug, Subcommand)]
 enum NextestSubcommand {
     /// A next-generation test runner for Rust. <https://nexte.st>
-    Nextest(AppImpl),
+    Nextest(AppOpts),
 }
 
 #[derive(Debug, Args)]
 #[clap(version)]
-struct AppImpl {
+struct AppOpts {
     /// Path to Cargo.toml
     #[clap(long, global = true, value_name = "PATH")]
     manifest_path: Option<Utf8PathBuf>,
@@ -108,6 +111,13 @@ enum Command {
             value_name = "FMT"
         )]
         message_format: MessageFormatOpts,
+
+        /// Type of listing
+        #[clap(long, arg_enum, default_value_t, value_name = "TYPE")]
+        list_type: ListType,
+
+        #[clap(flatten)]
+        reuse_build: ReuseBuildOpts,
     },
     /// Build and run tests
     ///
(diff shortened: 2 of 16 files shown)


cargo-nextest/src/dispatch.rs
```diff
@@ -83,6 +84,27 @@ impl ConfigOpts {
 
 #[derive(Debug, Subcommand)]
 enum Command {
+    /// List test binaries in workspace
+    ///
+    /// This command builds test binaries and list them.
+    /// Use --message-format json to get machine-readable output.
```
sunshowers, line 90:
```suggestion
    /// This command builds test binaries and lists them, without querying them for the tests they contain.
    /// Use --message-format json to get machine-readable output.
```

cargo-nextest/Cargo.toml
```diff
@@ -28,3 +28,4 @@ nextest-metadata = { version = "0.1.0", path = "../nextest-metadata" }
 owo-colors = { version = "3.2.0", features = ["supports-colors"] }
 shellwords = "1.1.0"
 supports-color = "1.3.0"
+serde_json = "1.0.79"
```
sunshowers, line 31: Please add a newline to the end of this file. (I need to add lint checks for all this...)

cargo-nextest/src/dispatch.rs
```diff
@@ -132,6 +154,11 @@ enum Command {
 
         #[clap(flatten)]
         reporter_opts: TestReporterOpts,
+
+        /// The list of Rust test binaries produced by `list-bins` in json.
+        /// Allow for splitting the tests build step from the tests run step.
+        #[clap(long)]
+        test_list: Option<Utf8PathBuf>,
```
sunshowers, line 161: Which heading does this show up in? I think it should just be in `OPTIONS`, I don't see a good place to put it in otherwise.

```suggestion
        bin_list: Option<Utf8PathBuf>,
```

nextest-metadata/src/test_list.rs
```diff
@@ -136,6 +133,28 @@ pub struct RustTestSuiteSummary {
...
+}
+
+/// A serializable suite of Rust test binaries.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct RustBinaryListSummary {
+    /// The list of test binaries
+    pub binaries: Vec<RustTestBinarySummary>,
+}
+
```
sunshowers, line 145: This should be a top-level `BinaryListSummary` with a `rust-binaries` key which contains a `BTreeMap<String, RustTestBinarySummary>`.

Regarding adding cargo metadata to the list, yeah I can see why that's useful! That's a bit hard to do at the moment because guppy doesn't expose (or even directly store) the metadata it was built from. What you could do is take the metadata at https://github.com/nextest-rs/nextest/blob/5fa8a77fec3310765d5a8260e5047ba61c09f4d9/cargo-nextest/src/dispatch.rs#L416-L417 and save it for later. Only parse it as JSON and include it in the final output if an option `--include-cargo-metadata` is passed in (and only accept `--include-cargo-metadata` if the output is serializable).

On the test running side, only accept `--bin-list` if the cargo metadata is available.

cargo-nextest/src/dispatch.rs
```diff
@@ -83,6 +84,27 @@ impl ConfigOpts {
 
 #[derive(Debug, Subcommand)]
 enum Command {
+    /// List test binaries in workspace
+    ///
+    /// This command builds test binaries and list them.
+    /// Use --message-format json to get machine-readable output.
+    ///
+    /// For more information, see <https://nexte.st/book/listing>.
+    ListBins {
```
sunshowers, line 93: Thinking about this more, I think maybe this should be part of `list`, as a `--no-query` (for "do not run/query binaries") option.

Also, `list` should accept a `--bin-list` parameter as well, just like the `run` command does.

sunshowers (commented):
Just gave it a quick look -- definitely the right direction! I have a few comments to the PR.

cargo-nextest/src/dispatch.rs
Guiguiprim: `list --bin-list <path>` ? I was going toward this at first (then I read you comment on the ticket again), so OK for me.

nextest-metadata/src/test_list.rs
Guiguiprim: I did think of keeping the raw metadata output, but wasn't sure about it... I will test it that way.

`rust_binaries` indexed by binary name ?

cargo-nextest/src/dispatch.rs
sunshowers: Yeah I think that makes sense because you may want to list out all the tests given a list of binaries.

However it occurs to me that maybe we shouldn't call the parameter --bin-list then to avoid confusion.  Maybe something like --binary-metadata? Happy to hear your suggestions too.

cargo-nextest/src/dispatch.rs
Guiguiprim: changed see new comment

nextest-metadata/src/test_list.rs
Guiguiprim: Done

cargo-nextest/src/dispatch.rs
Guiguiprim: Did a bit differently (see new comment).

Guiguiprim:
I changed some things, now we can:

```sh
# calls  `cargo metadata`, builds tests, saves binaries metadata, lists binaries
cargo nextest list --no-query --output-binaries-metadata binaries-metadata.json
# does not build, calls  `cargo metadata`, queries tests and list them
cargo nextest list --binaries-metadata binaries-metadata.json
# does not build, calls  `cargo metadata`, run tests
cargo nextest run --binaries-metadata binaries-metadata.json
```

```sh
# calls  `cargo metadata`, builds tests, saves binaries and cargo metadata, lists binaries
cargo nextest list --no-query --output-cargo-metadata cargo-metadata.json --output-binaries-metadata binaries-metadata.json
# does not build, does not call `cargo metadata`, queries tests and list them
cargo nextest list --cargo-metadata cargo-metadata.json --binaries-metadata binaries-metadata.json
# does not build, does not call `cargo metadata, run tests
cargo nextest run --cargo-metadata cargo-metadata.json --binaries-metadata binaries-metadata.json
```

Still needs to add path remapping and tests.

Guiguiprim:
A new deprecation warning is breaking the build with rust 1.59 apparently.

sunshowers:
> A new deprecation warning is breaking the build with rust 1.59 apparently.

Thanks for flagging, I'll fix this up!

sunshowers:
@Guiguiprim just pushed a fix, please rebase! thanks :)

sunshowers:
Just pushed another couple of fixes, hope these are enough

Guiguiprim:
Fun fact `cargo test --no-run` compiles both proc-macro test for the host target and tests for the given target when cross-compiling, so I had to add a filtering for this. But this actually gives us a nice feature that was missing in my current setup: running the proc-macros tests.

So at this point what I have is:

```sh
# On cross build machine
# Build all tests
cargo nextest list --no-query --output-cargo-metadata cargo-metadata.json \
 --output-binaries-metadata binaries-metadata.json
# List host tests (proc-macros ones)
cargo nextest list --cross-compile-filter host-only --cargo-metadata cargo-metadata.json \
 --binaries-metadata binaries-metadata.json
# Run host tests (proc-macros ones)
cargo nextest run --cross-compile-filter host-only --cargo-metadata cargo-metadata.json \
 --binaries-metadata binaries-metadata.json

# Copy tests and metadata in target machine 

# On target machine
# List tests (excluding proc-macros ones)
cargo nextest list --cross-compile-filter target-only --cargo-metadata cargo-metadata.json \
 --binaries-metadata binaries-metadata.json --workspace-remp c:/my-repo \
 --binaries-dir-remap c:/my-repo/tests-artifacts
# List tests (excluding proc-macros ones)
cargo nextest run --cross-compile-filter target-only --cargo-metadata cargo-metadata.json  \
 --binaries-metadata binaries-metadata.json --workspace-remp c:/my-repo \
 --binaries-dir-remap c:/my-repo/tests-artifacts
```

Just missing tests and I'm done.

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +156,69 @@ enum Command {
...
         reporter_opts: TestReporterOpts,
+
+        #[clap(flatten)]
+        no_build: NoBuildOptions,
     },
 }
 
+#[derive(Debug, Args)]
+#[clap(next_help_heading = "NO BUILD OPTIONS")]
+struct NoBuildOptions {
```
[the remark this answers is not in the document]
sunshowers: Also could this be called something like `ReuseBuildOptions`?

cargo-nextest/src/dispatch.rs
```diff
@@ -316,38 +414,71 @@ impl AppImpl {
...
+                if no_query {
+                    write_to_stdout(|writer| {
+                        binary_list
+                            .write(
+                                message_format.to_output_format(output.verbose),
+                                writer,
+                                output.color.should_colorize(Stream::Stdout),
+                            )
+                            .map_err(Into::into)
+                    })?;
```
sunshowers, line 453: It's not entirely clear to me what's happening here -- it seems like the binaries metadata might get written out twice, once to stdout and once to what's specified in `output-binaries-metadata`? This seems somewhat incorrect to me. I think `output_binaries_metadata` should just not exist.

Also, why add `output_cargo_metadata` when users can just run `cargo metadata` anyway? Some scripting around this is going to be required in any case -- seems simpler to just have consumers run `cargo metadata --format-version 1` themselves.

nextest-runner/src/test_list.rs
```diff
@@ -50,77 +53,93 @@ pub struct RustTestArtifact<'g> {
...
+    /// The package this artifact belongs to.
+    pub package_id: String,
+    /// The unique binary name defined in `Cargo.toml` or inferred by the filename.
+    pub name: String,
+    /// A unique ID.
+    pub id: String,
+
+    /// Binary build for the host target during cross compilation.
+    /// This include proc-macro tests.
+    pub build_host_only: bool,
```
sunshowers, line 69: Can this be something like `platform` with an enum with options `host` and `target`?

sunshowers (commented):
Thanks for doing this! I have a bunch of comments but overall this is looking good -- though it might take a couple more rounds of review.

So one of the things I'm concerned about is some of the other environmental state used in nextest. Here are a few examples:
1. The cwd is passed down to cargo for the build. This isn't an issue because here, building and running tests are on separate machines.
2. `.nextest/config.toml` from the workspace root. I think this should be remapped at least.
3. Target runner support. This one I'm a little more concerned about because cargo itself relies on the current working directory. I think the most sensible thing to do here would be to start the search at the workspace root.

Overall, I think it is important to document exactly what our expectations are.
* Do we expect a source checkout to be available? (I think yes for now, which simplifies point 2 above. I think turning this into "no" is going to require some more design work.)
* Do we expect `cargo` to be available? (I think probably no?)

One more thing I'd like to see is some website documentation -- a clear set of instructions which say "here's how to use this" because there's some nontrivial complexity here.

cargo-nextest/src/dispatch.rs
Guiguiprim: Indeed.

I did get lost / trapped in my design reflection without your previous comment about `--include-cargo-metadata` and wasn't sure about exposing the use to `cargo metadata` directly (they must use the right arguments).

But indeed it makes more sense to let them call it.

I removed both `--output-*` parameters, and will work on a proper documentation for the feature.

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +156,69 @@ enum Command {
...
+    /// If given the test binaries will be used without being rebuild.
+    /// This can be generated with `cargo nextest list --no-query --output-binaries-metadata binaries.json`.
+    #[clap(long, value_name = "PATH")]
+    binaries_metadata: Option<Utf8PathBuf>,
+
+    /// A path to cargo metadata in json.
+    /// If given the test binaries will be used without being rebuild.
+    /// This can be generated with `cargo nextest list --no-query --output-cargo-metadata cargo.json`.
+    #[clap(long, requires("binaries-metadata"), value_name = "PATH")]
+    cargo_metadata: Option<Utf8PathBuf>,
```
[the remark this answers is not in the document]
Guiguiprim: Done.

Yes, added `conflicts_with()`

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +156,69 @@ enum Command {
...
+
+    /// Remapping for the workspace root.
+    #[clap(long, requires("binaries-metadata"), value_name = "PATH")]
+    workspace_remap: Option<Utf8PathBuf>,
```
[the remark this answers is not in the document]
Guiguiprim: To `cargo_metadata`, fixed.

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +156,69 @@ enum Command {
...
+    binaries_dir_remap: Option<Utf8PathBuf>,
+
+    /// Filter binaries for one target when cross compiling
+    ///
+    /// Allows:
+    /// - building tests on cross-build machine
+    /// - runnings host tests on cross-build machine (proc-macro tests)
+    /// - runnings tests on target machine
+    #[clap(long, arg_enum, default_value_t, value_name = "FILTER")]
+    cross_compile_filter: CrossCompileFilterOpts,
```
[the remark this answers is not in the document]
Guiguiprim: Done

cargo-nextest/src/dispatch.rs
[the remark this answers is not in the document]
Guiguiprim: Renamed

cargo-nextest/src/dispatch.rs
```diff
@@ -316,38 +414,71 @@ impl AppImpl {
...
-        let graph = build_graph(self.manifest_path.as_deref(), output)?;
-
         match self.command {
             Command::List {
                 build_filter,
                 message_format,
+                no_query,
+                output_binaries_metadata,
+                output_cargo_metadata,
+                ref no_build,
```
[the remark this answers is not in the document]
Guiguiprim: Reworked

nextest-runner/src/test_list.rs
Guiguiprim: Done

nextest-runner/src/test_list.rs
```diff
@@ -166,6 +333,54 @@ pub struct RustTestSuite<'g> {
     pub testcases: BTreeMap<String, RustTestCaseSummary>,
 }
 
+/// An helper for path remapping
+///
+/// This is useful when running the test elsewhere from building them.
```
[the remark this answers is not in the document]
Guiguiprim: Done

nextest-runner/src/test_list.rs
```diff
@@ -166,6 +333,54 @@ pub struct RustTestSuite<'g> {
...
+    pub fn new(
+        graph: &PackageGraph,
+        workspace: Option<Utf8PathBuf>,
+        binaries_dir: Option<Utf8PathBuf>,
+    ) -> Option<Self> {
+        if workspace.is_none() && binaries_dir.is_none() {
+            return None;
+        }
+
+        // canonicalize workspace and binaries_dir ?
```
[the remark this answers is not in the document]
Guiguiprim: OK, comment removed

sunshowers:
Is this ready for another round of review?

Guiguiprim:
Yes it is.

- I reworked the structure a bit in `dispatch` to be sure to use the right `manifest_path` and `workspace_root` (using loaded cargo metadata and applying remapping if any).
- For `.nextest/config.toml`, I didn't do anything, if `cargo nextest` is invoked from within the repository clone on the target machine, it should work the same way I believe.
- I did checked that we used the right runner, in particular when using `--platform-filter host` not to use the target runner. But again, if it is invoked from the repository clone I believe that using cwd for config search should still work.
- I added documentation in `site/src/book/reusing-build.md`

nextest-runner/src/test_list.rs
```diff
@@ -50,77 +53,97 @@ pub struct RustTestArtifact<'g> {
...
+/// A Rust test binary built by Cargo.
+pub struct RustTestBinary {
+    /// The path to the binary artifact.
+    pub path: Utf8PathBuf,
+    /// The package this artifact belongs to.
+    pub package_id: String,
+    /// The unique binary name defined in `Cargo.toml` or inferred by the filename.
+    pub name: String,
+    /// A unique ID.
+    pub id: String,
```
sunshowers, line 65: Could you make `id` the first element of the struct?

nextest-metadata/src/test_list.rs
```diff
@@ -137,6 +144,35 @@ pub struct RustTestSuiteSummary {
     /// The path to the test binary executable.
     pub binary_path: Utf8PathBuf,
 
+    /// A unique binary ID.
+    pub binary_id: String,
```
sunshowers, line 148: Could you make this the first item in the struct?

nextest-metadata/src/test_list.rs
```diff
@@ -137,6 +144,35 @@ pub struct RustTestSuiteSummary {
...
+
+    /// Platform for which this binary was build.
+    /// (Proc-macro tests are build for the host.)
```
sunshowers, line 151:
```suggestion
    /// Platform for which this binary was built.
    /// (Proc-macro tests are built for the host.)
```

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +147,51 @@ enum Command {
...
+
+    /// Path to cargo metadata JSON.
+    #[clap(long, conflicts_with("manifest-path"), value_name = "PATH")]
+    cargo_metadata: Option<Utf8PathBuf>,
+
+    /// Remapping for the workspace root.
+    #[clap(long, requires("cargo-metadata"), value_name = "PATH")]
+    workspace_remap: Option<Utf8PathBuf>,
+
+    /// Filter binaries based on the platform they were build for.
```
sunshowers, line 175:
```suggestion
    /// Filter binaries based on the platform they were built for
```

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +147,51 @@ enum Command {
...
+
+    /// Remapping for the test binaries directory.
+    #[clap(long, requires("binaries-metadata"), value_name = "PATH")]
+    binaries_dir_remap: Option<Utf8PathBuf>,
+
+    /// Path to cargo metadata JSON.
+    #[clap(long, conflicts_with("manifest-path"), value_name = "PATH")]
+    cargo_metadata: Option<Utf8PathBuf>,
+
+    /// Remapping for the workspace root.
```
sunshowers, line 171: No period at the end of comments -- please keep to the standard

cargo-nextest/src/dispatch.rs
```diff
@@ -311,92 +382,234 @@ impl TestReporterOpts {
...
+            Some(path) => std::fs::read_to_string(path)?,
+            None => acquire_graph_data(manifest_path.as_deref(), output)?,
+        };
+        let graph = guppy::CargoMetadata::parse_json(&graph_data)?.build_graph()?;
+
+        let manifest_path = if reuse_build.cargo_metadata.is_some() {
+            Some(graph.workspace().root().join("Cargo.toml"))
+        } else {
+            manifest_path
+        };
```
sunshowers, line 455: This makes sense!

sunshowers (commented):
I think this is looking really good! Thanks for the code reorganization, it makes much more sense now.

Just a few more general comments:
1. I'm quite worried about baking in this behavior as stable immediately -- in particular I'm worried about how this will interact with issues like #82. I'm tempted to say that we should gate this behind a `NEXTEST_EXPERIMENTAL_REUSE_BUILD` env var for now. I think that should work for you since you can just set the env var, but it'll give us the flexibility to change the CLI in the future.
2. This really still needs more tests. I'd like to see tests for (a) listing out binaries without running them (this can become stable) and (b) splitting up the build and run phases like you've done here.

sunshowers:
Have you been using this, by the way?

I think we should also write a CI job on the nextest repo itself, that does an end-to-end test of this behavior. (Including running tests on the same arch, same arch with relocation, and a test on a different arch if possible--this last one may be extra difficult so it's ok to skip for now.

Guiguiprim:
- I added a gate under `NEXTEST_EXPERIMENTAL_REUSE_BUILD`, let me know if it is what you had in mind.
- I added some tests for `BinaryList` construction 
- I'm trying to add CI tests of the basic feature (having never use github-actions)

I haven't integrated it yet in my CI, but I made myself a cross-testing script:
- Builds the tests on my linux host
- Runs the proc-macro tests on my host
- Copy the repository and build artifacts inside my local Windows VM
- Runs the test in my VM (using only cargo-nextest, without cargo installed)
It does run smoothly for me.

Guiguiprim:
The failling tests on MacOS are due to the addition of a proc-macro lib in the tests input workspace and can be seen in master (https://github.com/nextest-rs/nextest/pull/84), I'm going to try to investigate but any help is welcome.

cargo-nextest/src/dispatch.rs
```diff
@@ -137,9 +147,51 @@ enum Command {
...
         #[clap(flatten)]
         reporter_opts: TestReporterOpts,
+
+        #[clap(flatten)]
+        reuse_build: ReuseBuildOpts,
     },
 }
 
+#[derive(Debug, Args)]
+#[clap(next_help_heading = "EXPERIMENTAL: REUSE BUILD OPTIONS")]
```
sunshowers, line 157: Could you change this to `REUSE BUILD OPTIONS (EXPERIMENTAL)`?

cargo-nextest/src/dispatch.rs
```diff
@@ -108,6 +111,13 @@ enum Command {
             value_name = "FMT"
         )]
         message_format: MessageFormatOpts,
+
+        /// Only list the test binaries, without querying them for the tests they contain.
+        #[clap(long)]
+        no_query: bool,
```
sunshowers, line 117: I've been thinking about this config option a bit and I think I'd be happier with it as a `--list-type` option, because the JSON output changes quite significantly based on the type.
* `full` (default; run binaries to query and list tests)
* `binaries-only` (do not run binaries)

I think a `--no-query` option doesn't quite explain this fully, but having `type` in the name of the option makes it clearer.

site/src/SUMMARY.md
```diff
@@ -17,6 +17,7 @@
   - [Other options](book/other-options.md)
   - [Environment variables](book/env-vars.md)
   - [Stability policy](book/stability.md)
+  - [Reusing build](book/reusing-build.md)
```
sunshowers, line 20: Could you remove this line for now? We'll add it after the new version of nextest is released.

sunshowers (commented):
This looks really good, just a couple of minor issues left! Definitely a lot more comfortable with this being behind an experimental env var. Once you rebase I'm happy to land this and iterate on it a little, then get this out there!

Thanks again for all your work.

sunshowers:
Let me know when it's ready for review! I'll try getting to it within a day of that.

Guiguiprim:
If this last run is green (it should be), I'm done.

As you probably saw, I struggled adding a complete suite of integration tests. (see the comment at the begin of `cargo-nextest/src/tests_integration/mod.rs` and the comment at the end of this file).

Guiguiprim:
Fun fact:

`BtreeMap<camino::Utf8PathBuf, _>` is not sorted the same way between rust `1.58` and `1.59` (I had two entries inverted).

Guiguiprim:
It's green !

The ball is yours.

Have a great day, I'm going home !

Guiguiprim:
Initially I was calling `cargo-nextest` binary from the integration tests, but due to issues on Windows, I moved to calling `args.exec()` from within the lib tests.

Guiguiprim:
I just rebased

sunshowers:
Apologies, been caught up with other stuff the last couple of days. I'm going to endeavor to get this out there Sunday or Monday.

Issue #73, opened 2022-02-23
In .cargo/config `runner` can be an array of string


Guiguiprim:
Recently support for custom runner was added, but it fails to parse some valid config file.

We currently assume that the runner parameter is a simple string, but it can be an array of string.

 see [target.runner](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner): `Type: string or array of strings ([program path and args])`

sunshowers:
Whoops! Going to try and fix this now.

Guiguiprim:
@sunshowers I just opened a PR for this !

sunshowers:
Addressed in #75 and version 0.9.8. Thanks!

PR #75, opened 2022-02-23
Improve runner parsing

Guiguiprim:
Fix for https://github.com/nextest-rs/nextest/issues/73

I'm not sure about the second commit. Maybe we should stop the ancestors walk at the workspace root ?
What was the goal behind `use_cargo_home` ? tests ? If so, I guess both approach would work.
Fun fact: the tests where broken before the first commit on my computer because they were reading my home config with an array of strings (even with the `use_cargo_home = false`)...

### DIFF
--- a/nextest-runner/src/target_runner.rs
+++ b/nextest-runner/src/target_runner.rs
@@ -4,7 +4,7 @@
 //! Adds support for [target runners](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner)
 
 use crate::errors::TargetRunnerError;
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use target_spec::Platform;
 
 #[derive(serde::Deserialize, Debug)]
@@ -28,8 +28,11 @@ impl TargetRunner {
     /// Acquires the [target runner](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner)
     /// which can be set in a [.cargo/config.toml](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure)
     /// or via a `CARGO_TARGET_{TRIPLE}_RUNNER` environment variable
-    pub fn for_target(target_triple: Option<&str>) -> Result<Option<Self>, TargetRunnerError> {
-        Self::get_runner_by_precedence(target_triple, true, None)
+    pub fn for_target(
+        target_triple: Option<&str>,
+        terminate_search_at: Option<&Utf8Path>,
+    ) -> Result<Option<Self>, TargetRunnerError> {
+        Self::get_runner_by_precedence(target_triple, None, terminate_search_at)
     }
 
     /// Configures the root directory that starts the search for cargo configs.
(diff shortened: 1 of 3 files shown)


nextest-runner/src/target_runner.rs
```diff
@@ -273,3 +303,116 @@ impl TargetRunner {
...
+            Ok(vec![s.to_owned()])
+        }
+
+        fn visit_seq<S>(self, seq: S) -> Result<Self::Value, S::Error>
+        where
+            S: SeqAccess<'de>,
+        {
+            Deserialize::deserialize(value::SeqAccessDeserializer::new(seq))
+        }
+    }
```
sunshowers, line 337: Thanks! I think this should return an enum rather than a single string (which I believe cargo will interpret as a unit rather than splitting it up).

sunshowers (commented):
Looks good, just one change required.

nextest-runner/src/target_runner.rs
```diff
@@ -168,11 +168,25 @@ impl TargetRunner {
...
+            let mut cargo_home_path = home::cargo_home_with_cwd(root.as_std_path())
+                .ok()
+                .and_then(|home| Utf8PathBuf::from_path_buf(home).ok());
+            if let Some(mut cargo_home_path) = cargo_home_path.as_mut() {
+                if let Some(home_config) = read_config_dir(&mut cargo_home_path) {
+                    if let Some(p) = configs.iter().position(|path| path == &home_config) {
+                        configs.remove(p);
+                    }
+                }
+            }
```
sunshowers, line 183: Honestly looking at this, I think the correct semantics for tests are something like "isolate it to the project root and don't iterate further up".

sunshowers:
I'm going to get a quick hotfix out to make nextest non-fatal in case of any issues (which I believe is the correct thing to do anyway), then let's figure out a good fix here.

nextest-runner/src/target_runner.rs
Guiguiprim: Indeed, and it is way easier actually.

nextest-runner/src/target_runner.rs
Guiguiprim: I don't really know how to do this, I don't think there is an easy way of finding the workspace root (from cargo env-vars at least).

Guiguiprim: We do have it in the `PackageGraph`.

nextest-runner/src/target_runner.rs
```diff
@@ -7,6 +7,13 @@ use crate::errors::TargetRunnerError;
 use camino::Utf8PathBuf;
 use target_spec::Platform;
 
+#[derive(serde::Deserialize, Debug)]
+#[serde(untagged)]
```
sunshowers, line 11: Could you add an `expecting` here?

nextest-runner/src/target_runner.rs
sunshowers: Hmm, we do in https://github.com/nextest-rs/nextest/blob/main/nextest-runner/tests/integration/fixtures.rs#L89-L106 don't we?

sunshowers: I'd just pass in something like `terminate_search_at: Option<&Utf8Path>`.

sunshowers: Oh I see, so there's two separate things here:
* Terminate search at the workspace root.
* Don't read cargo_home.

Given that the use cases for these two are linked to each other (isolation for tests) I think it makes sense to combine them into one.

Guiguiprim: done

nextest-runner/src/target_runner.rs
```diff
@@ -148,6 +161,10 @@ impl TargetRunner {
                 configs.push(config);
             }
 
+            if dir == workspace_root {
+                break;
+            }
```
sunshowers, line 166: Hmm, I don't think this is quite correct -- by default, cargo does search past the workspace root: https://doc.rust-lang.org/cargo/reference/config.html. So I think there should be two modes:
1. default mode, which emulates cargo
2. isolation mode, which stops at the workspace root and doesn't look at the home directory.

nextest-runner/src/target_runner.rs
Guiguiprim: Not sure what you mean.

sunshowers: Oh, I thought serde macros took an "expecting" atrribute. Doesn't look like they do. Never mind!

sunshowers:
I'll have a final look at this and land it later today. Thanks!

sunshowers:
Thanks! I have some minor plans for improvement but I'll land this and iterate on it myself.

PR #76, opened 2022-02-23
Add artifact kind to binary_id if it's not 'lib'

pmsanford:
In a crate with multiple targets that have the same name as the package, if both have tests, tests from one of the targets will be missing from the JSON output. I ran into this is in a lib + bin package. You can see a minimal example [here](https://github.com/pmsanford/nextest-json-suite-overwrite).

This is because of the way `RustTestArtifact.binary_id` is constructed. Currently, if the target name is equal to the package name, the package name is taken as the `binary_id`. There are cases (such as lib + bin) where two targets will have the same name as the package name. This results in `binary_id` being non-unique in those cases, and the doc comment for `binary_id` says it should be unique. Currently this only seems to affect the JSON output of the list command - the human readable output of list is still correct and all the tests are still run.

The approach this PR takes is, in cases where the target name is equal to the package name, check the target type. If it's anything other than lib, append the type to the name to build the binary ID. I also considered fixing this at the point where the summaries are built for serialization, but it seems like `binary_id` is truly intended to uniquely identify a test suite, so I thought this was a more appropriate place.

### DIFF
--- a/nextest-runner/src/test_list.rs
+++ b/nextest-runner/src/test_list.rs
@@ -68,7 +68,6 @@ impl<'g> RustTestArtifact<'g> {
                         let package = graph
                             .metadata(&package_id)
                             .map_err(FromMessagesError::PackageGraph)?;
-
                         // Tests are run in the directory containing Cargo.toml
                         let cwd = package
                             .manifest_path()
@@ -86,6 +85,11 @@ impl<'g> RustTestArtifact<'g> {
                         if artifact.target.name != package.name() {
                             binary_id.push_str("::");
                             binary_id.push_str(&artifact.target.name);
+                        } else if !artifact.target.kind.contains(&"lib".to_owned()) {
+                            if let Some(kind) = artifact.target.kind.get(0) {
+                                binary_id.push_str("::");
+                                binary_id.push_str(kind);
+                            }
                         }
 
                         binaries.push(RustTestArtifact {
--- a/nextest-runner/src/test_list.rs
+++ b/nextest-runner/src/test_list.rs
@@ -82,13 +82,36 @@ impl<'g> RustTestArtifact<'g> {
 
                         // Construct the binary ID from the package and build target.
                         let mut binary_id = package.name().to_owned();
-                        if artifact.target.name != package.name() {
+                        let binary_name = artifact.target.name;
+
+                        // To ensure unique binary IDs, we use the following scheme:
+                        // 1. If the target is a lib, use the package name.
+                        //      There can only be one lib per package, so this
+                        //      will always be unique.
+                        if !artifact.target.kind.contains(&"lib".to_owned()) {
                             binary_id.push_str("::");
-                            binary_id.push_str(&artifact.target.name);
-                        } else if !artifact.target.kind.contains(&"lib".to_owned()) {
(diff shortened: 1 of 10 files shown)


nextest-runner/src/test_list.rs
```diff
@@ -86,6 +85,11 @@ impl<'g> RustTestArtifact<'g> {
                         if artifact.target.name != package.name() {
                             binary_id.push_str("::");
                             binary_id.push_str(&artifact.target.name);
+                        } else if !artifact.target.kind.contains(&"lib".to_owned()) {
+                            if let Some(kind) = artifact.target.kind.get(0) {
```
pmsanford, line 89: Based on https://doc.rust-lang.org/cargo/commands/cargo-metadata.html it seems like kind will only ever have one item in it unless it's a lib. I don't have any experience working with cargo metadata, though, so I'm not sure this is the right approach.

This also means if there's nothing in the kind vec, it will silently use the package name for the binary_id - should this be an error instead?

sunshowers: Nothing in the kind vec is definitely weird and should at least be a warning I think.

nextest-runner/src/test_list.rs
```diff
@@ -86,6 +85,11 @@ impl<'g> RustTestArtifact<'g> {
                         if artifact.target.name != package.name() {
                             binary_id.push_str("::");
                             binary_id.push_str(&artifact.target.name);
```
sunshowers, line 87: Hmm, what happens if there's multiple `bin` targets within a crate, each with their own tests?

I think the best scheme for this is probably something like:
`<crate-name>` for the lib crate (unit tests)
`<crate-name>::<bin-name>` for test targets
`<crate-name>::<bin/example>/<bin-name>` for other targets, e.g in the included test, `nextest-tests::bin/nextest-tests`

What do you think? Happy to hear suggestions as well.

(And could you add more tests/examples around this? Thanks!)

sunshowers (commented):
Thank you for catching this and for the PR with the excellent description!

nextest-runner/src/test_list.rs
pmsanford: Aha, you're right - the problem is more general than I thought. I think your naming scheme makes sense - I'll give it a go.

pmsanford: So I think this works, with one small exception: if the name of the lib is different from the package, it's not reflected in the binary_id now. It's still unique and un-ambiguous, but we could also use
`<crate-name>::lib/<lib-name>` in that case if you think it'd be confusing not seeing it.

sunshowers:
Thank you! Will let a test run finish and then put it out there

sunshowers:
Could you update the PR for the test failures? We do support Rust 1.54 which doesn't support f-strings.

pmsanford:
Woops! Caught that in one file but missed it in the error impl

nextest-runner/src/test_list.rs
sunshowers: > if the name of the lib is different from the package, it's not reflected in the binary_id now.

Good catch, I keep forgetting about this use case. I wouldn't address this personally.

sunshowers:
Thanks again!

PR #84, opened 2022-03-02
Add proc-macro lib in tests fixtures

Guiguiprim:
Adding a proc-macro lib in the workspace used as tests input makes `nextest-runner::integration target_runner::run::test_run_with_target_runner` fail on MacOS.

I'm not yet sure why.

### DIFF
--- a/fixtures/nextest-tests/Cargo.toml
+++ b/fixtures/nextest-tests/Cargo.toml
@@ -23,5 +23,8 @@ test = true
 
 # Make this crate its own workspace.
 [workspace]
+members = [
+    "derive"
+]
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


sunshowers:
Looks like an issue with the dynamic linker -- going to dig into the environment variables that get passed in by regular cargo test.

sunshowers:
So the weird thing is that I can't actually reproduce this locally on a Mac -- there's some sort of difference between the CI and local environments that's causing this. Debugging this is going to be a bit of a pain!

sunshowers:
OK, so this is a macOS System Integrity Protection issue: https://apple.stackexchange.com/questions/212945/unable-to-set-dyld-fallback-library-path-in-shell-on-osx-10-11-1

Going to see how I can work around this.

sunshowers:
#87 should fix this.

sunshowers:
OK, if you rebase now it should work. Thanks!

PR #86, opened 2022-03-02
[cargo-nextest] remove allow attribute

ymgyt:
`CargoCli::all_args()` is used in `TestBuildFilter::compute()`
https://github.com/nextest-rs/nextest/blob/4d8c9200bc3415b38debb2084ce6c4a45bf1d18a/cargo-nextest/src/dispatch.rs#L210


sunshowers:
Thank you for the contribution! Will land as soon as tests pass.

Issue #21, opened 2022-02-10
Support reading configuration as environment variables


sunshowers:
Should be slotted in between CLI options and configuration in files, and should be prefixed by `NEXTEST_`.

sunshowers:
This is part of nextest 0.9.12: https://nexte.st/book/env-vars.html#environment-variables-nextest-reads

PR #113, opened 2022-03-18
[cargo-nextest] read configuration as environment variables

ymgyt:
Enable clap env feature to use env attribute api.
CargoOption has not changed.

https://github.com/nextest-rs/nextest/issues/21


sunshowers:
Looks good, thank you! I'll add some documentation for this before releasing it.

Issue #56, opened 2022-02-17
Compatibility with coverage tools?


pwoolcoc:
Is there any way to use this in conjunction with coverage tools? Maybe it would be better to go see if the coverage tool has support for this rather than the other way around, but if someone here has gotten nextest working with a coverage tool I'd love to hear about it.

sunshowers:
I'm not aware of any current efforts, but I'd certainly welcome some discussion about it. My ideal solution would involve not baking the specifics of a particular coverage tool into nextest itself, and having some sort of inter-process protocol to communicate with a daemon as necessary instead.

pwoolcoc:
turns out it works just fine with `rustc`'s built-in instrumentation and llvm's coverage tools. Repo is here: https://github.com/pwoolcoc/nextest-cov-example, output looks like:

<img width="1626" alt="image" src="https://user-images.githubusercontent.com/620392/154739713-c333eb67-1256-4ba7-a80f-6ee45cdd40c3.png">

skyzh:
I'm working on integrating nextest with llvm-cov. https://github.com/taiki-e/cargo-llvm-cov/pull/144

If the above PR gets merged, generating coverage with nextest should be as simple as `cargo llvm-cov nextest`

sunshowers:
@skyzh that's awesome! thank you

sunshowers:
> maybe we can update docs and close this issue?

Thanks again for getting this upstream! llvm-cov is one of several coverage tools one might want to use, so let's try and write up some generic documentation for this as well before closing this issue.

sunshowers:
https://github.com/nextest-rs/nextest/pull/119 has more documentation.

sunshowers:
https://nexte.st/book/test-coverage.html has the documentation!

PR #115, opened 2022-03-18
docs: add coverage support

skyzh:

close https://github.com/nextest-rs/nextest/issues/56

PTAL, thanks! cc @sunshowers

### DIFF
--- a/site/src/SUMMARY.md
+++ b/site/src/SUMMARY.md
@@ -16,6 +16,7 @@
   - [Target runners](book/target-runners.md)
   - [Other options](book/other-options.md)
   - [Environment variables](book/env-vars.md)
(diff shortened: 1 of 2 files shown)


skyzh:
We can also wait for llvm-cov to release a stable version, so that we don't need to install from git...

sunshowers:
Thank you for getting this done upstream!

Looks like Taiki added nextest to his install action -- would like to work with him on a small fixup and then update docs to point to that.

sunshowers:
Let's get https://github.com/taiki-e/install-action/pull/3 in then land this! Thanks again.

sunshowers:
Going to land this then edit it a bit. Thanks again!

PR #116, opened 2022-03-18
Make CargoOptions conflict with ReuseBuildOpts

Guiguiprim:
`binaries-metadata` is the main entry point of ReuseBuildOpts, so using it as the conflicting argument.

I hopped to be able to have less duplication, but didn't find a good way of doing it. I opened a discussion about this: https://github.com/clap-rs/clap/discussions/3563


sunshowers:
Thank you! I think this will do.

PR #117, opened 2022-03-18
Add expression based test filtering

Guiguiprim:
As you are up apparently, I push this to open the discussion.

Following what you said, I was thinking maybe something along the line of:

DSL
 - `package(name)`
 - `deps(package_name)`
 - `rdeps(package_name)`
 - `test(partial_name)`
 - `not(expr)`
 - `expr && expr` / `expr and expr`
 - `expr || expr` / `expr or expr`
 - `(expr)`

Like `-E "package(driver) or test(use_driver_)"`, `-E "package(utility_package) and rdeps(utility_package)"`


sunshowers:
Your general design looks great. Let's do:
* `not x` to match `and` and `or`, along with a short form `! x`
* `x & y` as the short form for `x and y` (not `&&`) because this is similar to [`BitAnd` for sets](https://doc.rust-lang.org/alloc/collections/btree_set/struct.BTreeSet.html#impl-BitAnd%3C%26%27_%20BTreeSet%3CT%3E%3E)
* `x | y` and `x + y` for `x or y` for similar reasons
* `x - y` for `x and not y`
* `all()` for all tests

> `-E 'package(utility_package) and rdeps(utility_package)'`

Note that this performs the *intersection* of the two sets, which using the `and` operator honestly makes a bit less clear.

Some other questions:
* What sort of parser were you planning to use? My personal suggestion would be to use a parser combinator, and if you do so, [nom](https://docs.rs/nom/latest/nom/) would be my first suggestion (especially since nextest already uses it as a transitive dependency).
* How are you planning to do error handling? Ideally we'd integrate with something like [miette](https://docs.rs/miette/latest/miette/), following [this suggestion](https://www.reddit.com/r/rust/comments/p9t3jh/miette_a_fancy_new_diagnostics_definition_and/ha1mmyr/) by Kat.

Guiguiprim:
You are right, I meant all the tests in package `utility_package` and all the tests using `utility_package`, we should be careful not to make it error prone.

- I was thinking of `nom` !
- I didn't though about precedence yet. But I will be careful about it.
- I can look at `miette` 
- I had the same reflection about regexes, it adds parsing challenges, but yes `r'..'` looks right. I guess I would consider `literal:` default and optional in that case.
- Multiple expression sound fine.

"Sorting tests by key" you meant filtering on test names ? I guess I can keep it for the end, but as it might be useful for me I'm inclined to do it.

I probably won't have much time in the days to come, so this will be somehow slow to come.

sunshowers:
> "Sorting tests by key" you meant filtering on test names ? I guess I can keep it for the end, but as it might be useful for me I'm inclined to do it.

I mean sorting the order in which tests are run.

Guiguiprim:
I finally changed a bit the syntax:

- `package(name)` => contains `name`
- `package(=name)` => equal `name`
- `package(/name/)` => matches regex `name`

This makes it easier to parse and (I think) easier to use. `/../` is already a convention in some languages for regex.

The current limitations are:
- `equal` and `contains` can not contains `)`
- `regex` can not contains `/`

In both cases, this should not matter as they are both characters that can not be part of a test or package name.

Guiguiprim:
I have a first draft of filtering.

`deps()` and `rdeps()` are not implemented yet, but everything else is, so the following works:

- `-E "package(runner)"`
- `-E "package(=cargo-nextest)"`
- `-E "package(/nextest-.*/)"`
- `-E "test(parse) + test(basic) - test(env)"`
- `-E "all()"`
- `-E "none()"`
- `-E "test(parse) & test(expr_)"`
- `-E "test(expr_) | test(cargo)"`

Still todo:
- [ ] `deps()` / `rdeps()`
- [ ] Proper parsing error handling
- [ ] Support multiple `-E <expr>`
- [ ] Unit tests around matching

As of now, I have issues implementing `deps()`, [guppy::graph::DependsCache::depends_on()](https://docs.rs/guppy/latest/guppy/graph/struct.DependsCache.html#method.depends_on) doesn't seem to work...

Guiguiprim:
So it looks like I need to find out why my test of `-E "deps(runner)"` does not run any tests...

nextest-runner/src/test_filter/expression.rs
```diff
@@ -0,0 +1,128 @@
...
+            .packages()
+            .filter(|p| p.in_workspace())
+            .filter(|p| matcher.is_match(p.name()));
+
+        let id = artifact.package.id();
+        let mut cache = graph.new_depends_cache();
+
+        packages.any(|p| p.id() != id && cache.depends_on(p.id(), id).unwrap_or(false))
+    }
+}
```
sunshowers, line 105: This seems like it'll blow up the time complexity of the operation quite significantly:
* the number of artifacts is `O(number of workspace packages)`
* the complexity of graph traversal is `O(number of workspace packages + number of edges)`
* filtering by package is itself `O(number of workspace packages)`.

So it's worse than cubic in the number of workspace packages. For workspace-wide operations it's generally good to avoid anything past `O(n log n)` where `n` is the number of workspace packages.

Let's skip `deps` and `rdeps` operations for now and figure out how to implement them efficiently in a separate PR -- I think there's a bunch of ways we can rewrite this to make it more efficient.

nextest-runner/src/test_filter/expression_parsing.rs
```diff
@@ -0,0 +1,316 @@
+use super::expression::*;
```
sunshowers, line 1: Could you make sure to add license headers at the top of each file? They're important for code provenance.

sunshowers (commented):
This is looking great -- I'm really excited about how this is progressing!

nextest-runner/src/test_filter/expression.rs
Guiguiprim: You're right, I didn't though about that.

And we (I) need to remember to remove `--no-deps` form our call to `cargo metadata`...

Guiguiprim: We might "compile" the parsed expression using the packages graph to replace `deps()` and `rdeps()` by a set of accepted package id. This would reduce the cost inside `includes(artifact, test_name)`

nextest-runner/src/test_filter/expression_parsing.rs
Guiguiprim: Done

nextest-runner/src/test_filter/expression.rs
Guiguiprim: Rework with the compile approach. Let me known If you think this is better.

nextest-runner/src/test_filter/expression.rs
```diff
@@ -0,0 +1,128 @@
...
+    Deps(NameMatcher),
+    /// All tests in a package reverse dependencies
+    Rdeps(NameMatcher),
+    /// All tests matching a name
+    Test(NameMatcher),
+    /// All tests
+    All,
+    /// No tests
+    None,
+    // Possible addition: Binary(NameMatcher)
```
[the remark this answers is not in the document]
Guiguiprim: That was the idea. And I wasn't going to do it in this iteration.

Guiguiprim:
I'm done with my compilation optimization pass, if you want to have another look.

sunshowers:
Thank you—I have other stuff coming up within the next couple of days but will try and take a look soon.

Guiguiprim:
I worked on parsing error with `miette`.

- it does make the parsing code more complex but the output is somewhat nice.
- I also moved all the filtering stuff in a new crate.
- I added some parsing tests

parsing_error

I still need to add tests on the matching itself.

Guiguiprim:
Here we go, I:
- Added support for multiple expressions
- Added filtering tests
- Conditionally included `--no-deps` in the call to `cargo metadata`
- Made the feature experimental
- Added a first draft of documentation (this is not my stronger suit)

At this point all is working, and I'm ready for reviews and critics !

sunshowers (commented):
Thank you! It might take a little while for me to get through this, have some errands today and medical things for the rest of the week. The other maintainer is busy as well :)

Guiguiprim:
OK, noted, no problem.
Take care of yourself.

fixtures/tests-workspace-metadata.json
```diff
@@ -0,0 +1 @@
.../tests-workspace/crate-c)","dep_kinds":[{"kind":null,"target":null}]}],"features":[]},{"id":"crate_e 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-e)","dependencies":["crate_d 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-d)"],"deps":[{"name":"crate_d","pkg":"crate_d 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-d)","dep_kinds":[{"kind":null,"target":null}]}],"features":[]},{"id":"crate_f 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-f)","dependencies":["crate_d 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-d)"],"deps":[{"name":"crate_d","pkg":"crate_d 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-d)","dep_kinds":[{"kind":null,"target":null}]}],"features":[]},{"id":"crate_g 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-g)","dependencies":["crate_f 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-f)"],"deps":[{"name":"crate_f","pkg":"crate_f 0.1.0 (path+file:///home/fakeuser/tests-workspace/crate-f)","dep_kinds":[{"kind":null,"target":null}]}],"features":[]}],"root":null},"target_directory":"/home/fakeuser/tests-workspace/target","version":1,"workspace_root":"/home/fakeuser/tests-workspace","metadata":null}
```
sunshowers, line 1: Could you reformat this as pretty JSON?

nextest-filtering/src/parsing.rs
```diff
@@ -0,0 +1,668 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum Expr {
+    Not(Box<Expr>),
+    Union(Box<Expr>, Box<Expr>),
+    Intersection(Box<Expr>, Box<Expr>),
+    Set(SetDef),
+
+    Error,
+}
```
sunshowers, line 69: You could make this:

```
pub(crate) enum Expr {
    Valid(ValidExpr),
    Error,
}
```

and have `ValidExpr` be the valid expressions.

fixtures/tests-workspace/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+members = [
+  "crate-a",
+  "crate-b",
+  "crate-c",
+  "crate-d",
+  "crate-e",
+  "crate-f",
+  "crate-g",
+]
+
```
sunshowers, line 11: Do we need this workspace checked into the repo?

sunshowers (approved):
Wow, this is absolutely incredible work. Thank you so much for doing it.

I think I'm going to try it out for a little bit locally and then land it. I don't really see any issues with it at first glance.

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+pub mod error;
+mod parsing;
+
+use parsing::{Expr, RawNameMatcher, SetDef};
+
+/// Matcher for name
+///
+/// Used both for package name and test name
+#[derive(Debug, Clone)]
+pub enum NameMatcher {
```
sunshowers, line 20: Could you move the contents of lib.rs into another file? The reason is that anything that's private here is fully accessible to any other modules.

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+    None,
+    // Possible addition: Binary(NameMatcher)
+}
+
+/// Filtering expression
+///
+/// Used to filter tests to run.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FilteringExpr {
+    /// Accepts every tests not in the given expression
```
sunshowers, line 61:
```suggestion
    /// Accepts every test not in the given expression
```

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+    Not(Box<FilteringExpr>),
+    /// Accepts every tests in either given expression
```
sunshowers, line 63:
```suggestion
    /// Accepts every test in either given expression
```

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+    Union(Box<FilteringExpr>, Box<FilteringExpr>),
+    /// Accepts every tests in both given expression
```
sunshowers, line 65:
```suggestion
    /// Accepts every test in both given expressions
```

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+    Intersection(Box<FilteringExpr>, Box<FilteringExpr>),
+    /// Accepts every tests in a set
```
sunshowers, line 67:
```suggestion
    /// Accepts every test in a set
```

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+                if !errors.is_empty() {
+                    for err in errors {
+                        let report = miette::Report::new(err).with_source_code(input.to_string());
+                        eprintln!("{:?}", report);
+                    }
+                    return Err(error::FilteringExprParsingError(input.to_string()));
+                }
+
+                let in_workspace_packages: Vec<_> =
+                    graph.packages().filter(|p| p.in_workspace()).collect();
```
sunshowers, line 229: You can just do `graph.resolve_workspace().packages(DependencyDirection::Forward).collect()` which will avoid an iteration over the full package graph. Not a huge deal.

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+                let mut cache = graph.new_depends_cache();
+                match compile_expr(&expr, &in_workspace_packages, &mut cache) {
+                    Some(expr) => Ok(expr),
+                    None => {
+                        // should not happen
```
sunshowers, line 234: Could you elaborate on `should not happen`?

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+                        let err = error::Error::Unknown;
+                        let report = miette::Report::new(err).with_source_code(input.to_string());
+                        eprintln!("{:?}", report);
+                        Err(error::FilteringExprParsingError(input.to_string()))
+                    }
+                }
+            }
+            Err(_) => {
+                // should not happen
```
sunshowers, line 243: as well as here?

sunshowers (commented):
Just a few minor things.

nextest-filtering/src/error.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    fn set_tracable_info(mut self, info: TracableInfo) -> Self {
+        self.1 = self.1.set_tracable_info(info);
+        self
+    }
+}
+
+#[derive(Debug)]
+pub struct FilteringExprParsingError(pub String);
```
sunshowers, line 68: Is there a way to make this an `ExpectedError` in the cargo-nextest crate? That will make it be printed as a nicer error (since this is a user error and not an internal error within nextest).

cargo-nextest/src/dispatch.rs
```diff
@@ -308,8 +318,17 @@ impl TestBuildFilter {
             &path_mapper,
             self.platform_filter.into(),
         )?;
-        let test_filter =
-            TestFilterBuilder::new(self.run_ignored, self.partition.clone(), &self.filter);
+        let exprs = self
+            .expr_filter
+            .iter()
+            .map(|input| FilteringExpr::parse(input, graph))
+            .collect::<Result<Vec<_>, _>>()?;
```
sunshowers, line 325: Could this be reordered to happen before the build happens? This will cause parsing errors to be printed out immediately rather than after the build.

sunshowers (commented):
Using it locally now, a couple of initial comments.

nextest-filtering/src/parsing.rs
```diff
@@ -0,0 +1,668 @@
...
+            }
+            Err(err) => Err(err),
+        }
+    }
+}
+
+fn is_identifier_char(c: char) -> bool {
+    // This is use for NameMatcher::Contains(_) and NameMatcher::Equal(_)
+    // The output should be valid part of a test-name or a package name.
+    c == ':' || c.is_xid_continue()
```
sunshowers, line 188: Hmm, unfortunately this doesn't seem quite correct -- package names (and really test names with custom test harnesses) can have `-` in them, which is not XID_CONTINUE. In fact, test names can even be things like `foo.rs (line 30)` -- this is what doctests look like, for example.

I think we may want to accept arbitrary Unicode here for tests, and add ways to escape `(` and other special characters.

nextest-filtering/src/parsing.rs
```diff
@@ -0,0 +1,668 @@
...
+}
+
+#[tracable_parser]
+fn parse_identifier_part(input: Span) -> IResult<Option<String>> {
```
sunshowers, line 192: Could you return a separate `Identifier` type? The constructor for this should accept a string and normalize it internally to NFC [the way Rust does](https://github.com/rust-lang/rfcs/blob/master/text/2457-non-ascii-idents.md#reference-level-explanation).

nextest-filtering/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+        .iter()
+        .filter(|p| matcher.is_match(p.name()))
+        .map(|p| p.id());
+    let mut set = HashSet::new();
+    for id1 in packages {
+        for p2 in all_packages {
+            let id2 = p2.id();
+            if id1 == id2 {
+                continue;
+            }
```
sunshowers, line 118: I'd match the behavior of https://bazel.build/reference/query here and in `rdependencies_packages` below and remove this check. `deps()` and `rdeps()` should return the transitive closure.

For a stretch goal, you could implement the depth limit listed there.

fixtures/tests-workspace/Cargo.toml
sunshowers: What about checking in a tarball?

cargo-nextest/src/errors.rs
```diff
@@ -131,6 +139,13 @@ impl ExpectedError {
                 );
                 None
             }
+            Self::FilterExpressionParseError { expressions } => {
+                log::error!(
+                    "Failed to parse some filtering expressions: {:?}",
+                    expressions
+                );
```
sunshowers, line 146:
```suggestion
            Self::FilterExpressionParseError { .. } => {
                log::error!("failed to parse filter expression");
```

nextest-filtering/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+readme = "README.md"
+repository = "https://github.com/nextest-rs/nextest"
+documentation = "https://docs.rs/nextest-filtering"
+keywords = ["nextest", "test-runner"]
+categories = ["development-tools::testing"]
+rust-version = "1.56"
+
+[features]
+default = []
+# Must be called `trace` which is a bit infortunate
```
sunshowers, line 16:
```suggestion
# Must be called `trace` which is a bit unfortunate
```

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,47 @@
...
+- `='unicode_text'`: match anything equal to `unicode_text`
+- `/reg/`: match anything matching the regex `reg`
+
+Unicode string:
+- can contains escaped quote: `\'`
+- can contains unicode sequence: `\u{xxx}` (where `xxx` is an 1 to 6 digits hexadecimal number)
+
+Operations:
+- `set_1 & set_2` , `set_1 and set_2`: include everything in both `set_1` and `set_2`
+- `set_1 | set_2`, `set_1 or set_2`, `set_1 + set_2`: include everything in either `set_1` or `set_2`
```
sunshowers, line 34: I'd be clear here and use the words "union" and "intersection".

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,47 @@
...
+- `not set`: include everything not included in `set`
```
sunshowers, line 35: `! set` works too I think!

sunshowers (approved):
Looking fantastic! Just a couple more changes and I'm happy to finish up the rest and get this out there!

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,47 @@
...
+
+A filtering expression define a set of tests, any test in the set will be run.
+
+Basic sets:
+- `all()`: include everything
+- `test(name-matcher)`: include all tests matching `name-matcher`
+- `package(name-matcher)`: include all tests in packages matching `name-matcher`
+- `deps(name-matcher)`: include all tests in packages depended by packages matching `name-matcher` and include all tests in packages matching `name-matcher`
+- `rdeps(name-matcher)`: include all tests in packages depending on packages matching `name-matcher` and include all tests in packages matching `name-matcher`
+- `none()`: include nothing
```
sunshowers, line 19: Could you add a `partition` operator as well?

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,47 @@
...
+
+Unicode string:
+- can contains escaped quote: `\'`
+- can contains unicode sequence: `\u{xxx}` (where `xxx` is an 1 to 6 digits hexadecimal number)
+
+Operations:
+- `set_1 & set_2` , `set_1 and set_2`: include everything in both `set_1` and `set_2`
+- `set_1 | set_2`, `set_1 or set_2`, `set_1 + set_2`: include everything in either `set_1` or `set_2`
+- `not set`: include everything not included in `set`
+- `set_1 - set_2`: equivalent to `set_1 and not set_2`
```
sunshowers, line 36: What is the operator precedence here -- are there tests for that? Could you document this?

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,47 @@
...
+- `package(name-matcher)`: include all tests in packages matching `name-matcher`
+- `deps(name-matcher)`: include all tests in packages depended by packages matching `name-matcher` and include all tests in packages matching `name-matcher`
+- `rdeps(name-matcher)`: include all tests in packages depending on packages matching `name-matcher` and include all tests in packages matching `name-matcher`
+- `none()`: include nothing
+
+Name matcher:
+- `text`: match anything containing `text`
+- `'unicode_text'`: match anything containing `unicode_text`
+- `=text`: match anything equal to `text`
+- `='unicode_text'`: match anything equal to `unicode_text`
```
sunshowers, line 25: Hmm, I don't think the single-quote scheme will work because on Unix shells the escaping becomes really awkward. I think we have to parse all text without using quote marks, then read input character by character, looking for the escapes you've listed below.

site/src/book/filtering-expression.md
Guiguiprim: Doc added.

There is a test for it `nextest-filtering parsing::test_parse_epxr_precedence`

site/src/book/filtering-expression.md
Guiguiprim: Like `--partition`, this might be a little bit more complex and requires state.

Can we do it after, in an other PR ?

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,52 @@
...
+- `none()`: include nothing
+
+Name matcher:
+- `unicode_text`: match anything containing `unicode_text`
+- `=unicode_text`: match anything equal to `unicode_text`
+- `/reg/`: match anything matching the regex `reg`
+
+Unicode string:
+- can contains escaped closing parenthesis: `\)`
+- can contains unicode sequence: `\u{xxx}` (where `xxx` is an 1 to 6 digits hexadecimal number)
```
sunshowers, line 28: Also `\r`, `\n` etc, right?

site/src/book/filtering-expression.md
```diff
@@ -0,0 +1,52 @@
...
+Operations:
+- `set_1 & set_2` , `set_1 and set_2`: the intersection of `set_1` and `set_2`
+- `set_1 | set_2`, `set_1 or set_2`, `set_1 + set_2`: the union of `set_1` or `set_2`
+- `not set`, `!set`: include everything not included in `set`
+- `set_1 - set_2`: equivalent to `set_1 and not set_2`
+- `(set)`: include everything in `set`
+
+Precedences from lowest to highest:
+- `not`, `()`
+- `&`, `|`, `-`
```
sunshowers, line 39: How does an expression with `&` and `|` get resolved? In general I'd expect us to follow a similar pattern to https://doc.rust-lang.org/reference/expressions.html#expression-precedence -- from highest to lowest:

* `()`
* `!`, `not`
* `&`, `and`
* `|`, `+`, `or`

Only real difference is that `+` is grouped with `|`.

sunshowers (approved):
Looks great -- just a couple more changes, and I'm happy to land this and finish it up myself if you like.

sunshowers:
Going to land this and clean it up, thanks again for all your incredible efforts here.

Issue #123, opened 2022-03-20
use `CARGO_TERM_COLOR` for term color?


skyzh:
https://github.com/nextest-rs/nextest/blob/ceb871d41e386eae16f16a83674219fb5ea3a957/cargo-nextest/src/output.rs#L23-L32

As `nextest` is a cargo subcommand, it's better to follow cargo's convention? Therefore, I suggest using `CARGO_TERM_COLOR` instead of `NEXTEST_COLOR` environment variable for this config.

PR #124, opened 2022-03-20
chore: use CARGO env variables for output settings

skyzh:

fix https://github.com/nextest-rs/nextest/issues/123


sunshowers:
Thanks! I think CARGO_TERM_COLOR makes sense but the verbose flag should probably be separated out between cargo and nextest.

skyzh:
Thanks! I've reverted the changes. PTAL cc @sunshowers

Issue #144, opened 2022-04-13
Add a way to escape `/` in regexes


sunshowers:
While testing out expression filtering, I noticed that there's currently no way to escape a `/` in a regex. Ideally we'd support this via a `\/` sequence.

cc @Guiguiprim, what's the best way to handle this in your opinion?

Guiguiprim:
I guess in a way similar to how unicode string are parsed.

Guiguiprim:
That is exactly what I was thinking of. I'm going to have a try at it.


Speaking of escaping I realize that the unicode parsing (that I took from `nom` examples) also support escaping whitespaces, aka:

```
"test(d\
d)
```

is equivalent to

```
test(dd)
```

I'm not sure if we should keep it or how to document it.

More generally, I wonder if I shouldn't have simplified it a bit more...

sunshowers:
Ah yeah I was looking at that. Not really worth keeping I think. We can always add it later if there's demand.

sunshowers:
Done in #146, thanks!

PR #146, opened 2022-04-13
Add `/` escaping in regex, remove useless whitespace escaping in unicode string

Guiguiprim:
Fix for #144

- Adds the `/` for regex
- Removes the whitespace escaping from unicode string
- Adds a small example for quick manual filtering expression parsing test


sunshowers (approved):
Excellent, thank you!

Issue #155, opened 2022-04-15
Default package filtering to exact while keeping default test filtering as contains


sunshowers:
Having used filter expressions for a bit -- generally, most users expect package filtering to be based exactly on the package name, and test filtering to match on part of the test name. So default to these while allowing users to change them with the matchers that already exist.

This blocks the full release of filter expressions.

cc @Guiguiprim (sorry for ccing you to these, lmk if I should stop!)

Guiguiprim:
Oh... That feels indeed the right move.

I will look at it also.

I'm happy to help you finish properly this feature ! (You already did a nice job at cleaning it up 🙂)

sunshowers:
Fixed in #156.

PR #156, opened 2022-04-15
[nextest-filtering] Add comma support, change matcher default, ...

### LINKED ISSUE #153: Handle commas in filter expressions
We should leave open commas as an extension point for further arguments. This means rejecting commas in name matchers with unary set operators for now, and adding `\,` as an escape (doesn't need to be done within a regex context).

This probably blocks the full release so filing this to track this issue.

cc @Guiguiprim if you're interested in solving this.

Guiguiprim:
Fixes #153 and #155

- Fix `parser` example output following some refactoring of error handling
- Generate error on empty strings like `test()` and `test(=)` (I think we should ?)
- Add support for `,` as an arguments separator
- Changes default matching strategy to `equal` for packages sets
- Add `~` as a shorter way of creating `contains` matcher

### DIFF
--- a/nextest-filtering/examples/parser.rs
+++ b/nextest-filtering/examples/parser.rs
@@ -8,6 +8,7 @@
 use camino::Utf8PathBuf;
 use clap::Parser;
 use guppy::graph::PackageGraph;
+use nextest_filtering::errors::FilterExpressionParseErrors;
 
 #[derive(Debug, Parser)]
 struct Args {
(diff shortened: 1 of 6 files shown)


sunshowers (approved):
I love these changes and the new `~` shortcut -- in fact I think I'm going to go ahead and remove the `contains:` operator in favor of the `~` operator!

Thank you so much -- your work is invaluable.

Issue #213, opened 2022-05-18
feature suggestion: terminate slow tests after a given deadline and regard them as failed


steveeJ:
without this, terminating a frozen test suite relies on an external watcher that keeps track of time. since the test harness and nextest itself already track time for each test, it would be great to have deadline support natively in the cargo stack.

for this potential feature it's actually an advantage that nextest runs each test in a separate process, because terminating processes works reliably as opposed to terminating threads which requires cooperation.

sunshowers:
Thanks for the request! Yes, I agree this makes a lot of sense.

It would be interesting to see how this interacts with #27. I guess we can apply it to the entire process.

sunshowers:
This was released as cargo-nextest 0.9.18.

PR #214, opened 2022-05-20
feat: support termination of slow tests

steveeJ:
this PR shows an opinionated minimal implementation for #213 and it meant to serve as tangible initial input for the design of this feature.

cc @sunshowers

### DIFF
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -55,10 +55,13 @@ success-output = "never"
 # to false.
 fail-fast = true
 
-# Treat a test that takes longer than this period as slow, and print a message.
-# Given a non-zero positive integer the test will be terminated after the number
+# Treat a test that takes longer than the configured 'period' as slow, and print a message.
+#
+# Optional: specify the parameter 'terminate-after' with a non-zero integer,
+# which will cause slow tests to be terminated after the specified number of
 # periods have passed.
-slow-timeout = { period = "60s", terminate-after = 0 }
+# Example: slow-timeout = { period = "60s", terminate-after = 2; }
+slow-timeout = { period = "60s" }
 
 [profile.default.junit]
 # Output a JUnit report into the given file inside 'store.dir/<profile-name>'.
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -55,8 +55,10 @@ success-output = "never"
 # to false.
 fail-fast = true
 
-# Treat a test that takes longer than this as slow, and print a message.
-slow-timeout = "60s"
+# Treat a test that takes longer than this period as slow, and print a message.
+# Given a non-zero positive integer the test will be terminated after the number
+# periods have passed.
+slow-timeout = { period = "60s", terminate-after = 0 }
 
 [profile.default.junit]
 # Output a JUnit report into the given file inside 'store.dir/<profile-name>'.
--- a/nextest-runner/src/config.rs
+++ b/nextest-runner/src/config.rs
@@ -10,7 +10,7 @@ use crate::{
 use camino::{Utf8Path, Utf8PathBuf};
 use config::{builder::DefaultState, Config, ConfigBuilder, File, FileFormat};
 use serde::{de::IntoDeserializer, Deserialize};
-use std::{collections::HashMap, fmt, time::Duration};
+use std::{collections::HashMap, fmt, num::NonZeroUsize, time::Duration};
 
 /// Overall configuration for nextest.
 ///
@@ -305,7 +305,7 @@ pub struct SlowTimeout {
     #[serde(with = "humantime_serde")]
     pub(crate) period: Duration,
     #[serde(default)]
-    pub(crate) terminate_after: Option<usize>,
+    pub(crate) terminate_after: Option<NonZeroUsize>,
 }
 
 fn require_deserialize_slow_timeout<'de, D>(deserializer: D) -> Result<SlowTimeout, D::Error>
@@ -331,7 +331,7 @@ where
         fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
             write!(
                 formatter,
-                "a table ({{ period = \"60s\", terminate-after = \"2\" }}) or a string (\"60s\")"
+                "a table ({{ period = \"60s\", terminate-after = 2 }}) or a string (\"60s\")"
             )
         }
 
@@ -417,7 +417,7 @@ mod tests {
 
     #[test_case(
         "",
-        SlowTimeout { period: Duration::from_secs(60), terminate_after: Some(0)},
+        Ok(SlowTimeout { period: Duration::from_secs(60), terminate_after: None }),
         None
 
         ; "empty config is expected to use the hardcoded values"
@@ -427,7 +427,7 @@ mod tests {
             [profile.default]
             slow-timeout = "30s"
         "#},
-        SlowTimeout { period: Duration::from_secs(30), terminate_after: None },
+        Ok(SlowTimeout { period: Duration::from_secs(30), terminate_after: None }),
         None
 
         ; "overrides the default profile"
@@ -440,14 +440,60 @@ mod tests {
             [profile.ci]
             slow-timeout = { period = "60s", terminate-after = 3 }
         "#},
-        SlowTimeout { period: Duration::from_secs(30), terminate_after: None, },
-        Some(SlowTimeout { period: Duration::from_secs(60), terminate_after: Some(3), })
+        Ok(SlowTimeout { period: Duration::from_secs(30), terminate_after: None }),
+        Some(SlowTimeout { period: Duration::from_secs(60), terminate_after: Some(NonZeroUsize::new(3).unwrap()), })
 
         ; "adds a custom profile 'ci'"
     )]
+    #[test_case(
+        indoc! {r#"
+            [profile.default]
+            slow-timeout = { period = "60s", terminate-after = 3 }
+
+            [profile.ci]
+            slow-timeout = "30s"
+        "#},
+        Ok(SlowTimeout { period: Duration::from_secs(60), terminate_after: Some(NonZeroUsize::new(3).unwrap()) }),
+        Some(SlowTimeout { period: Duration::from_secs(30), terminate_after: None, })
+
+        ; "ci profile uses string notation"
+    )]
+    #[test_case(
+        indoc! {r#"
+            [profile.default]
+            slow-timeout = { period = "60s" }
+        "#},
+        Ok(SlowTimeout { period: Duration::from_secs(60), terminate_after: None }),
+        None
+
+        ; "partial table"
+    )]
+    #[test_case(
+        indoc! {r#"
+            [profile.default]
+            slow-timeout = { period = "60s", terminate-after = 0 }
+        "#},
+        Err("err: invalid value: integer `0`, expected a nonzero usize"),
+        None
+
+        ; "zero terminate-after should fail"
+    )]
+    #[test_case(
+        indoc! {r#"
+            [profile.default]
+            slow-timeout = "60s"
+
+            [profile.ci]
+            slow-timeout = { terminate-after = 3 }
+        "#},
+        Err("err: missing field `period`"),
+        None
+
+        ; "partial slow-timeout table should error"
+    )]
     fn slowtimeout_adheres_to_hierarchy(
         workspace_config: &str,
-        expected_default: SlowTimeout,
+        expected_default: Result<SlowTimeout, &str>,
         maybe_expected_ci: Option<SlowTimeout>,
     ) {
         let workspace_dir = tempdir().unwrap();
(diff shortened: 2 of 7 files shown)


sunshowers:
Thanks for the PR! I'm currently in the middle of reviewing this, doing some research into some possible solutions here right now.

nextest-runner/default-config.toml
```diff
@@ -55,6 +55,9 @@ fail-fast = true
 # Treat a test that takes longer than this as slow, and print a message.
 slow-timeout = "60s"
 
+# Number of slow-timeout periods to tolerate before terminating a test.
+slow-timeout-tolerance = 2
```
sunshowers, line 59: What I'd do here is something like:

```
slow-timeout = { display-after = "60s", kill-after = 2 }
```

Define `kill-after = 0` as the default (never kill).

Also, maintain the current compatibility with strings by using a custom deserialize impl. See https://github.com/nextest-rs/nextest/compare/main...sunshowers:slow-timeout for an example of how to do that (feel free to use that code as you wish).

sunshowers (changes requested):
Thank you! I'm pretty happy with this feature. I think the UI needs a few changes but otherwise it looks good.

nextest-runner/default-config.toml
steveeJ: thanks for providing readily available code for this feature! i cherry-picked your commit. however, i removed the Default/Custom distinction as it seemed redundant. please let me know if i missed something that made that distinction necessary that got lost with my changes.

nextest-runner/src/runner.rs
```diff
@@ -378,6 +383,16 @@ impl<'a> TestRunner<'a> {
                             test_instance: test,
                             elapsed: stopwatch.elapsed(),
                         });
+
+                        if timeout_hit > self.slow_timeout_tolerance {
+                            handle
+                                .kill()
```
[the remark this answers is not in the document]
steveeJ: it's possible to implement the grace period you've outlined on unix systems, which i did in my most recent commit. the grandchild situation is unfortunate and you probably saw that that [it's mentioned in duct's gotcha.md](https://github.com/oconnor663/duct.py/blob/master/gotchas.md#killing-grandchild-processes).

i'm happy to improve the unix solution i'm proposing here, but would like to leave anything else outside of the scope of this PR.

nextest-runner/default-config.toml
```diff
@@ -52,8 +52,10 @@ success-output = "never"
 # to false.
 fail-fast = true
 
-# Treat a test that takes longer than this as slow, and print a message.
-slow-timeout = "60s"
+# Treat a test that takes longer than this period as slow, and print a message.
+# Given a non-zero positive integer the test will be shutdown the number periods
```
sunshowers, line 56: I like `period` but let's rename it `kill-after` to make it clear to users what happens.

nextest-runner/src/config.rs
```diff
@@ -283,11 +284,58 @@ struct DefaultProfileImpl {
...
     fail_fast: bool,
-    #[serde(with = "humantime_serde")]
-    slow_timeout: Duration,
+    #[serde(deserialize_with = "deserialize_slow_timeout")]
+    slow_timeout: SlowTimeout,
     junit: DefaultJunitImpl,
 }
 
+/// Type for the slow-timeout config key.
+#[derive(Clone, Debug, Deserialize, Default)]
```
sunshowers, line 293: Can this be `Copy`?

nextest-runner/src/reporter.rs
```diff
@@ -493,11 +494,12 @@ impl<'a> TestReporter<'a> {
                         passed,
                         flaky,
                         failed,
+                        timeout,
```
sunshowers, line 497: Let's call this `timed_out`

nextest-runner/src/reporter.rs
```diff
@@ -540,6 +542,15 @@ impl<'a> TestReporter<'a> {
                     )?;
                 }
 
+                if *timeout > 0 {
+                    write!(
+                        writer,
+                        "{} {}, ",
+                        timeout.style(self.styles.count),
+                        "timeout".style(self.styles.fail),
```
sunshowers, line 550: `"timed out"`

nextest-runner/src/reporter.rs
```diff
@@ -802,6 +813,7 @@ struct Styles {
     pass: Style,
     retry: Style,
     fail: Style,
+    timeout: Style,
```
sunshowers, line 816: Is this used anywhere?

nextest-runner/src/runner.rs
```diff
@@ -369,15 +371,63 @@ impl<'a> TestRunner<'a> {
...
+                                #[cfg(unix)]
+                                let exited = {
+                                    use duct::unix::HandleExt;
+                                    use libc::SIGTERM;
+
+                                    let _ = handle.send_signal(SIGTERM);
+
+                                    let mut i = 0;
+
+                                    // give the process a grace periodof 10s
```
sunshowers, line 403:
```suggestion
                                    // give the process a grace period of 10s
```

nextest-runner/src/runner.rs
```diff
@@ -577,6 +630,9 @@ pub struct RunStats {
     /// The number of tests that failed.
     pub failed: usize,
 
+    /// The number of tests that timed out.
+    pub timeout: usize,
```
sunshowers, line 634: Let's call this `timed_out` as above to match `failed` and `exec_failed`.

nextest-runner/src/config.rs
```diff
@@ -309,9 +357,8 @@ struct CustomProfileImpl {
     success_output: Option<TestOutputDisplay>,
     #[serde(default)]
     fail_fast: Option<bool>,
-    #[serde(with = "humantime_serde")]
-    #[serde(default)]
-    slow_timeout: Option<Duration>,
+    #[serde(default, deserialize_with = "deserialize_slow_timeout")]
+    slow_timeout: SlowTimeout,
```
sunshowers, line 361: This doesn't look right -- what happens if the custom profile doesn't have a slow timeout `period` set? In that case it should inherit the default profile's `period` setting rather than setting it to the default `Duration` (which is zero).

nextest-runner/src/config.rs
```diff
@@ -283,11 +284,58 @@ struct DefaultProfileImpl {
...
     junit: DefaultJunitImpl,
 }
 
+/// Type for the slow-timeout config key.
+#[derive(Clone, Debug, Deserialize, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct SlowTimeout {
+    #[serde(with = "humantime_serde")]
+    pub(crate) period: Duration,
+    pub(crate) shutdown_after: Option<usize>,
```
sunshowers, line 298: So this should probably be deserialized into an enum:

```
enum KillAfter {
    Never,
    Periods(usize), // can also use NonZeroUsize here
}
```

Then the 0 is deserialized into Never and anything else is deserialized into the `Periods` variant.

Again, the `Option` is a bit confusing here -- if a profile doesn't have the option set, it should fall back to the repository-specific default config, which in turn should fall back to the default profile shipped with nextest. See [Hierarchical configuration](https://nexte.st/book/configuration.html#hierarchical-configuration) for how it should work.

(This is why the separation between `Default` and `Custom` is important -- it is required for hierarchical configuration to work properly. If Rust had higher-kinded types a la Haskell, we could unify the code for these--but unfortunately it doesn't.)

sunshowers (changes requested):
Thanks! Getting close to being ready -- a few more changes required. As explained below, the `Default`/`Custom` separation is actually quite important given the way hierarchical configuration works.

nextest-runner/src/config.rs
steveeJ: i understand, the initial version i proposed did not adhere to the hierarchical configuration logic.
i just pushed a commit that implements it and i'm curious for your feedback. the duplication bothers me and i'd like to avoid it if possible.

nextest-runner/src/config.rs
steveeJ: instead of `Clone` or in addition?

nextest-runner/default-config.toml
sunshowers: That sounds good to me.

nextest-runner/src/config.rs
sunshowers: It would be in addition -- `Copy` requires (derived) `Clone`.

sunshowers:
The hierarchical config solution looks really good to me! Thanks for doing that.

Could you add some tests for the cases here?

nextest-runner/src/reporter.rs
steveeJ: no, removed :slightly_smiling_face:

sunshowers:
I think this just needs some tests for config parsing but otherwise it looks ready to land. Thanks again for doing this.

steveeJ:
> The hierarchical config solution looks really good to me! Thanks for doing that.

my pleasure! i'm happy we found a solution that doesn't involve redundancy :relieved: 

 > Could you add some tests for the cases here?

added :+1: please let me now if i missed an important case.

nextest-runner/src/config.rs
```diff
@@ -345,4 +414,74 @@ mod tests {
...
+            [profile.default]
+            slow-timeout = "30s"
+
+            [profile.ci]
+            slow-timeout = { period = "60s", terminate-after = 3 }
+        "#},
+        SlowTimeout { period: Duration::from_secs(30), terminate_after: None, },
+        Some(SlowTimeout { period: Duration::from_secs(60), terminate_after: Some(3), })
+
+        ; "adds a custom profile 'ci'"
```
sunshowers, line 446: Could you add more test cases here:

```toml
[profile.default]
slow-timeout = { period = "60s" }
```

```toml
[profile.default]
slow-timeout = { period = "60s", terminate-after = 0 }
```

```toml
[profile.default]
slow-timeout = { period = "60s", terminate-after = 3 }

[profile.ci]
slow-timeout = "30s"
```

```toml
[profile.default]
slow-timeout = "60s"

[profile.ci]
slow-timeout = { terminate-after = 3 }
```

This one I think should error out -- if you're specifying `terminate-after` you should also specify the period to make it really clear how long you're going to terminate after.

nextest-runner/default-config.toml
```diff
@@ -55,8 +55,10 @@ success-output = "never"
 # to false.
 fail-fast = true
 
-# Treat a test that takes longer than this as slow, and print a message.
-slow-timeout = "60s"
+# Treat a test that takes longer than this period as slow, and print a message.
+# Given a non-zero positive integer the test will be terminated after the number
+# periods have passed.
```
sunshowers, line 60:
```suggestion
# terminate-after is an optional parameter: if it is non-zero, the test will be
# terminated after the specified number of periods have passed.
```

Also add a note here saying something like:

```toml
# If specified as `slow-timeout = "60s"`, terminate-after is set to 0.
```

nextest-runner/src/config.rs
```diff
@@ -291,11 +292,76 @@ struct DefaultProfileImpl {
...
+{
+    struct V;
+
+    impl<'de2> serde::de::Visitor<'de2> for V {
+        type Value = Option<SlowTimeout>;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            write!(
+                formatter,
+                "a table ({{ period = \"60s\", terminate-after = \"2\" }}) or a string (\"60s\")"
```
sunshowers, line 334:
```suggestion
                "a table ({{ period = \"60s\", terminate-after = 2 }}) or a string (\"60s\")"
```

nextest-runner/src/config.rs
```diff
@@ -291,11 +292,76 @@ struct DefaultProfileImpl {
...
 
+/// Type for the slow-timeout config key.
+#[derive(Clone, Copy, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(rename_all = "kebab-case")]
+pub struct SlowTimeout {
+    #[serde(with = "humantime_serde")]
+    pub(crate) period: Duration,
+    #[serde(default)]
+    pub(crate) terminate_after: Option<usize>,
```
sunshowers, line 308: So I'm a little concerned here that terminate_after can be both `Some(0)` and `None`, which are two separate values that just happen to be treated the same. I'm wondering if this can be represented as an `Option<NonZeroUsize>`, or at least with a comment here clearly explaining that `Some(0)` and `None` are the same.

steveeJ: i hadn't been aware that `NonZeroUsize` exists as a type. i took your last mentioning of it as a descriptive example :slightly_smiling_face:  as i share your concern about the UX here i'm currently experimenting using `NonZeroUsize`.

steveeJ:
@sunshowers thanks a lot for the thoughtful suggestions. i addressed them all and think they're a great improvement.

on my previous comment

> on a self-review it occurred to me that i haven't added a test for the test termination itself. i think this is important to ensure that the feature keeps working. i'm going to add this test. other than that i'm happy with this PR.

i looked through integration tests and couldn't confidently start adding the case i have in mind. at this late hour of my day i found them a bit too intricate to make quick progress :smile: if it is convenient for you i'd greatly appreciate your assistance in adding a test or even just a stub for me to get started there.

sunshowers:
I think this covers everything -- looks good, thanks. I'll add the termination test and get a release out later today or tomorrow.

sunshowers:
Just released this as cargo-nextest 0.9.18.

PR #230, opened 2022-05-25
add support for creating and using archives

### LINKED ISSUE #229: Reduce the binary size needed for build reusing
Currently the entire target directory is needed when reusing build on another machine. This is quite a setback for us because previously we only need to transfer a specific test binary under `target/debug/deps/` generated by `cargo test --no-run`.

I'm wondering if it's possible to reduce the binaries that must be shared between two machines, and make this information easy enough to use. Or better yet, package the necessary test artifacts by nextest itself.

This would be especially useful for test partitioning. For our use case, test cases from one large repo are sharded to tens of partitions. With this feature, each partition wouldn't have to download all the test binaries.

sunshowers:
TODO:
- [x] Create archives.
- [x] Output correct file count.
- [x] Extract archives.
- [x] `--archive` and `--extract-to` flags.
- [x] `--binaries-metadata` + `--cargo-metadata` clases with `--archive`.
- [x] `--archive` clashes with Cargo-related options.
- [x] `cargo nextest archive` shouldn't accept reuse-build options.
- [x] Also grab libraries (.so files) from the build output.
- [x] While archiving, always save cargo metadata with deps.

Issue: #229

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomicwrites"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eb8f2cd6962fa53c0e2a9d3f97eaa7dbd1e3cbbeeb4745403515b42ae07b3ff6"
+dependencies = [
+ "tempfile",
+ "winapi",
+]
+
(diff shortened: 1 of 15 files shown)


sunshowers:
Pretty happy at this -- apologies to future readers for clubbing together everything in one commit, though!

PR #265, opened 2022-06-09
support passing filter as test binary arguments

tabokie:
Adding partial support for specifying test binary arguments like cargo. It can
greatly simplify the transitioning from cargo to nextest.

Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -90,11 +90,13 @@ impl AppOpts {
         match self.command {
             Command::List {
                 cargo_options,
-                build_filter,
+                mut build_filter,
                 message_format,
                 list_type,
                 reuse_build,
+                mut test_binary_args,
             } => {
+                build_filter.merge_test_binary_args(&mut test_binary_args);
                 let base = BaseApp::new(
                     self.output,
                     reuse_build,
@@ -111,11 +113,13 @@ impl AppOpts {
                 profile,
                 no_capture,
                 cargo_options,
-                build_filter,
+                mut build_filter,
                 runner_opts,
                 reporter_opts,
                 reuse_build,
+                mut test_binary_args,
             } => {
+                build_filter.merge_test_binary_args(&mut test_binary_args);
                 let base = BaseApp::new(
                     self.output,
                     reuse_build,
@@ -139,6 +143,7 @@ impl AppOpts {
                 archive_file,
                 archive_format,
                 zstd_level,
+                test_binary_args,
             } => {
                 let app = BaseApp::new(
                     self.output,
@@ -208,6 +213,10 @@ enum Command {
 
         #[clap(flatten)]
         reuse_build: ReuseBuildOpts,
+
+        /// Test binary arguments. Partially supported.
+        #[clap(value_name = "args", last = true)]
+        test_binary_args: Vec<String>,
     },
     /// Build and run tests
     ///
@@ -243,6 +252,10 @@ enum Command {
 
         #[clap(flatten)]
         reuse_build: ReuseBuildOpts,
+
+        /// Test binary arguments. Partially supported.
+        #[clap(value_name = "args", last = true)]
+        test_binary_args: Vec<String>,
     },
     /// Build and archive tests
     ///
@@ -282,6 +295,10 @@ enum Command {
         )]
         zstd_level: i32,
         // ReuseBuildOpts, while it can theoretically work, is way too confusing so skip it.
+
+        /// Test binary arguments. Unused. Kept for compatibility reason.
+        #[clap(value_name = "args", last = true)]
+        test_binary_args: Vec<String>,
     },
 }
 
@@ -350,10 +367,9 @@ struct TestBuildFilter {
     #[clap(
         long,
         possible_values = RunIgnored::variants(),
-        default_value_t,
         value_name = "WHICH",
     )]
-    run_ignored: RunIgnored,
+    run_ignored: Option<RunIgnored>,
 
     /// Test partition, e.g. hash:1/2 or count:2/3
     #[clap(long)]
@@ -402,7 +418,7 @@ impl TestBuildFilter {
             self.platform_filter.into(),
         )?;
         let test_filter = TestFilterBuilder::new(
-            self.run_ignored,
+            self.run_ignored.unwrap_or_default(),
             self.partition.clone(),
             &self.filter,
             filter_exprs,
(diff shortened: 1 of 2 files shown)


tabokie:
@sunshowers Another feature request while migrating from cargo. PTAL~

sunshowers:
Thanks for the contribution?

Hmm, I'm not totally convinced by the direction of this diff. Could you elaborate on the goal and intended audience for this a bit more? Is it to be able to pass in arguments like `--include-ignored` after a `--`? If so, I'm hesitant to introduce this sort of hidden alternative method to control the ignored status. Instead, I'd rather look through the passed in filters, warn on such arguments (and also `--exact` and `--skip` once filter expression support is stable) and direct users to the recommended ways instead.

tabokie:
Yes, I understand your concern. Personally I think of it as a first step to support test binary arguments (instead of a temporary workaround for certain filters). Ultimately we might want to push down unsupported test binary arguments to internal calls.

Aside from that, currently my main use case is really to keep the test command consistent with cargo workflow. E.g. `cargo $CARGO_TEST_COMMAND --workspace -- --ignored` can work for both `CARGO_TEST_COMMAND="test"` and `CARGO_TEST_COMMAND="nextest run"`

cargo-nextest/src/dispatch.rs
```diff
@@ -86,15 +86,21 @@ impl AppOpts {
                 .iter()
                 .any(|expr| FilteringExpr::needs_deps(expr))
         }
+        fn handle_unsupported_test_binary_args(args: Vec<String>) {
+            log::warn!("skipping unsupported test binary arguments: {:?}", args);
+        }
```
sunshowers, line 91: This should error, not warn.

cargo-nextest/src/dispatch.rs
```diff
@@ -139,7 +148,9 @@ impl AppOpts {
                 archive_file,
                 archive_format,
                 zstd_level,
+                test_binary_args,
```
sunshowers, line 151: The `archive` command shouldn't support these at all.

cargo-nextest/src/dispatch.rs
```diff
@@ -281,6 +300,10 @@ enum Command {
             allow_hyphen_values = true
         )]
         zstd_level: i32,
+
+        /// Test binary arguments. Unused. Kept for compatibility reason.
+        #[clap(value_name = "args", last = true)]
+        test_binary_args: Vec<String>,
```
sunshowers, line 306: I'd drop this entirely -- `archive` has no equivalent with `cargo test`.

cargo-nextest/src/dispatch.rs
```diff
@@ -412,6 +434,30 @@ impl TestBuildFilter {
...
+                true
+            }
+        });
+        for f in ignore_filter {
+            if self.run_ignored.is_some() && self.run_ignored != Some(f) {
+                log::warn!("skipping conflicting ignore filter: {}", f);
+            }
+            self.run_ignored = Some(f);
+        }
+    }
```
sunshowers, line 460: Need tests for this with the different cases.

cargo-nextest/src/dispatch.rs
```diff
@@ -412,6 +434,30 @@ impl TestBuildFilter {
...
+                self.filter.push(s.to_owned());
+                false
+            } else {
+                true
+            }
+        });
+        for f in ignore_filter {
+            if self.run_ignored.is_some() && self.run_ignored != Some(f) {
+                log::warn!("skipping conflicting ignore filter: {}", f);
+            }
```
sunshowers, line 457: Think it's best to error out entirely here so that `--run-ignored` and `-- --ignored` are treated as incompatible with each other.

cargo-nextest/src/dispatch.rs
```diff
@@ -412,6 +434,30 @@ impl TestBuildFilter {
...
+                false
+            } else if s == "--ignored" {
+                ignore_filter.push(RunIgnored::IgnoredOnly);
+                false
+            } else if !s.starts_with('-') {
+                self.filter.push(s.to_owned());
+                false
+            } else {
+                true
+            }
```
sunshowers, line 452: Is this how test binaries interpret conflicts between `--include-ignored` and `--ignored`? Please double-check.

Also, `ignore_filter` can just be an `Option` I think if the last one between `--include-ignored` and `--ignored` wins.

cargo-nextest/src/dispatch.rs
```diff
@@ -1013,6 +1059,11 @@ mod tests {
             // ---
             "cargo nextest list -E deps(foo)",
             "cargo nextest run --filter-expr 'test(bar)' --package=my-package test-filter",
+            // ---
+            // Test binary arguments
+            // ---
+            "cargo nextest run -- --ignored",
+            "cargo nextest list -- --run-ignored",
```
sunshowers, line 1066: The second one should be `--include-ignored`.

sunshowers (changes requested):
Thanks for the response. I do see the use case and after having given it some thought, I think it would be OK to support an extremely limited form of this. `--ignored` and `--include-ignored`, test binary literals, and nothing else.

cargo-nextest/src/dispatch.rs
tabokie: Verified that the test binary will return `error: the options --include-ignored and --ignored are mutually exclusive`. Will keep it consistent.

tabokie:
> it would be OK to support an extremely limited form of this. --ignored and --include-ignored, test binary literals, and nothing else.

In this case, I'll use clap to implement the parsing logic instead.

cargo-nextest/src/dispatch.rs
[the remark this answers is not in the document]
sunshowers: This comment still needs to be addressed (and tests added for it).

sunshowers (commented):
Thanks for doing this! One more comment.

cargo-nextest/src/dispatch.rs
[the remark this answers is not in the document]
tabokie: Good catch. Didn't notice this is a valid usage.

cargo-nextest/src/dispatch.rs
```diff
@@ -243,6 +258,10 @@ enum Command {
 
         #[clap(flatten)]
         reuse_build: ReuseBuildOpts,
+
+        /// Test binary arguments. Partially supported.
+        #[clap(value_name = "args", last = true)]
+        test_binary_args: Vec<String>,
```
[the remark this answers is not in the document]
tabokie: Currently it's
```
USAGE:
    cargo nextest run [OPTIONS] [FILTERS]... [-- <args>...]
ARGS:
    <FILTERS>...
            Test name filter

    <args>...
            Test binary arguments. Partially supported. Kept for compatibility reason
```
For reference, this is cargo's:
```
USAGE:
    cargo test [OPTIONS] [TESTNAME] [-- <args>...]
ARGS:
    <TESTNAME>    If specified, only run tests containing this string in their names
    <args>...     Arguments for the test binary
```

Is it close enough?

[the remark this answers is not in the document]
sunshowers: Would prefer this to be renamed `test-binary-args` to make it clear that this is distinct from `cargo nextest run`'s own args, but I can do this myself.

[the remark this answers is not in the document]
tabokie: No problem, just double checking.

cargo-nextest/src/dispatch.rs
```diff
@@ -412,6 +454,44 @@ impl TestBuildFilter {
...
+                ignore_filters.push((s.clone(), RunIgnored::All));
+                false
+            } else if s == "--ignored" {
+                ignore_filters.push((s.clone(), RunIgnored::IgnoredOnly));
+                false
+            } else if !s.starts_with('-') {
+                self.filter.push(s.to_owned());
+                false
+            } else if s.chars().all(|c| c == '-') {
+                false
```
sunshowers, line 471: This is still not quite there:
1. This needs to just be `--` (a single `-` doesn't match, nor does `---`).
2. Once this is hit, it should go into a mode where the `s.starts_with('-')` check is disabled and everything is pushed into the filter.

sunshowers:
That needless_bool lint suggestion is unhelpful and should be disabled for this block.

sunshowers:
Looks fantastic now, thanks! Going to land this once CI passes -- hopefully get a release out in a few days as well.

PR #283, opened 2022-06-19
treat benchmarks as normal tests

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

Last I checked this behavior is consistent with cargo test. Bench cases that appear in test binary list should be considered a normal test.

### DIFF
--- a/nextest-runner/src/list/test_list.rs
+++ b/nextest-runner/src/list/test_list.rs
@@ -503,22 +503,19 @@ impl<'g> TestList<'g> {
         // ...
 
         list_output.lines().filter_map(move |line| {
-            if line.ends_with(": benchmark") {
-                // These lines are produced by the default Rust benchmark harness (#[bench]).
-                // Ignore them.
-                return None;
-            }
-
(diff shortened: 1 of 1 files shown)


sunshowers:
Thanks for the PR!

What happens if a benchmark is invoked? Is it run just once or many times?

Note that the most popular benchmarking library criterion.rs doesn't currently work with nextest: https://github.com/bheisler/criterion.rs/issues/562

tabokie:
@sunshowers In the case of `cargo test`, all `#[bench]` benches will behave just like normal unit tests (run just once). They will be run even if `--tests` is passed. Cases managed by `criterion.rs` will not be run, because they are not contained in the test binary generated by `cargo test`.

sunshowers:
Thanks -- in that case this makes sense. I'll add a note to the release notes explaining that criterion still won't work but at least it won't break due to this.

sunshowers:
This is now out as cargo-nextest 0.9.22 (binaries will be out shortly).

Issue #302, opened 2022-06-21
Test output strips ANSI colors


aaronabramov:
This is a great test runner! it cut the run time of our tests suite by about 10x, however i can't figure out how to configure colors in test output.

If i have a simple test that looks like:
```rust
#[test]
fn panic_with_ansi_colors() {
    panic!(
        "
        This is a panic!

        \u{1b}[1;31mThis line should be red\u{1b}[0m
        \u{1b}[2mThis line should be dimmed\u{1b}[0m

        "
    );
}
```

The colors in the terminal always get stripped from the output:

![Screen Shot 2022-06-21 at 2 53 24 PM](https://user-images.githubusercontent.com/940133/174886652-df09ff67-7b31-4aae-bb63-c34e92a998e0.png)

there's a `--color` option in the runner, but i assume it's for the runner itself and not the tests

aaronabramov:
ah.. interesting. this seems to be intentional https://github.com/nextest-rs/nextest/blob/main/nextest-runner/src/reporter.rs#L992

is there any way to overwrite this?

sunshowers:
Hi Aaron! Dug into the history and it looks like it was carried over from the initial implementation a long time ago. I think we can relax this as long as we output an ANSI reset at the very end.

aaronabramov:
This is so so so much better. thank you!

![Screen Shot 2022-06-21 at 4 41 53 PM](https://user-images.githubusercontent.com/940133/174902101-f062f092-29df-4334-aaf1-facff8da7970.png)

PR #303, opened 2022-06-21
Don't strip ANSI colors

aaronabramov:
Addressing https://github.com/nextest-rs/nextest/issues/302

Unfortunately sending the reset symbol won't fix the 
```
', src/lib.rs:3:5
```
part, but i think that's expected.

Screen Shot 2022-06-21 at 3 43 12 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1112,7 +1112,6 @@ dependencies = [
  "serde",
  "serde_json",
  "shell-words",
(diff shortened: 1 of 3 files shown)


sunshowers:
Thanks! Going in a slightly different direction -- #304.

aaronabramov:
oh. yes, that makes more sense! 😛

PR #343, opened 2022-07-08
Update coverage.yml

tbraun96:


sunshowers:
Thanks for the PR! Could you provide some more details about it? Also, what do we need to do to make the end-to-end upload work properly?

sunshowers:
I cleaned your PR up in #418 and we now have coverage data being uploaded to codecov.io. Thanks!

PR #361, opened 2022-07-16
Add Homebrew install instructions

rouge8:
I added cargo-nextest to Homebrew :)

### DIFF
--- a/site/src/book/pre-built-binaries.md
+++ b/site/src/book/pre-built-binaries.md
@@ -61,6 +61,14 @@ If you have [cargo-binstall](https://github.com/ryankurte/cargo-binstall) availa
 cargo binstall cargo-nextest --secure
 ```
 
+### Using Homebrew
+
+If you have [Homebrew](https://brew.sh/) available, you can install nextest with:
+
+```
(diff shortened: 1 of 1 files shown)


sunshowers:
Wow, thank you!

Are you going to maintain it as new versions come out? I'm a little worried about nextest falling behind.

sunshowers:
Also, thanks for using the default-no-update feature! This is exactly what it's for.

rouge8:
> Wow, thank you!
> 
> Are you going to maintain it as new versions come out? I'm a little worried about nextest falling behind.

Sure I can try!

sunshowers:
I think this should be on the "Installing from source" page because the Homebrew formula https://github.com/Homebrew/homebrew-core/pull/105954 (like most Homebrew formulas) does a source install.

rouge8:
The PR is a source install, but when they merge they add bottles so it’s installed from binaries.

sunshowers:
Oh right. That makes sense.

Thanks again, will merge later today.

sunshowers:
Thanks again!

Issue #392, opened 2022-07-23
Pre-built binary for ARM


teymour-aldridge:
Would it be possible build binaries for ARM (Linux)? I would be happy to submit a patch if this would be welcome.

sunshowers:
Hi! ARM binaries would absolutely be welcome. You can use https://github.com/sunshowers/test-binary-release as a playground if you need to.

sunshowers:
Released this as part of cargo-nextest 0.9.29 -- binaries will be available in the next 20 minutes.

PR #398, opened 2022-07-24
Build binaries for Linux aarch64 on CI.

teymour-aldridge:
Fixes https://github.com/nextest-rs/nextest/issues/392.

@sunshowers thanks for sharing the link to the playground, it was very helpful. I did experiment with it, and I _think_ this should build the binaries correctly, but I'm not 100% sure.


sunshowers:
Thanks, this looks good. I'll try it out with a pre-release some time today.

PR #399, opened 2022-07-24
Add x86_64-unknown-linux-musl binary release

### LINKED ISSUE #391: Pre-built binary release for musl?
Currently there is only x86_64-unknown-gnu binary release, it'd be nice to include a musl version.

I'd like to use it to test on Alpine Linux, see https://github.com/PyO3/maturin/pull/1004.

messense:
I can confirm that [it compiles for Linux musl target](https://github.com/messense/nextest/runs/7488680242?check_suite_focus=true).

Closes #391

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,6 +101,8 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


sunshowers:
Looks good! I'll do a beta release to make sure it works, hopefully some time today.

PR #401, opened 2022-07-24
[nextest-runner] parse CLI configs passed in via --config

### LINKED ISSUE #400: Test runner config isn't picked up from --config
cargo-miri very very recently started passing the test runner as a command line argument instead of an environment variable. It looks to me like nextest doesn't handle that.

I took a shot at adding CLI parsing here, but I can't come up with a clean way to do it (I suspect I just don't know this codebase). It seems like the easiest place is to add some logic here: https://github.com/nextest-rs/nextest/blob/a809da0ddefe71297e804558199d23bfed637d9d/nextest-runner/src/cargo_config.rs#L251-L270
but the code there assumes that configs have a path they came from, and of course `--config` is not a path. But I can confirm that if I `push` something to that `Vec` and make up a path, cargo-miri works again.

sunshowers:
Follow the same rules that Cargo does:
* configs form valid TOML files
* hierarchical config: env vars over CLI options
* configs followed from left to right

Tested against Miri past https://github.com/rust-lang/miri/pull/2428.

Closes #400.

cc @saethlin

### DIFF
--- a/cargo-nextest/src/cargo_cli.rs
+++ b/cargo-nextest/src/cargo_cli.rs
@@ -125,9 +125,11 @@ pub(crate) struct CargoOptions {
     #[clap(long, group = "cargo-opts")]
     offline: bool,
 
+    // NOTE: this does not conflict with reuse build opts since we let target.runner be specified
+    // this way
     /// Override a configuration value (unstable)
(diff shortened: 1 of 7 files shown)


saethlin:
Oh wow this is very thorough. Thanks!

PR #414, opened 2022-07-27
mention Miri's cross-interpretation feature

RalfJung:
The fact that Miri can do this still regularly catches people by surprise, so it seems worth advertising here, if you are okay with that. :)

I also added a link to explain what UB is.


sunshowers:
Looks good, thanks!

PR #465, opened 2022-08-12
fix(runner): cancel retry if cancel signal is received

skyzh:

Currently, a test will still be retried even if nextest receives a cancel signal. This PR fixes this behavior by not retrying after cancelled. To reproduce this issue, press Ctrl+C immediately after `cargo nextest run` with the following config and test crate:

nextest.toml:

```toml
[profile.default]
retries = 5
slow-timeout = { period = "5s" }
status-level = "all"
final-status-level = "slow"
```

lib.rs:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test_timeout() {
        println!("test");
        std::thread::sleep(std::time::Duration::from_secs(10));
        println!("test");
    }    
}
```

### DIFF
--- a/nextest-runner/src/runner.rs
+++ b/nextest-runner/src/runner.rs
@@ -291,7 +291,9 @@ impl<'a> TestRunnerInner<'a> {
                                     // The test succeeded.
                                     run_statuses.push(run_status);
                                     break;
-                                } else if attempt < total_attempts {
+                                } else if attempt < total_attempts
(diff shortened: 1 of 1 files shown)


sunshowers:
Good catch, thanks! Will get a new version out today.

sunshowers:
Binaries should be out within 20 minutes or so. Thanks again.

skyzh:
Thanks a lot! No hurry 😋🥰

PR #469, opened 2022-08-13
fix: respect test runner during build reuse

### LINKED ISSUE #468: Setting target runner when running from an archive
I am using cargo nextest to run cross compiled builds for `x86_64-pc-windows-msvc` using wine.
To achieve I have set `CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_RUNNER="wine"`.

Running without reusing builds works as expected: 

```
cargo nextest run --target x86_64-pc-windows-msvc
```

However, I want to run tests across multiple containers in parallel and reuse builds across them.
My workflow looks roughly like this:

```
cargo nextest archive --target x86_64-pc-windows-msvc --archive-file=foo.tar.zst
cargo nextest run --archive-file=foo.tar.zst
```

However this does currently not work because the second invocation does not respect `CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_RUNNER="wine"`.

I have tried digging into what is going on exactly but I am not quite sure.
I found that setting `CARGO_BUILD_TARGET=x86_64-pc-windows-msvc` is a workaround.
That would suggest the problem is simply that cargo nextest is using the wrong target for running the tests.
However specifying `--target` is not allowed when `--archive-file` is given.
Furthermore setting `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="wine"` (or the same for MUSL) does not seem to be respected either. From glancing at the sourcecode I am guessing without this manual flag the target-triple is set to `None` (because no build actually runs?) and therefore no runner is actually used.

To me the a solution seems to be one of:
* Allow passing `--target` to `cargo nextest run` even if `--archive` is given
* Save the target into the archive and respect it properly during `cargo nextest run`
* (Using some existing way to detect the correct target?)

If there is consensus on how to resolve this, I could send a PR if that would be desired although I am not super familiar with the codebase yet

pascalkuthe:
Fixes #468 

Previously `CARGO_TARGET_<target>_RUNNER` would be ignored if a build was
loaded from an archive file with --archive, because no/the incorrect target-triple was
detected in that case.

This patch changes the runner detection to account for those cases.
The target triple is now stored in the `BuildMetadata`.
If cargo is invoked the triple will be detected with the usual mechanism while
the `BuildMetadata` is generated.
During build-reuse the target triple is loaded from the `BuildMetadataSummary`.

If the target was not manually set, the host target-triple will be used
during serialization. This ensures correct behaviour in cases where
runner has a different default toolchain.

I have tested this locally with a docker container with cross compilation for windows and running with wine. I am not sure if this needs any tests. The changes are quite simple and testing this would probably be quite involved (because you need a working target runner that works for something that is not host).

I think change logs need to be adjusted but I am not sure how exactly this is handled for this project so I held off on that for now.

### DIFF
--- a/nextest-runner/src/cargo_config.rs
+++ b/nextest-runner/src/cargo_config.rs
@@ -165,7 +165,7 @@ impl fmt::Display for TargetTripleSource {
                 write!(f, "`build.target` within `{path}`")
             }
             Self::Metadata => {
-                write!(f, "--binaries-metadata option")
+                write!(f, "--archive-file or --binaries-metadata option")
             }
         }
     }
--- a/nextest-runner/src/cargo_config.rs
+++ b/nextest-runner/src/cargo_config.rs
@@ -11,6 +11,7 @@ use camino::{Utf8Path, Utf8PathBuf};
 use once_cell::sync::OnceCell;
 use serde::Deserialize;
 use std::{collections::BTreeMap, fmt};
+use target_spec::Platform;
 use toml_edit::Item;
 
 /// Represents a target triple that's being cross-compiled against.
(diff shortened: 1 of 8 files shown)


nextest-runner/src/cargo_config.rs
```diff
@@ -126,6 +164,9 @@ impl fmt::Display for TargetTripleSource {
             } => {
                 write!(f, "`build.target` within `{path}`")
             }
+            Self::Metadata => {
+                write!(f, "--binaries-metadata option")
```
sunshowers, line 168:
```suggestion
                write!(f, "--archive-file or --binaries-metadata options")
```

sunshowers (approved):
I think this looks great, thanks!

I agree that testing this is going to require infrastructure we don't yet have. Maybe this will be an incentive for me to write some (I think we'll want to build the tests on x86_64 linux targeting aarch64 linux, then use cross or something to run them).

PR #470, opened 2022-08-14
[WIP] [nextest-runner] add rudimentary support for handling SIGTSTP and SIGCONT

sunshowers:
Some interesting issues we've faced while developing this:

## Race in posix_spawn

Sometimes nextest just hangs if Ctrl-Z comes in at the wrong time. This is due to a fundamental race in posix_spawn: https://sourceware.org/pipermail/libc-help/2022-August/006263.html

To solve this issue, we use a double-spawn approach (with `NEXTEST_EXPERIMENTAL_DOUBLE_SPAWN=1` set in the environment and with Rust beta, currently):

1. Block SIGTSTP before calling posix_spawn.
2. Rather than spawning the test process directly, spawn `cargo-nextest __double-spawn <test-binary and args>`.
3. In the child process, unblock SIGTSTP before execing the child process.

## Pausing timers

We need a way to pause all our timers when we're pausing, then resume them afterwards. To do this, introduce a `PausableTimer` with `pause()` and `resume()` methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1248,6 +1248,7 @@ dependencies = [
  "once_cell",
  "owo-colors",
  "pathdiff",
+ "pin-project-lite",
  "pretty_assertions",
  "proptest",
  "proptest-derive",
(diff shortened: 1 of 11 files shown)


sunshowers:
Posted about the issue to the glibc mailing list: https://sourceware.org/pipermail/libc-help/2022-August/006263.html

PR #518, opened 2022-09-08
[doc] add [test-threads default] tip to run-help.txt

jiangying000:
It cost me some time to find the default value, maybe add a tip

### DIFF
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -42,7 +42,7 @@ final-status-level = "flaky"
 # Accepted values are
 # * "immediate": output failures as soon as they happen
 # * "final": output failures at the end of the test run
-# * "immediate-final": output failures as soon as they happen and at the end of
+# * "immediate-final": output failures as soon as they happen and at the end of, combination of "immediate" and "final"
 #   the test run
 # * "never": don't output failures at all
 #
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -42,8 +42,8 @@ final-status-level = "flaky"
 # Accepted values are
 # * "immediate": output failures as soon as they happen
 # * "final": output failures at the end of the test run
-# * "immediate-final": output failures as soon as they happen and at the end of, combination of "immediate" and "final"
-#   the test run
+# * "immediate-final": output failures as soon as they happen and at the end of
+#   the test run; combination of "immediate" and "final"
 # * "never": don't output failures at all
 #
 # For large test suites and CI it is generally useful to use "immediate-final".
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -599,7 +599,8 @@ pub struct TestRunnerOpts {
     #[clap(long)]
     no_run: bool,
 
-    /// Number of tests to run simultaneously [possible values: integer or "num-cpus"]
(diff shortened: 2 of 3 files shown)


sunshowers:
Thank you for the contribution! Unfortunately this isn't quite the right way to go about it: `run-help.txt` is generated from `cargo nextest run -h`, so the way to do this would be:

1. Add the hint to the help text here: https://github.com/nextest-rs/nextest/blob/188ab238cc71f25bb70fd36fede0b3ef82001bb7/cargo-nextest/src/dispatch.rs#L602
2. Run `cargo install --path cargo-nextest` to install this version of cargo-nextest
3. Run `./scripts/regenerate-help-text.sh`.

Thanks again :)

nextest-runner/src/reporter.rs
```diff
@@ -842,16 +842,17 @@ impl<'a> TestReporterImpl<'a> {
                                     false => self.failure_output,
                                 };
 
-                                if self.final_status_level >= final_status_level {
+                                // Print out the final status line so that status lines are shown
+                                // for tests that e.g. failed due to signals.
+                                if self.final_status_level >= final_status_level
+                                    || test_output_display.is_final()
+                                {
```
jiangying000, line 849: this diff looks strange, since the lines are all updated in nextest/main, and I thought shouldn't has the diff when I want to merge into *[nextest-rs:main](https://github.com/nextest-rs/nextest) from [jiangying000:main](https://github.com/jiangying000/nextest)*
https://github.com/nextest-rs/nextest/blob/main/nextest-runner/src/reporter.rs#L845-L850

jiangying000: The code in nextest-rs:main looks just the same as jiangying000:main, why is the diff ?

jiangying000:
> Thank you for the contribution! Unfortunately this isn't quite the right way to go about it: `run-help.txt` is generated from `cargo nextest run -h`, so the way to do this would be:
> 
> 1. Add the hint to the help text here: https://github.com/nextest-rs/nextest/blob/188ab238cc71f25bb70fd36fede0b3ef82001bb7/cargo-nextest/src/dispatch.rs#L602
> 2. Run `cargo install --path cargo-nextest` to install this version of cargo-nextest
> 3. Run `./scripts/regenerate-help-text.sh`.
> 
> Thanks again :)

thanks for detailed guidance.

sunshowers:
I'd recommend doing a rebase and force push rather than a merge.

sunshowers (changes requested):
Could you rebase and force push? Happy to land it after you do that. Currently I'm not sure what state it's in.

jiangying000:
yes, I will update the pr, sorry for my procrastination

nextest-runner/default-config.toml
```diff
@@ -43,7 +43,7 @@ final-status-level = "flaky"
 # * "immediate": output failures as soon as they happen
 # * "final": output failures at the end of the test run
 # * "immediate-final": output failures as soon as they happen and at the end of
-#   the test run
+#   the test run; combination of "immediate" and "final"
```
jiangying000, line 46: add `; combination of "immediate" and "final"` make the sentence more explicit

sunshowers:
I've fixed this up and landed it in https://github.com/nextest-rs/nextest/commit/ee12fb1aa4c664a3c5d04f3adaaad91d56204b9d. Thanks!

PR #534, opened 2022-09-21
Pass environment variables defined in `config.toml`

arxanas:


nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +241,23 @@ impl CargoConfigs {
...
+    pub fn env(&self) -> Result<BTreeMap<String, String>, CargoConfigError> {
+        let env = self
+            .discovered_configs()?
+            .filter_map(|config| match config {
+                DiscoveredConfig::CliOption { config, source: _ }
+                | DiscoveredConfig::File { config, source: _ } => Some(config),
+                DiscoveredConfig::Env => None,
+            })
+            .flat_map(|config| {
+                dbg!(config);
```
sunshowers, line 254: I assume you're going to remove this line?

nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +241,23 @@ impl CargoConfigs {
...
+    /// The environment variables to set when running Cargo commands.
+    pub fn env(&self) -> Result<BTreeMap<String, String>, CargoConfigError> {
+        let env = self
+            .discovered_configs()?
+            .filter_map(|config| match config {
+                DiscoveredConfig::CliOption { config, source: _ }
+                | DiscoveredConfig::File { config, source: _ } => Some(config),
+                DiscoveredConfig::Env => None,
+            })
+            .flat_map(|config| {
```
sunshowers, line 253: Hmm, how would precedence work here? I think if you want to follow this approach to building up the environment map, you'll want to add a `.rev()` here.

nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +241,23 @@ impl CargoConfigs {
         &self.cwd
     }
 
+    /// The environment variables to set when running Cargo commands.
+    pub fn env(&self) -> Result<BTreeMap<String, String>, CargoConfigError> {
```
sunshowers, line 245: Could you add a type alias `type EnvironmentMap = BTreeMap<String, String>`?

nextest-runner/src/cargo_config.rs
```diff
@@ -498,6 +515,8 @@ pub(crate) struct CargoConfig {
     #[serde(default)]
     pub(crate) build: CargoConfigBuild,
     pub(crate) target: Option<BTreeMap<String, CargoConfigRunner>>,
+    #[serde(default)]
+    pub(crate) env: BTreeMap<String, String>,
```
sunshowers, line 519: This needs to support the full spec, with `force` and `relative` (and also needs to include tests for all the possible combinations).

nextest-runner/src/list/test_list.rs
```diff
@@ -952,6 +965,8 @@ pub(crate) fn make_test_command(
         cmd.env(format!("NEXTEST_BIN_EXE_{}", name), &path);
     }
 
+    cmd.envs(env);
```
sunshowers, line 968: This would need to be expanded out once `force` is supported.

sunshowers (commented):
I think this is a great start! Left a few comments.

nextest-runner/src/cargo_config.rs
arxanas: How do you recommend that I structure the tests? I'm not sure how to write an integration test that verifies that the environment variables are being set.

arxanas:
@sunshowers can you approve the workflow runs?

nextest-runner/src/cargo_config.rs
sunshowers: That is a really good question! You'd do it through adding assertions to the `test_cargo_env_vars` test fixture: https://github.com/nextest-rs/nextest/blob/main/fixtures/nextest-tests/tests/basic.rs#L117. You can add an `[env]` section to https://github.com/nextest-rs/nextest/blob/main/fixtures/nextest-tests/.cargo/config to define environment variables.

So we'd have to introduce a bunch of assertions:
1. Plain env var that doesn't already exist: pick a name that no one's going to set like `__NEXTEST_PLAIN_ENV_DOES_NOT_EXIST` and set it to some sort of well-known value e.g. `plain-env-value`, then assert in the fixture that it's the same.
2. Plain env var that already exists: define a variable, and then also set it in https://github.com/nextest-rs/nextest/blob/main/nextest-runner/tests/integration/fixtures.rs#L200-L204 (rename `set_rustflags` to `set_env_vars` first) and https://github.com/nextest-rs/nextest/blob/515eef9f85a517a688dfde6946a8fcdbb4af9192/cargo-nextest/src/tests_integration/fixtures.rs#L131.
3. `force = true`: same as 2, except make sure it's set to the new value
4. `relative = true`
5. `relative = true, force = true`

sunshowers:
Just did! Unfortunately each time you update I'll have to re-review it. If you do another PR with a simple change (like the change to `set_env_vars` above) I can land that and then it'll be easier for you. Let me know what works.

arxanas:
@sunshowers I updated this PR with my latest changes. It looks like for the `basic` test, `CargoConfigs` is picking up the configs in `.cargo/config.toml` instead of `fixtures/nextest-tests/.cargo/config.toml`, which I guess makes sense, but doesn't help for testing. Should I be invoking the test fixture differently? Currently, I'm running

```sh
$ cargo local-nt run -- basic
```

Similarly, I don't see that the target triple was detected for the `basic` test, although it doesn't seem like there's any tests which depend on that.

If I do set it in `.cargo/config.toml` instead of inside the fixture config, it causes other problems: since nextest is invoking itself (IIUC?), the outermost invocation successfully picks up the environment variables, which means that the innermost invocation sees them in the environment as if they had been explicitly set by the user. It then thinks that the user meant to explicitly set those environment variables, so it declines to set any environment variables which aren't marked as `force`. It should actually not observe the variables in the environment and it should set all of them, except for the ones that were previously set in `set_env_vars`.

fixtures/nextest-tests/.cargo/config.toml
```diff
@@ -0,0 +1,7 @@
+# TODO: move into `config`.
+[env]
+__NEXTEST_ENV_VAR_FOR_TESTING_PURPOSES_1 = "testing-value-1"
+__NEXTEST_ENV_VAR_FOR_TESTING_PURPOSES_2 = "testing-value-2"
+__NEXTEST_ENV_VAR_FOR_TESTING_PURPOSES_3 = { value = "testing-value-3", force = true }
+__NEXTEST_ENV_VAR_FOR_TESTING_PURPOSES_4 = { value = "./testing-value-4", relative = true }
+__NEXTEST_ENV_VAR_FOR_TESTING_PURPOSES_5 = { value = "./testing-value-5", force = true, relative = true }
```
sunshowers, line 7: Could you give these variables more descriptive names?

nextest-runner/src/list/test_list.rs
```diff
@@ -59,6 +60,9 @@ pub struct RustTestArtifact<'g> {
     /// The working directory that this test should be executed in.
     pub cwd: Utf8PathBuf,
 
+    /// The environment variables to set when running the test.
+    pub env: EnvironmentMap,
```
sunshowers, line 64: I think this should be passed down as runtime context in `TestList::new`, similar to how the target runner is passed down. This is part of the runtime context, not build-time context.

nextest-runner/src/list/test_list.rs
```diff
@@ -811,7 +823,7 @@ impl<'a> TestInstance<'a> {
         test_list: &TestList<'_>,
         target_runner: &TargetRunner,
     ) -> std::process::Command {
-        let platform_runner = target_runner.for_build_platform(self.bin_info.build_platform);
+        let platform_runner = target_runner.for_build_platform(self.suite_info.build_platform);
```
sunshowers, line 826: I like this rename!

nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +242,54 @@ impl CargoConfigs {
         &self.cwd
     }
 
+    /// The environment variables to set when running Cargo commands.
+    pub fn env(&self) -> Result<EnvironmentMap, CargoConfigError> {
+        let env = self
+            .discovered_configs()?
+            .rev() // TODO: necessary? what order are the discovered configs?
```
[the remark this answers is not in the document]
arxanas: The intention was to handle Windows, where environment variables are case-insensitive. I added a comment to explain further. Let me know if I should adjust the comment or drop support for Windows.

arxanas:
@sunshowers Thanks for the feedback. I believe I addressed all the concerns.

I think the tests are passing, except for the integration tests which try to run `cargo nextest run` on a project in a different directory (i.e. set `--manifest-path`). As far as I can tell, those invocations *shouldn't* pick up the environment variables defined in the fixture's `config`, because only the current working directory and its ancestors are supposed to be searched:

> If, for example, Cargo were invoked in /projects/foo/bar/baz, then the following configuration files would be probed for and unified in this order:
> * /projects/foo/bar/baz/.cargo/config.toml
> * /projects/foo/bar/.cargo/config.toml
> * /projects/foo/.cargo/config.toml
> * /projects/.cargo/config.toml
> * /.cargo/config.toml
> * $CARGO_HOME/config.toml which defaults to:
>   * Windows: %USERPROFILE%\.cargo\config.toml

> ...
> At present, when being invoked from a workspace, Cargo does not read config files from crates within the workspace. i.e. if a workspace has two crates in it, named /projects/foo/bar/baz/mylib and /projects/foo/bar/baz/mybin, and there are Cargo configs at /projects/foo/bar/baz/mylib/.cargo/config.toml and /projects/foo/bar/baz/mybin/.cargo/config.toml, Cargo does not read those configuration files if it is invoked from the workspace root (/projects/foo/bar/baz/).

From https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure.

But I changed it so that the cargo search path uses the provided `--manifest-path`, and it works for most test cases. Let me know what you think we should do about this case. (It currently doesn't work for testing from an archive.)

sunshowers:
Thanks! Just saw this, apparently I missed your last message. Sorry about that!

> As far as I can tell, those invocations shouldn't pick up the environment variables defined in the fixture's config, because only the current working directory and its ancestors are supposed to be searched:

Yes, that's right. So if you can find a way to flag those runs (perhaps by setting an environment variable like `NOT_IN_WORKSPACE_CWD` or something), then in the test itself you can branch on that env var.

> I tried changing it so that the cargo search path uses the provided --manifest-path, and it works for most test cases

I think matching cargo's behavior is the right thing to do here, and cargo's behavior is that `--manifest-path` isn't considered for cargo search paths. So I'd suggest instead:

1. Going back to the old behavior of searching for configs in the cwd.
2. Flagging test runs with `--manifest-path` with the env var as above, and/or replacing some of those `--manifest-path` invocations with cding into the directory. (With nextest, each test runs in its own process, so you can even chdir the current process without worrying about trampling over other tests.)

nextest-runner/src/list/test_list.rs
```diff
@@ -882,6 +894,50 @@ pub(crate) fn make_test_command(
...
+        name,
+        value,
+        force,
+        relative,
+    } in env
+    {
+        let name_os_string = make_windows_compatible_key(OsString::from(name));
+        if !existing_keys.insert(name_os_string) && !force {
+            continue;
+        }
```
sunshowers, line 924: Is this the right order of things? What happens if there are multiple `[env]` vars defined in several config files, each with its own `force` setting?

My understanding is that the evaluation happens in two phases:

1. Configuration is all collected together, and lower-precedence configs are discarded.
2. Configuration is then applied.

The upshot of this is that if your cwd is `foo/bar`, and `foo/bar/.cargo/config.toml` is:

```toml
[env]
MY_VAR = { value = "foo-bar-config", force = true }
```

And `foo/.cargo/config.toml` is:

```toml
[env]
MY_VAR = { value = "foo-config", force = true }
```

Then Cargo will have `MY_VAR` set to `foo-bar-config`, whereas this code sets `MY_VAR` to `foo-config` if I've understood it correctly.

I'm not particularly sure what happens on Windows, though! It might be worth looking at Cargo's source code to get this right. I've had to do this several times for this code, and it's sometimes resulted in me filing bugs with upstream cargo (https://github.com/rust-lang/cargo/issues/10991, https://github.com/rust-lang/cargo/issues/10992).

nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +242,50 @@ impl CargoConfigs {
...
+            .flat_map(|(config, source)| {
+                let source = match source {
+                    CargoConfigSource::CliOption => None,
+                    CargoConfigSource::File(path) => Some(path.clone()),
+                };
+                config
+                    .env
+                    .clone()
+                    .into_iter()
+                    .map(move |(name, value)| (source.clone(), name, value))
```
sunshowers, line 263: Is this extra clone required?

nextest-runner/src/cargo_config.rs
```diff
@@ -241,6 +242,50 @@ impl CargoConfigs {
...
+        let env = self
+            .discovered_configs()?
+            .filter_map(|config| match config {
+                DiscoveredConfig::CliOption { config, source }
+                | DiscoveredConfig::File { config, source } => Some((config, source)),
+                DiscoveredConfig::Env => None,
+            })
+            .flat_map(|(config, source)| {
+                let source = match source {
+                    CargoConfigSource::CliOption => None,
```
sunshowers, line 256: Should test this as well.

sunshowers (approved):
Looks good, just a couple of comments that I'll address myself before landing.

nextest-runner/src/list/test_list.rs
```diff
@@ -676,6 +683,9 @@ impl<'g> RustTestArtifact<'g> {
             &self.package,
             dylib_path,
             &self.non_test_binaries,
+            // We don't need to set the user's custom environment variables defined in `config.toml`
+            // just to get the list of tests.
+            &Vec::new(),
```
sunshowers, line 688: This isn't quite correct -- custom test harnesses might depend on these variables.

nextest-runner/src/list/test_list.rs
```diff
@@ -882,6 +894,50 @@ pub(crate) fn make_test_command(
...
+        }
+
+        let value = if *relative {
+            let base_path = match source {
+                Some(source_path) => source_path,
+                None => panic!("Cannot use a relative path for environment variable {name:?} whose source is not a config file"),
+            };
+            match base_path.parent() {
+                Some(base_path) => base_path.join(value).into_string(),
+                None => value.clone(),
```
sunshowers, line 933: This needs one more level of `parent()`: https://doc.rust-lang.org/cargo/reference/config.html#env

> Setting the relative flag evaluates the value as a config-relative path *that is relative to the parent directory of the .cargo directory* that contains the config.toml file

Going to fix this.

nextest-runner/src/list/test_list.rs
sunshowers: So... as far as I can tell this still isn't resolved. I added a bunch of assertions (search through my updates for `__NEXTEST_TESTING_EXTRA_CONFIG`) and they don't seem to be behaving as expected unfortunately.

I've addressed most of the other issues but I think this one is big enough that I'm going to leave it to you.

sunshowers (changes requested):
I've done a number of updates and fixes myself, but the one I'm requesting you do is a more significant change.

nextest-runner/src/list/test_list.rs
arxanas: @sunshowers I tried implementing this, but ran into some trouble. What exactly is the specification here? When passing extra configs via CLI, they need to override environment variables set anywhere else? Currently, when providing a file in this manner, it becomes indistinguishable from a config file which was discovered normally:

https://github.com/nextest-rs/nextest/blob/ffd00c9b92de878dfc1e5cca16d24f76d9e120d2/nextest-runner/src/cargo_config.rs#L302

So I would have to start tracking that information, yes?

In any event, does it really make sense for the extra configs to always override other things, even if `force` wasn't set to `true`? If I run `FOO=1 cargo nextest run --config path/to/config.toml`, I would personally expect `FOO` to take precedence even if there was a `FOO` set in `config.toml` without `force = true`.

sunshowers: Hmm, I'm not sure what you mean here -- environment variables like `FOO=1` definitely should take precedence over config files that don't have `force = true` whether or not they're specified over the CLI. Config files over the CLI do take precedence over `.cargo/config` though (which should already be the case, you shouldn't have to touch it).

If you look at https://github.com/nextest-rs/nextest/pull/534/files#diff-b0286642fcb8773419c7282d5488e89e7fd9f2fe08c15c4cf5fe66c004f1bd24, all of the ones that don't have the environment variable set before execution (`__NEXTEST_TESTING_EXTRA_CONFIG_FORCE_IN_EXTRA` etc) should get set by `extra-config.toml`, and all of the ones that do have it set before execution (`__NEXTEST_TESTING_EXTRA_CONFIG_OVERRIDE_FORCE_IN_EXTRA` etc) should get set by `extra-config.toml` only if `force = true`. In all cases, `main-config.toml` should be completely ignored for these environment variables. (You can verify this yourself by setting the variables in the environment, then running `cargo test`.)

Does that make sense?

arxanas: > environment variables like FOO=1 definitely should take precedence over config files that don't have force = true whether or not they're specified over the CLI

Okay, I made an assumption about the environment variables based on the comment below, but it doesn't actually say what I thought it was, so this sounds makes sense.

I think I'm confused by this test case:

https://github.com/nextest-rs/nextest/blob/65977f9c8e948c5506e3e7f455b46458d9d76b31/fixtures/nextest-tests/.cargo/config#L28

According to my intuition, It seems like this should be `PASSED`, and we should take the value from the main config? (Actually, looking closer, neither the main config nor the extra config has a `PASSED` value...)

But according to the spec in this comment:

https://github.com/nextest-rs/nextest/blob/65977f9c8e948c5506e3e7f455b46458d9d76b31/fixtures/nextest-tests/.cargo/extra-config.toml#L4-L5

it should take on the value in the extra config, even though `force = true` in the main config.

Currently, it seems that there's no way to distinguish a cargo config that came from `config` versus `extra-config.toml`, so the only information we have to let us implement the specced behavior is the relative ordering of the config files. But, using the same rules as for environment variables and config files with different precedences discovered upwards in the directory tree, neither ordering would cause us to take on the value from the extra config file:

- If `config` is handled before `extra-config.toml`, then we set the value from the main config. When we see the value in `extra-config.toml` later, we don't use it, because we have already set a value.
- If `extra-config.toml` is handled before `config`, then we set the value from the extra config. But when we see the value in `config` later, we see that `force` is `true`, so we override it.

So in either case, the value from the main config takes precedence because `force` is `true`. I don't see an obvious way to keep the value from the extra config in this case. (Initially, I tried to check to see if the `source` is `CargoConfigSource::CliOption`, but this source is later lost when we read from a file rather than setting the config option directly on the command-line.)

It should be possible to preserve the information that a certain config value came from an extra config file (i.e. from a command-line option), but this isn't preserved currently IIUC, and implementing it seems to be a hassle, and I wonder if it would be the best behavior anyways.

sunshowers: Can we just keep track of which variables we've seen and ignored in another hashset? Then we could just consult that and ignore any further attempts to set those variables in configs with later precedence.

arxanas:
Re CI:

- The retry test case on macOS doesn't reproduce for me with `cargo local-nt run --no-default-features`. Is it flaky?
- It looks like there's a path-related test failing under Windows:

https://github.com/nextest-rs/nextest/blob/27f7f8d249048512b316866dfca97aa50e127271/fixtures/nextest-tests/tests/basic.rs#L165-L172

Do you understand why it's failing? (Unfortunately, I don't have a Windows machine to test on.)

sunshowers:
The macOS test is definitely flaky, I've been meaning to get to the bottom of it for a while.

The Windows issues are pretty fun, I know what's going on. The rest of the PR looks really good. Let me see if I can fix it up and land it shortly.

sunshowers (approved):
Looks fantastic, thanks for working through everything.

sunshowers:
I'm going to get this out in the next couple of days. Thanks again!

sunshowers:
BTW I ended up redoing [the `EnvironmentMap` construction](https://github.com/nextest-rs/nextest/commit/3cde3d0b2f9403c83e1ac77ebde4b23edcfa43b4) to fix some lingering issues including case-sensitivity on Windows.

PR #555, opened 2022-10-01
Implement backoff on retries

tomasol:
Implement fixed and exponential backoffs between retries. 
Sample configuration:
```toml
retries = { count = 5, backoff = "exponential", delay = "1s", max-delay = "10s", jitter = true }
```
The code was heavily inspired by the [again crate](https://github.com/softprops/again/), but the [iterator](https://github.com/softprops/again/blob/6db8c5e56bf93b1177ed35a926f8ff7d4adbf93b/src/lib.rs#L181) is not public so I had to copy it.
There are no tests yet and no good documentation.
Reporting: adds the delay as follows: `1/3 RETRY in 1s`.
The jitter currently picks a random number from the interval (0, calculated_delay), which might not be ideal.


sunshowers:
Thanks for doing this! Apologies for the delay, getting around to reviewing this now.

nextest-runner/default-config.toml
```diff
@@ -12,6 +12,10 @@ dir = "target/nextest"
 # "retries" defines the number of times a test should be retried. If set to a
 # non-zero value, tests that succeed on a subsequent attempt will be marked as
 # non-flaky. Can be overridden through the `--retries` option.
+# Examples
+# * retries = 3
+# * retries = { backoff = fixed, count = 2, delay: "1s" }
+# * retries = { backoff = exponential, count = 10, delay: "1s", jitter: true, max_delay: "10s" }
```
sunshowers, line 18: This needs to be in the TOML format:

`backoff = "fixed", delay = "1s", jitter = true`, etc.

nextest-runner/src/reporter.rs
```diff
@@ -686,10 +686,15 @@ impl<'a> TestReporterImpl<'a> {
...
             } => {
                 if self.status_level >= StatusLevel::Retry {
-                    let retry_string =
-                        format!("{}/{} RETRY", run_status.attempt, run_status.total_attempts);
+                    let retry_string = format!(
+                        "{}/{} RETRY in {}",
+                        run_status.attempt,
+                        run_status.total_attempts,
+                        humantime::Duration::from(*delay_before_next_attempt)
+                    );
```
sunshowers, line 697: So the issue here is that there's a limit of 12 characters here, and something like "1/3 RETRY in 1s" exceeds that limit.

I've been thinking about this and I think the way I'd probably want to display this is, if there is a delay:

```
TRY 1/3 FAIL [<duration>]  <binary name> <test name>
       DELAY [ in 1.000s]  <binary name> <test name>
```

And then, when the next attempt is started:

```
   2/3 RETRY               <binary name> <test name>
```

If there is no delay, then, the `DELAY` line is skipped. for consistency:

```
TRY 1/3 FAIL [<duration>]  <binary name> <test name>
   2/3 RETRY               <binary name> <test name>
```

For its implementation, this would also need `TestEvent` to be extended, and specifically `TestRetry` to be split into two separate events:

* an attempt just failed, will retry in the future with maybe a delay
* the next attempt has started.

nextest-runner/src/runner.rs
```diff
@@ -295,11 +370,19 @@ impl<'a> TestRunnerInner<'a> {
...
+                                        .expect("backoff delay must be non-empty");
                                     let _ = this_run_sender.send(InternalTestEvent::Retry {
                                         test_instance,
                                         run_status: run_status.clone(),
+                                        delay_before_next_attempt: delay,
                                     });
                                     run_statuses.push(run_status);
+                                    select! {
+                                    _ = sleep(delay) => {}
+                                    _ = cancellation_receiver.recv() => {}
```
sunshowers, line 384: Could you indent these one step?

sunshowers (commented):
I think this looks really good! Just a few comments.

sunshowers:
I'll finish this up myself, thanks!

sunshowers:
OK, I've done a bunch of improvements here. What's left here is:

1. Switch `RetryPolicy` to being an enum with `Fixed { count, delay, jitter }` and `Exponential { count, delay, jitter, max-delay }`. This prevents `max-delay` from being specified for `Exponential`, where it makes less sense. We also want to prevent `Exponential` with a zero delay.
2. Add tests for parsing invalid values.

I can do this some time tomorrow if you don't get to it by then.

tomasol:
Thanks for working on this PR! Unfortunately I won't be able to continue until next week.
What do you think about the `jitter` function? Currently it trims the delay to `(0 .. delay)`, maybe `(delay * 0.5 .. delay)` would be more appropriate? 
Two issues I found after submitting this PR:
* the code first calls `jitter` and then trims the result to `max_delay`. This should be swapped. This is safe since `jitter` can only make the value smaller.
* `current_factor` keeps getting increased even when `max_delay` has been already reached. This leads to an overflow.

sunshowers:
Thanks for catching those! I fixed up the issues you mentioned as well. I think this is ready to go now.

Thanks again for working on this.

sunshowers:
This is now out as part of cargo-nextest 0.9.38 :)

tomasol:
Awesome :+1:

PR #559, opened 2022-10-04
Optimization: Collect output in parallel in `TestRunner::run_test_inner`

NobodyXu:
* Add new dep tokio-util v0.7.4 to nextest-runner
* Add new dep scopeguard v1.1.0 to nextest-runner
* Create new type `ReadTask` for spawning the read tasks and handles the cancellation.
* Do not wrap `child_std{out, err}` in `BufReader`  since we read the output into a
  `BytesMut` that is at least 4096 large, having an internal buffer here won't help and
   would only incur additional copies.

This moves collection of output to a separate tokio task instead of doing it in the `tokio::select!` loop, which can be the bottleneck since `tokio::select!` is quite complex.

This also simplify the implementation by replacing the second `tokio::select!` loop with a spawned job.

All of the `*_done` boolean flags is also removed.

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1157,6 +1157,7 @@ dependencies = [
  "test-case",
  "thiserror",
  "tokio",
+ "tokio-util",
  "toml_edit",
  "twox-hash",
  "uuid",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1144,6 +1144,7 @@ dependencies = [
  "proptest-derive",
  "quick-junit",
  "regex",
+ "scopeguard",
  "self_update",
  "semver",
  "serde",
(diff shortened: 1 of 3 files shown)


sunshowers:
Thank you so much for your contribution!

After looking at it in detail over a few hours, I'm respectfully going to have to unfortunately decline most of it. I really appreciate the effort put in! However, in my performance profiling, reading from stdout/stderr has just never been the bottleneck compared to things like process spawning. This adds a couple extra deps and a significant amount of complexity to a process that's pretty straightforward today, even if it isn't optimal.

The removal of `BufReader` is an easy change, though, and if you want to create a PR with that change I'd be happy to land it.

Thanks again!

sunshowers (changes requested):
Requesting changes as mentioned in previous post.

NobodyXu:
Got it, I will open another PR that removes `BufReader`.

NobodyXu:
> The removal of BufReader is an easy change, though, and if you want to create a PR with that change I'd be happy to land it.

I've opened a new PR #560

PR #563, opened 2022-10-07
Refactor: Simplify `select!` code and replace confusing `*_done` `boolean`s with `FusedFuture`

NobodyXu:
Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/nextest-runner/src/runner.rs
+++ b/nextest-runner/src/runner.rs
@@ -700,18 +700,14 @@ impl<'a> TestRunnerInner<'a> {
             let leaked = if collect_output_fut.is_terminated() {
                 false
             } else {
-                let sleep = tokio::time::sleep(leak_timeout);
-
-                tokio::select! {
-                    res = &mut collect_output_fut => {
+                match tokio::time::timeout(leak_timeout, collect_output_fut).await {
+                    Ok(res) => {
                         res?;
                         false
                     }
-                    () = sleep => {
-                        // stdout and/or stderr haven't completed yet. In this case,
-                        // and mark this as leaked.
-                        true
-                    }
+                    // stdout and/or stderr haven't completed yet. In this case,
+                    // and mark this as leaked.
+                    Err(_) => true,
                 }
             };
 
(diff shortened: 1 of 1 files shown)


sunshowers:
Thanks! I don't think this is quite correct—we definitely need to loop until either the timer is hot or stdout and stderr are both closed

sunshowers:
I'm also honestly not a fan of the fused future in this case, since you have to check if they're terminated anyway it just adds verbosity.

NobodyXu:
> Thanks! I don't think this is quite correct—we definitely need to loop until either the timer is hit or stdout and stderr are both closed.

I created a new future `collect_output_fut` which `try_join` both `stdout_fut` and `stderr_fut`, so that we don't need to poll for both variable.

Since we only need to poll `collect_output_fut` and `sleep`in the second `select!`, there's no need for the loop.
We simply let them race to completion and pick the one that completes first and return `false` or `true` correspondingly.

NobodyXu:
> I'm also honestly not a fan of the fused future in this case, since you have to check if they're terminated anyway it just adds verbosity.

I would argue it is still an improvement.

Before that, we would check `*_done` in every `select!` statement, while now we only checked it once before the second `select!`.

Also, using fused future in select is actually the easy way to go and is more readable if you want to poll the same future multiple times, I read somewhere that they actually argue that all futures used in `select!` should be fused or print out an error.

With `fused`, you know immediately that this branch is simply ignored once it is completed and would not panic at runtime or prohibit some UBs.
With boolean flags, you need to actually think about them a while before realizing what it's doing.

NobodyXu:
Actually the second `select!` can be completely replaced by [`tokio::time::Timeout`](https://docs.rs/tokio/latest/tokio/time/struct.Timeout.html) which is more ergonomic than `select!`.

NobodyXu:
> Actually the second `select!` can be completely replaced by [`tokio::time::Timeout`](https://docs.rs/tokio/latest/tokio/time/struct.Timeout.html) which is more ergonomic than `select!`.

Done!

Now it is even more simpler.

sunshowers:
Thanks but I'm afraid I'm going to have to reject this again. I actually have some WIP code that needs the second select to be as it is, since it also selects across a channel in addition to the existing futures.

As I said I'm still not a fan of FusedFuture in cases where the code behaves differently based on whether the future has terminated. In the end it's a matter of taste, and as the maintainer I like the current approach more. Unless there's a clear correctness or performance issue, I think minor refactoring like this is just not that useful.

NobodyXu:
Got it, I will close this PR.

sunshowers:
BTW this thread and specifically this comment might be interesting: https://users.rust-lang.org/t/why-doesnt-tokio-select-require-fusedfuture/46975/13

PR #597, opened 2022-10-20
Update test-coverage.md

taiki-e:
The GitHub Actions example in test-coverage.md currently uses nightly Rust, but cargo-llvm-cov works on stable Rust since Rust 1.61 which instrument-coverage stabilized.


sunshowers:
Thanks!

Issue #622, opened 2022-11-04
profile with retries = 0 doesn't work


xxchan:
```toml
[profile.default]
retries = 5

[profile.ci]
retries = 0
```

`cargo nextest run -P ci` will retry 5 times instead of 0 times. In an older version `0.9.23` (quite old...) it works correctly.

xxchan:
`0.9.38` is the first version where this issue occurs.

PR #623, opened 2022-11-04
fix: handle profile with 0 retries

xxchan:
fix https://github.com/nextest-rs/nextest/issues/622


sunshowers:
Aha, thank you so much for the fix! Planning to get it out there today, hopefully.

sunshowers:
This is now out as cargo-nextest 0.9.43.

PR #626, opened 2022-11-06
Update uuid version in nextest-tests

erdnaxe:
This patch proposes to use the same uuid crate version between nextest crates and fixture crate.
It also removes a warning about a double declaration of license.


sunshowers:
Thanks! The first commit is great -- the second one is actually deliberate, we use that to test environment variables. I'll revert it locally and add a comment explaining this.

sunshowers:
Actually, there's already a comment explaining that. I'll just land the first commit then. Thanks again for your contribution!

PR #680, opened 2022-11-29
[cargo-nextest] remove all NEXTEST_ env vars from argument parsing test

sunshowers:
cargo llvm-cov started setting NEXTEST_TEST_THREADS=1, which breaks this
test.


taiki-e:
Btw, cargo-llvm-cov no longer sets this env var since 0.5.20 (https://github.com/taiki-e/cargo-llvm-cov/pull/279, 2023-06).

PR #681, opened 2022-11-29
cli: support --example EXAMPLE and --examples

jedbrown:
These are cargo test options to run unit tests that live in examples.


sunshowers:
Looks good, thanks!

PR #700, opened 2022-12-09
fix: missing graph_with_deps for run

### LINKED ISSUE #699: List and run command doesn't match on execution
Hello,

I'm working with a project using multiple crates in a single workspace structure.
Some crates are top level crate that use other crate as dependencies.
I tried to make the `filter-expr` work to execute tests of a particular branch of crate but without success.
I scratch my head to understand how the `filters` were working and I'm sorry if this issue end up being useless if I miss something in the doc.

My issue is that the `cargo nextest list -E 'deps(my-deps)' --workspace` is giving me a list of tests that I want to see executed. The `list` command show (for me) the expected output.

Using the same filter on the `run` command do nothing, it doesn't execute any tests (flagged `0` tests across X `binaries`).

Thanks for reading !

Freyskeyd:
This PR aims to fix #699 

It solved my issue but I don't know if it's expected to work that way...

Let me know !

Signed-off-by: Freyskeyd <simon.paitrault@gmail.com>

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -145,7 +145,7 @@ impl AppOpts {
                     cargo_options,
                     self.config_opts,
                     self.manifest_path,
-                    false,
+                    build_filter_needs_deps(&build_filter),
                     output_writer,
                 )?;
                 let app = App::new(base, build_filter)?;


sunshowers:
Yeah, that's definitely wrong. Thank you!

PR #719, opened 2022-12-21
Doc improvement: junit configuration

jesjos:
This had me confused b/c I tried adding the profile to my Cargo.toml instead of my nextest.toml


sunshowers:
Thanks!

PR #759, opened 2023-01-11
Update install-action to v2 and fix docs

taiki-e:
- Update install-action to v2 (Only docs need to be updated since the `@<tool_name>` shorthand is used in `.github/workflows`.)
- Fix docs on specifying version in install-action. The current docs mentions `version: ` input option, but such an option has never been supported by install-action, and the correct syntax is the `<tool_name>@<version>` syntax in the `tool:` input option.)

### DIFF
--- a/site/src/book/pre-built-binaries.md
+++ b/site/src/book/pre-built-binaries.md
@@ -144,13 +144,13 @@ The action will download pre-built binaries from the URL above and add them to `
 To install a version series or specific version, use this instead:
 
 ```yml
-- uses: taiki-e/install-action@v1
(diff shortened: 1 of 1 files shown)


taiki-e:
Failure of coverage is another regression from https://github.com/taiki-e/cargo-llvm-cov/pull/237.
I fixed it in cargo-llvm-cov 0.5.7 (https://github.com/taiki-e/cargo-llvm-cov/compare/v0.5.6...v0.5.7), so let's retry CI.

sunshowers:
Thank you!

PR #768, opened 2023-01-13
Fix release workflow failure

taiki-e:
This fixes release workflow failure: https://github.com/nextest-rs/nextest/actions/runs/3908669436/jobs/6679052947

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,8 +201,6 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -201,8 +201,6 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: ${{ matrix.target }}
```
taiki-e, line 205: upload-rust-binary-action internally calls `rustup target add`, so this is unneeded.

https://github.com/taiki-e/upload-rust-binary-action/blob/a9a2c3d8f238f311a925a074bd6ef21f00115960/main.sh#L126-L129

sunshowers:
Ahh, this is a much better fix than mine. Thank you!

Issue #846, opened 2023-04-02
nextest-metadata is outdated (`--format`)


kdy1:
```rust
let tests = ListCommand::new()
            .exec()
            .context("failed to get the list of tests using nextest")?;
```

The code above fails because `ListCommand` specifies `--format`

https://github.com/nextest-rs/nextest/blob/651bcaed8378b4daa4a36ac5782f050cc55e10dc/nextest-metadata/src/test_list.rs#LL83

sunshowers:
Thank you for filing this! That should be `--message-format`.

Would you like to submit a PR with a test? If not I can try getting to it some time this weekend.

PR #848, opened 2023-04-05
metadata: Use `--mesage-format` instead of  `--format`

kdy1:
Closes https://github.com/nextest-rs/nextest/issues/846


sunshowers:
Thank you so much!

PR #849, opened 2023-04-08
Update `nom-tracable` from 0.8.0 to 0.9.0

novedevo:
The old version, from two years ago, depends on an outdated version (5.1.2) of `nom` that generates a [future incompatibility report](https://github.com/rust-lang/rust/issues/79813), also visible in [`nextest`'s CI](https://github.com/nextest-rs/nextest/actions/runs/4611847684/jobs/8152051562#check-step-8). I also ran `cargo update` to bring all dependencies specified in `Cargo.lock` up to their most recent compatible versions, just for good measure. These are in separate commits, so you can cherry-pick just the `nom-tracable` update if you don't want the other dependency updates. `cargo clippy` and `cargo nextest run --all-features` both pass.

Thank you for all your work on `nextest`!


sunshowers:
Thank you!

PR #857, opened 2023-04-21
Support [profile.default-llvm-cov]

def-:
Depends on https://github.com/taiki-e/cargo-llvm-cov/pull/259

This is done similarly for miri and makes sense since both increase running times.


sunshowers:
Hmm, wouldn't this need to also include the settings in `default-config.toml`?

Issue #867, opened 2023-05-05
`--archive-file` doesn't seem to remember `--cargo-profile`


TriplEight:
Also `--archive-file` doesn't allow `--cargo-profile` flag, and it can't be mentioned in `nextest.toml` to use it with `--config-file` so I don't know what to do.

Setup:
The first job generates the nextest archive using a custom `--cargo-profile testnet`
```
cargo nextest archive --color always --locked --workspace --cargo-profile testnet --config-file scripts/nextest.toml --features=runtime-benchmarks,runtime-metrics --archive-file $TESTS_ARCHIVE
```

The second job reads a generated archive and uses `--partition` to run the tests
```
cargo nextest --color always --config-file scripts/ci/gitlab/nextest.toml --archive-file $TESTS_ARCHIVE --partition count:${CI_NODE_INDEX}/${CI_NODE_TOTAL}
```

It runs the most of the tests normally. But we have some benches that are cargo-target-specific and can only success with this profile. Instead, these tests run until job's timeout.

Everything runs smoothly without `--archive-file` in one job.

TriplEight:
added some docs. 
But I'm not yet positive that cargo-metadata includes  cargo profile - my tests are still failing and taking much longer and I'm not sure why.

sunshowers:
Hi, and thanks for the report.

The saved metadata doesn't explicitly include the cargo profile, but it does implicitly have that information because a run from a saved archive file doesn't invoke cargo at any point. As far as I know, cargo profiles only affect build-time settings, so I'm not sure why nextest would cause different behavior between using an archive and not using one. (Your tests might assume something different, though.)

Some comments:

1. Have you tried reproducing this locally? If so, it might be worth inspecting your test binaries to see what's going on.
2. Could you paste the contents of your custom cargo profile?

TriplEight:
1. Yes, it's reproducing.
2. here:
  ```toml
  [profile.release]
  panic = "unwind"
  opt-level = 3
  
  [profile.custom]
  inherits = "release"
  debug = 1
  debug-assertions = true
  overflow-checks = true
  ```

PR #870, opened 2023-05-05
Elaborate why one should use compilation options on `cargo nextest archive` and not while reading `--archive-file`

TriplEight:
This could have saved me a lot of time.

I was getting the errors like (cc https://github.com/nextest-rs/nextest/issues/867)

<details>

```
error: the argument '--archive-file <PATH>' cannot be used with:
   --lib
   --bin <BIN>
   --bins
   --example <EXAMPLE>
   --examples
   --test <TEST>
   --tests
   --bench <BENCH>
   --benches
   --all-targets
   --package <PACKAGES>
   --workspace
   --exclude <EXCLUDE>
   --all
   --release
   --cargo-profile <NAME>
   --build-jobs <JOBS>
   --features <FEATURES>
   --all-features
   --no-default-features
   --target <TRIPLE>
   --target-dir <DIR>
   --ignore-rust-version
   --unit-graph
   --future-incompat-report
   --frozen
   --locked
   --offline
   -Z <FLAG>
```

</details>

And couldn't understand why.

It turns out that compilation options are kept in the metadata (which could've been mentioned in the above error) and are not expected when the archive is being read.


sunshowers:
Hi there, missed this. Hoping to review it this weekend, thanks!

sunshowers:
Actually, this looks great. Thanks so much for the contribution!

PR #898, opened 2023-07-06
[quick-junit] Add support for test case properties

### LINKED ISSUE #897: Add a way to include custom properties into junit report
Hello, 

I would like to know if it is possible to include custom properties into junit report for each test ?

This could be great in order to mimic functionalities provided by [pytest-jira-xray ](https://pypi.org/project/pytest-jira-xray/). 

Thanks

skycoop:
Fixes #897

Simple change adding support for test case properties. These properties are used by some systems for metadata like attachments, links, and test steps, so are useful to be able to set.

### DIFF
--- a/quick-junit/src/report.rs
+++ b/quick-junit/src/report.rs
@@ -301,6 +301,9 @@ pub struct TestCase {
 
     /// Other fields that may be set as attributes, such as "classname".
     pub extra: IndexMap<String, String>,
+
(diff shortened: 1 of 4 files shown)


sunshowers (approved):
This is awesome, thanks!

(Non-breaking change because I remembered to mark `TestCase` as `#[non_exhaustive]`, yay)

PR #956, opened 2023-09-11
Add flag `--cargo-quiet`

ggwpez:
Change:
- Add `--cargo-quiet` flag to `nextest run` to suppress standard build output. Warnings and errors are still printed.

This makes it possible to run it with less noise. LMK if anything is missing.

Does this file need to be updated? Dont know how that is used https://github.com/ggwpez/nextest/blob/71ebdf65fbe24a64382524ae01e1ab57804f111e/site/help-text/list-help.txt#L31

### DIFF
--- a/cargo-nextest/src/cargo_cli.rs
+++ b/cargo-nextest/src/cargo_cli.rs
@@ -82,6 +82,10 @@ pub(crate) struct CargoOptions {
     #[arg(long, value_name = "NAME", group = "cargo-opts")]
     cargo_profile: Option<String>,
 
+    /// Run cargo in quiet mode
+    #[arg(long, group = "cargo-opts")]
+    cargo_quiet: bool,
+
     /// Number of build jobs to run
     #[arg(long, value_name = "JOBS", group = "cargo-opts")]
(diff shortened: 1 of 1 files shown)


sunshowers:
Hi, and thanks for the contribution!

I'm wondering, would it be okay to use `CARGO_TERM_QUIET=true` instead? I'm a little hesitant to add this option because I'd like to have some sort of unified model of quietness (and verbosity) between cargo and nextest.

ggwpez:
>I'm a little hesitant to add this option because I'd like to have some sort of unified model of quietness (and verbosity) between cargo and nextest.

Yes that is why i named it `--cargo-quiet` instead of `--quiet`; to still allow for a global `quiet`.

>I'm wondering, would it be okay to use CARGO_TERM_QUIET=true instead?

Yup that works. So feel free to close this. Will do myself if stale in two weeks. 
 
Thanks for your time!

sunshowers:
Decided to take this. I think this is simple enough and if we do end up unifying nextest and cargo, we can keep supporting this alternative approach as well.

Issue #942, opened 2023-08-28
Contributing features back to cargo?


rex-remind101:
This is very cool, but is there a reason these features aren't added directly back to `cargo test`? If there is a reason could it be documented?

sunshowers:
Hi there, thanks for the report. There are a few reasons nextest is a separate project:

* As documented at https://nexte.st/book/how-it-works.html, nextest has a significantly thicker interface. cargo test cannot just change how it works without breaking backwards compatibility, while nextest did not have this constraint at the time it was created.
* It's easier to experiment with improvements without having to worry about the long-term stability guarantees provided by Cargo, or go through the (necessarily) heavyweight Rust RFC process.
* Ultimately, as mentioned in the readme, nextest is a gift. I only have limited time to work on it these days and it's a far more efficient use of my time to maintain nextest than to try and port the changes over to cargo test.

With all that said, there _is_ currently [an ongoing effort](https://epage.github.io/blog/2023/06/iterating-on-test/) to add parts of nextest to cargo test. I'd love to see how that plays out over time. However I expect that nextest will always have a role as a place to experiment with UX and workflow improvements.

I'd welcome a PR which captures this information, probably added to the bottom of https://nexte.st/book/how-it-works.html. Would you like to submit this PR?

rex-remind101:
Thanks for the thoughtful explanation! I can add a pr once the weekend arrives.

rex-remind101:
https://github.com/nextest-rs/nextest/pull/963 lmk if this works :)

sunshowers:
Thanks! Closing since this is now at https://nexte.st/book/how-it-works.html#contributing-features-back-to-cargo.

PR #963, opened 2023-09-18
added section to how-it-works on contributing back to cargo

rex-remind101:
addresses https://github.com/nextest-rs/nextest/issues/942


sunshowers:
Thanks! Did some editing but otherwise this looks great.

Issue #931, opened 2023-08-13
Configurartion for a graceful shutdown period for tests manually aborted with Ctrl+C or other signals


Veetaha:
I have a use case where my tests have a really heavy setup and teardown logic. This logic happens within the test functions themselves. Each test function individually runs a `terraform apply` at the begging of the tests and `terraform destroy` deploying various resources to the AWS cloud temporarily.

`terraform` and AWS cloud are quite heavy-weight. It may take the from 1 to 10 minutes to perform the setup and maybe 1-5 minutes to teardown and cleanup the resources.

Right now `cargo-nextest` sends a `SIGKILL` to the test process leaving it no chance to gracefully terminate. It would solve this problem if `cargo-nextest` could allow configuring the behavior of graceful shutdown for tests. For example, if I could say in the config file that "these tests" selected by the DSL (in my case I don't need it, I need to select them all, but I think it may be useful for other use cases) need to be shutdown gracefully. Then configure the UNIX signal that will be sent to the tests, and the timeout for the tests to exit by themselves, after which `cargo nextests` will send a SIGKILL.

sunshowers:
Hi there -- please see https://nexte.st/book/slow-tests.html#how-nextest-terminates-tests. On encountering a timeout, nextest sends a SIGTERM, waits 10 seconds, and then sends a SIGKILL. You can customize the timeout with the `grace-period` setting.

The initial signal is not configurable, but I hope SIGTERM would work for you. If it doesn't work for you could you explain why? If there's a solid use case for a non-SIGTERM signal I'd be happy to accept patch contributions.

Veetaha:
I think I missed to mention the main thing in the issue. The problem here is not with the tests that time out. The problem is with the following scenario.

Given I already have tests in the process of running.
I suddenly want to abort the tests.
I press <kbd>Ctrl+C</kbd> on the terminal.
I expect that my tests receive a SIGINT and they gracefully shut down themselves and perform cleanup of the resources they created for themselves.

Right now `cargo-nextest` always sends `SIGKILL` to the tests when I abort it with <kbd>Ctrl+C</kbd>, but I want `cargo-nextest` to send `SIGINT` to my tests instead.

[...]

PR #1039, opened 2023-10-15
generalize slow-timeout.grace-period to all shutdown events

kallisti-dev:
Implementation for #931

Removes `slow-timeout.grace-period` and replaces it with `shutdown-grace-period` at the top-level config. The new config value applies to all shutdown events.

### DIFF
--- a/nextest-runner/src/runner.rs
+++ b/nextest-runner/src/runner.rs
@@ -803,7 +803,7 @@ impl<'a> TestRunnerInner<'a> {
                             // attempt to terminate the slow test.
                             // as there is a race between shutting down a slow test and its own completion
                             // we silently ignore errors to avoid printing false warnings.
-                            imp::terminate_child(&mut child, TerminateMode::Timeout(slow_timeout.grace_period), forward_receiver, job.as_ref()).await;
+                            imp::terminate_child(&mut child, TerminateMode::Timeout, forward_receiver, job.as_ref(), slow_timeout.grace_period).await;
                             status = Some(ExecutionResult::Timeout);
                             if slow_timeout.grace_period.is_zero() {
                                 break child.wait().await;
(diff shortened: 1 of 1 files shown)


kallisti-dev:
Should we mention the new behavior on ctrl+c and other shutdown events somewhere in the book? I wasn't sure where would be an appropriate place to mention the added behavior.

sunshowers:
Thanks. I'll get to reviewing this next week, but a quick comment is that we cannot break compatibility. If slow-timeout.grace-period is mentioned, it shouldn't be ignored.

A thing that would probably be reasonable to do is to preferentially use shutdown-grace-period but still support the old value.

sunshowers:
Another option is to just use slow-timeout.grace-period for shutdown grace periods as well. I think that might end up being simpler overall.

kallisti-dev:
> Thanks. I'll get to reviewing this next week, but a quick comment is that we cannot break compatibility. If slow-timeout.grace-period is mentioned, it shouldn't be ignored.

The new docs don't mention it anywhere, but I agree that compatibility should be maintained if possible.
 
> A thing that would probably be reasonable to do is to preferentially use shutdown-grace-period but still support the old value.

My initial thought was to have `slow-timeout.grace-period` overlap on top of `shutdown-grace-period` and specifically only handle the timeout events instead of all shutdown events, thus preserving the old behavior, but this seemed unnecessarily complicated and could be difficult to explain in documentation. Having them being aliases of each other for compatibility, with precedence for the new value, makes sense and is a simple change.

sunshowers:
> Much simpler from the perspective of an existing user, but it does hide the fact that the grace-period affects more than just timeout events.

Personally I think that's fine to be honest. I don't really see a distinction between a test shutting down because of a timeout and a test shutting down because a user hit ctrl-c.

Definitely aiming for simplicity in this case.

kallisti-dev:
Sounds good. I'll update the PR to go back to using slow-timeout.grace-period

kallisti-dev:
Updated

sunshowers:
Looks good, thanks! I'll get this out within the next 1-2 days.

sunshowers:
Released as part of nextest 0.9.61.

PR #1054, opened 2023-10-22
Strip ANSI escapes from JUnit output

MaienM:
Previously the escape character itself was removed, but the rest of the escape sequence was left in place, resulting in useless and seemingly random characters in the output. (Unless some tools expect this and know how to handle it, but that isn't the case for the tools I use.)

There was already a test that asserted this. I assumed the goal of that test was just to validate that illegal characters (like the escape character) didn't make it into the output, and I've adjusted it to verify that the entire sequence is stripped out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1995,6 +1995,7 @@ dependencies = [
  "nextest-workspace-hack",
  "owo-colors",
  "quick-xml 0.30.0",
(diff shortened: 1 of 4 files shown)


sunshowers:
Thank you! This was most likely an oversight on my end.

sunshowers:
Released in nextest 0.9.61.

PR #1073, opened 2023-10-27
Use bare minimum chrono features

littledivy:
Hey, we are using quick-junit in Deno. I'm trying to get rid of dynamic library that cost startup time. It turns out `chrono`'s "clock" features pulls in `iana-time-zone` which pulls in the `core-foundation` on macOS. 

This PR just disables the default features (including "clock") in chrono, no functional changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1735,6 +1735,7 @@ version = "0.1.0"
 dependencies = [
  "backtrace",
  "cc",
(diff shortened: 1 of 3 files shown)


sunshowers:
Thank you!

sunshowers:
Released in quick-junit 0.3.5.

PR #1086, opened 2023-11-01
Add libtest json output option

Jake-Shadle:
There are a couple of todos, but the biggest missing piece is combined stdout/err output.

Opening this as a draft while I make the combined output helper.

Issue #20, opened 2022-02-10
Machine-readable output for cargo nextest run


sunshowers:
The output should be similar to what cargo itself provides, complete with `--message-format json` and `--message-format json-render-status` options.

sunshowers:
This is something that we'd still like to do but it's just the work that needs to be done.

**Help wanted:** If you're interested in tackling this project, here's the general set of things we need to do:

1. Define a format with data structures for how to. The format should carefully be designed to allow future extensibility (e.g. enums should be [open universes](https://sunshowers.io/posts/open-closed-universes/) for forward compatibility).
2. Add that format to `nextest-metadata`.
3. Make `nextest-runner` produce output in that format.
4. Carefully work through backwards compatibility concerns: if we need to make a change in the future, how would it be done?
6. Add tests for the format: probably can migrate the nextest [integration tests](https://github.com/nextest-rs/nextest/tree/main/integration-tests/tests/integration) to the format.
7. Add tests for forward compatibility as well -- need to make sure that future additions don't break the format.
8. Write documentation for the format.
9. Stretch goal: Make `nextest-metadata` able to execute cargo nextest run with that format.

I'd estimate this to be around 80 hours of work for a newcomer to the project -- it's a chunky, nontrivial project but totally doable for someone new and interested.

Jake-Shadle:
We're interested in doing this work so that we can feed structured test results to [this](https://buildkite.com/test-analytics). (Note that while it supports JUnit which nextest can already produce that option has limitations that make it less appealing).

Before I started work on this, I wanted to run the idea I have for the initial implementation by you first before I do work that then gets rejected.

Basically the idea would be to have a `--message-format libtest-json --format-version <n>` where n is a serial incrementing number (the same `--format-version <n>` option already exists for `cargo metadata`) that follows the current structured output of [nightly](https://github.com/rust-lang/rust/issues/49359) so that users can specify a particular revision of the format for stability, even in the face of breaking changes. This would allow [tools](https://github.com/buildkite/test-collector-rust) that understand the (current) nightly output to be integrated with nextest, similar to the rationale for supporting JUnit.

[...]

PR #1088, opened 2023-11-02
Add support for combined test output

Jake-Shadle:
This PR is branched off of #1086, but can be split out to its own PR separately, just wanted to show that this is working with all of the current tests, as well as allowing improved capabilities for better emulating `libtest` style output.

Basically, instead of capturing stdout and stderr into their own separate buffers, test output (other process capturing like test listing is not touched in this PR) is captured into a single interleaved buffer, where each chunk of data is recorded with a timestamp when it was received, its byte range in the interleaved buffer, and which stream it came from. With this information it's possible to satisfy all of the existing code and tests by reconstructing each individual stream.

The libtest format can then just used the interleaved buffer for its output, more closely matching the output of libtest.

```json
{"type":"test","event":"failed","name":"myplayground::myplayground$fails_expectedly_with_wrong_value","exec_time":0.015041828,"stdout":"\nrunning 1 test\nthis is stdout\nthis is stderr\nthread 'fails_expectedly_with_wrong_value' panicked at src/lib.rs:52:5:\nfiddlesticks\nstack backtrace:\n   0: rust_begin_unwind\n             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5\n   1: core::panicking::panic_fmt\n             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14\n   2: myplayground::fails_expectedly_with_wrong_value\n             at ./src/lib.rs:52:5\n   3: myplayground::fails_expectedly_with_wrong_value::{{closure}}\n             at ./src/lib.rs:48:40\n   4: core::ops::function::FnOnce::call_once\n             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/ops/function.rs:250:5\n   5: core::ops::function::FnOnce::call_once\n             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/ops/function.rs:250:5\nnote: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\ntest fails_expectedly_with_wrong_value - should panic ... FAILED\n\nfailures:\n\n---- fails_expectedly_with_wrong_value stdout ----\nnote: panic did not contain expected string\n      panic message: `\"fiddlesticks\"`,\n expected substring: `\"fiddlesticksy\"`\n\nfailures:\n    fails_expectedly_with_wrong_value\n\ntest result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.01s\n\n"}
```


Jake-Shadle:
Oh, looks like my assumption that MacOS would work since FIONREAD does work correctly on FreeBSD was...incorrect. Will fix tomorrow.

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+        match self.write(s.as_bytes()) {
+            Ok(_) => Ok(()),
+            Err(_) => Err(std::fmt::Error),
+        }
+    }
+}
+
+pin_project! {
+    #[project = StreamProj]
+    enum Stream {
```
NobodyXu, line 197: I think you could use [`tokio_util::either::Either`](https://docs.rs/tokio-util/latest/tokio_util/either/enum.Either.html) here, it is probably what you want and it implements `AsyncRead` and `AsyncBufRead`.

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+///
+/// This size is not totally arbitrary, but rather the (normal) page size on
+/// most linux, windows, and macos systems.
+const CHUNK_SIZE: usize = 4 * 1024;
+
+/// Collects the stdout and/or stderr streams into a single buffer
+#[allow(clippy::needless_lifetimes, clippy::let_and_return)]
+pub fn collect_test_output<'a>(
+    stdout: Option<tokio::process::ChildStdout>,
+    stderr: Option<tokio::process::ChildStderr>,
```
NobodyXu, line 249: Can the API be changed to `Option<(ChildStdout, ChildStderr)>` to avoid the `unreachable!()` used in match?

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+                );
+            }
+        };
+
+        let mut out_done = false;
+        let mut err_done = false;
+
+        while !out_done || !err_done {
+            let (read, which) = tokio::select! {
+                buf = stdout.fill_buf() => {
```
NobodyXu, line 278: Nit:

```suggestion
                res = stdout.fill_buf() => {
```

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+                }
+                buf = stderr.fill_buf() => {
+                    let buf = buf.map_err(crate::errors::CollectTestOutputError::ReadStderr)?;
+                    push_chunk(acc, buf, false);
+                    (buf.len(), false)
+                }
+            };
+
+            if which {
+                stdout.consume(read);
```
NobodyXu, line 291: I think we can do consume and the eof check within `tokio::select!!` directly

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+            } else {
+                stderr.consume(read);
+                err_done = read == 0;
+            }
+        }
+
+        Ok(())
+    };
+
+    read_loop
```
NobodyXu, line 302: You can return that future directly without binding it to a variable.

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+#[inline]
+fn push_chunk(acc: &mut TestOutputAccumulator, chunk: &[u8], stdout: bool) {
```
NobodyXu, line 306: I think it'd be better to have an enum:

```
enum Source {
    Stdout,
    Stderr
}
```

It's much better than a `bool`

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,322 @@
...
+    let start = acc.buf.len();
+
+    if acc.buf.capacity() - start < chunk.len() {
+        acc.buf.reserve(CHUNK_SIZE);
```
NobodyXu, line 310: Don't think this is necessary, `extend_from_slice` would also do the reserve

NobodyXu (changes requested):
Sorry if I sound rude, but I am honestly shocked to see unsafe and raw platform-dependent syscalls used instead of tokio to achieve async read.

nextest-runner/src/test_output.rs
Jake-Shadle: Not without lifetime errors.

NobodyXu: That's...strange because the return type has the right lifetime annotation.

nextest-runner/src/test_output.rs
Jake-Shadle: It will only reserve for the size of the slice potentially giving a ton of extra allocations since most reads will be significantly less than 4k.

nextest-runner/src/test_output.rs
Jake-Shadle: Normally I would agree, but this is not part of the public API so the introduction of an enum just adds unnecessary friction.

nextest-runner/src/test_output.rs
NobodyXu: Hmmm ok

nextest-runner/src/test_output.rs
NobodyXu: I was just a bit worried about misuse of the `bool`, but since it is only used within `push_chunk` I guess it's ok.

sunshowers:
Thanks for the PR!

I haven't looked at it yet but I have to ask again (https://github.com/nextest-rs/nextest/pull/1086#issuecomment-1789479177):

For stdout/stderr, can we just do something like:

```
*** STDOUT ***
<stdout>
*** STDERR ***
<stderr>
```

Jake-Shadle:
We could, I'm just not sure why? The split output can be extremely confusing when looking at a failed test because the relationship between the streams is completely broken.

sunshowers:
Hmm, so I guess the same argument would apply to nextest's own output as well. Why should we not provide combined output across stdout and stderr? (Maybe we should!)

Jake-Shadle:
I mean, this PR enables either of those styles of output since it records which stream the output came from so that the libtest output can be combined and the rest of the code still functions as it does on main, so it wouldn't be a stretch to either convert it all or allow users to decide. The code is quite trivial.

NukeManDan:
Trying out this PR (very excited to see this getting done!!) and I am a big fan of the separation of stdout and stderr - so just wanted to be sure that output would be the raw `reporter` message in stdout enabling easy capture/piping of it without necessarily writing the report to disk (as is the case with JUnit XML files, that are the only alt. approach for me to do report post processsing). It is also handy to be able to screen out stderr with all the logs from nextest and only print the raw report with `2> /dev/null.

I am working now to get a workflow that would enable a terminal (local or CI logs) to print out all details and also capturing JSON for further processing AFAIK. As the PR here works now, I am able to use the command `cargo nextest r --message-format libtest-json -E"kind(test)" > test-report.json` and _almost_ get the exact behavior I want (terminal prints useful logs, json not printed, but is in stored in the piped file for integration test only).

One request: each item in the report is valid json, but the overall set of these events flushed to stdout IMHO should be a proper json array, so in my case I would not need to post process into a valid json file by wrapping these into an array. :pray:  That seem reasonable? As it is now you get:
```json
{"type":"suite",...}
{"type":"test",...}
...
```
But I would want:
```json
[
{"type":"suite",...},
{"type":"test",...},
...
]
```

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,258 @@
...
+            timestamp: Instant::now(),
+            stdout: self.stdout,
+        });
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
```
NukeManDan, line 182: Just guessing this is related to a bug I noticed: When I run multiple integration tests at least, only the _last_ module in those tests is present in the stdout of the report. Snippet of an example run, so I know tests are run, just not reported i the json:

```
       FAIL [   0.003s] pba-qualifier-exam::h_advanced_traits_tests ic_4
        FAIL [   0.002s] pba-qualifier-exam::h_advanced_traits_tests mixed_2
        FAIL [   0.003s] pba-qualifier-exam::h_advanced_traits_tests ic_3
        FAIL [   0.003s] pba-qualifier-exam::h_advanced_traits_tests ic_2
        FAIL [   0.003s] pba-qualifier-exam::h_advanced_traits_tests omni_3
        FAIL [   0.004s] pba-qualifier-exam::i_extension_traits_tests simple_non_empty
        FAIL [   0.004s] pba-qualifier-exam::i_extension_traits_tests simple_empty
        FAIL [   0.003s] pba-qualifier-exam::i_extension_traits_tests trait_empty
        FAIL [   0.003s] pba-qualifier-exam::i_extension_traits_tests trait_non_empty
        FAIL [   0.002s] pba-qualifier-exam::k_macros_tests map_single_entry
        PASS [   0.003s] pba-qualifier-exam::k_macros_tests map_empty
        FAIL [   0.003s] pba-qualifier-exam::k_macros_tests map_three_entries
        FAIL [   0.003s] pba-qualifier-exam::m_builder_tests builder_all_set
        FAIL [   0.003s] pba-qualifier-exam::m_builder_tests builder_correct_defaults
        FAIL [   0.002s] pba-qualifier-exam::m_builder_tests builder_fail
        FAIL [   0.002s] pba-qualifier-exam::m_builder_tests typed_builder_correct_defaults
        FAIL [   0.002s] pba-qualifier-exam::m_builder_tests typed_builder_all_set
        FAIL [   0.372s] pba-qualifier-exam::h_advanced_traits_tests british_engine_only_works_with_btu
        FAIL [   0.818s] pba-qualifier-exam::h_advanced_traits_tests internal_combustion_only_works_with_uranium
        FAIL [   1.318s] pba-qualifier-exam::h_advanced_traits_tests nuclear_reactor_only_works_with_uranium
        FAIL [   1.741s] pba-qualifier-exam::k_macros_tests impl_get_single_pub_u64
        FAIL [   2.240s] pba-qualifier-exam::k_macros_tests impl_get_mixed_types_and_vis
        FAIL [   2.749s] pba-qualifier-exam::k_macros_tests impl_get_single_u32
        FAIL [   3.758s] pba-qualifier-exam::k_macros_tests impl_get_multiple_arrays
     Running [ 00:00:03] [=========================================================================================================> ] 101/102: 0 running, 4 passed, 98 failed, 0 skipped     {"type":"suite","event":"started","test_count":7,"nextest":{"crate":"pba-qualifier-exam","test_binary":"k_macros_tests","kind":"test"}}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$impl_get_mixed_types_and_vis"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$impl_get_multiple_arrays"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$impl_get_single_pub_u64"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$impl_get_single_u32"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$map_empty"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$map_single_entry"}
{"type":"test","event":"started","name":"pba-qualifier-exam::k_macros_tests$map_three_entries"}
{"type":"test","event":"failed","name":"pba-qualifier-exam::k_macros_tests$map_single_entry","exec_time":0.00232729,"stdout":"\nrunning 1 test\nthread '<unnamed>' panicked at tests/k_macros_tests.rs:19:5:\nassertion `left == right` failed\n  left: None\n right: Some(3)\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\nthread 'map_single_entry' panicked at tests/k_macros_tests.rs:15:1:\nexplicit panic\ntest map_single_entry ... FAILED\n\nfailures:\n\nfailures:\n    map_single_entry\n\ntest result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 6 filtered out; finished in 0.00s\n\n"}
```

Jake-Shadle: That flush is irrelevant, that's only a writer for an in-memory buffer, it has nothing to do with stdout.

Jake-Shadle: That being said I can repro the bug and will fix it.

Jake-Shadle:
> One request: each item in the report is valid json, but the overall set of these events flushed to stdout IMHO should be a proper json array, so in my case I would not need to post process into a valid json file by wrapping these into an array. 🙏 That seem reasonable? As it is now you get:

That's not how libtest works, I think if you want an improved JSON/etc output you should comment on #20 about what format would suit you, #1086 set it up so that adding new machine readable formats should be easy.

sunshowers:
(I think newline-delimited JSON is correct.)

nextest-runner/src/test_output.rs
Jake-Shadle: Fixed.

Jake-Shadle:
I've marked this as ready for review as I think this is good enough to bring in, especially considering that the libtest-json is marked as experimental.

The test output is now always combined, and for non-json test output still preserves the current behavior of emitting stdout and stderr parts separately, passing all tests, however the combined output is now used as the `stdout` value when using --message-format libtest-json for failed tests, including the filtering out of the pieces of data emitted by libtest in the human format that are not part of the output when using the libtest json format.

The TestOutput additionally has a lines iterator that can help in future formats that may want to emit the output itself in a structured way, with eg. timing data, stdout vs stderr, etc.

sunshowers:
Thanks. This makes sense.

nextest-runner/src/reporter/structured/libtest.rs
```diff
@@ -349,31 +367,47 @@ impl<'cfg> LibtestReporter<'cfg> {
...
+                        // Unfortunately, to replicate the libtest json output,
+                        // we need to do our own filtering of the output to strip
+                        // out the data emitted by libtest in the human format
+                        write!(out, r#","stdout":""#).map_err(fmt_err)?;
+
+                        let mut in_test_output = false;
+
+                        for line in last_status.output.lines() {
+                            let line = if in_test_output {
+                                let data = line.lossy();
```
sunshowers, line 382: Looks like this chunk of code and the one immediately below doesn't have tests. Could you add some?

nextest-runner/src/test_output.rs
```diff
@@ -0,0 +1,628 @@
...
+
+    /// Gets a writer the can be used to write to the accumulator as if a child
+    /// process was writing to stderr
+    #[inline]
+    pub fn stderr(&mut self) -> TestOutputWriter<'_> {
+        TestOutputWriter {
+            acc: self,
+            stdout: false,
+        }
+    }
```
sunshowers, line 163: Could you rename these to `stdout_mut` and `stderr_mut`?

sunshowers (approved):
Just a couple of comments and then this is good to go. Thanks!

Jake-Shadle:
Oh, that was easier than I thought it would be.

Basically, we now have 3 separate `CaptureStrategy`:

* None - Used when --no-capture is used, stdout/stderr are not piped
* Split - Used when not outputting JSON/structured data, this is the same as the current code, which is that each stream is captured separately, so when combined in a single buffer they will ordered correctly only in the same stream, but may have quite a different ordering if streams are mixed
* Combined - Used when outputting JSON, I went for the most minimally invasive change without changing too much of the current code around process spawning and output capturing, by configuring the Command::stdout/err to use the same pipe (pipe on unix, named pipe/file on windows) and then keeping the single read end as the output from the program. This usage of a single pipe/file for the child process means that the output is as accurate as it can be, but does come with the consequence that the output is always attributed to stdout since we can no longer differentiate between streams.

This is using methods that were only just added in 1.74.0 which is why all of the 1.70.0 checks are failing. If moving to that MSRV isn't acceptable then I can remove the combined output capture, it would just need to be noted on the libtest output that the interleaved output in the `stdout` field can be **wildly** inaccurate even in simple situations where stdout and stderr printing are mixed together.

Note I also replaced `windows` with `windows-sys`, I hadn't noticed this crate was the sole user of `windows` in the crate graph, there was no reason for using `windows` since no COM is used by this crate, and `windows` is ridiculously larger (11.8MiB uncompressed) and slower to compile than `windows-sys`. If you want to keep `windows` I can revert that change.

sunshowers:
> FYI still working on this, while adding some tests I noticed that my naive assumption that combining 2 different streams would mostly be correct in most situations was completely wrong on both linux and windows (and most likely mac/bsd/etc)

Could you elaborate on what made this completely wrong? At a high level it seems like it would be fine, given that `tokio::select!` randomly picks the next future to poll from.

Also re Rust 1.74, stderr and stdout are just file descriptors (HANDLEs on Windows) so we should be able to convert them into `OwnedFd`/`OwnedHandle` and from there to `File` instances, unless I'm missing something. We can definitely bump up the MSRV to 1.72 right now, but based on my read of the docs that shouldn't be necessary.

Replacing windows with windows-sys is fine and was something I'd been wanting to do at some point, so thanks for doing that!

Jake-Shadle:
> Could you elaborate on what made this completely wrong? At a high level it seems like it would be fine, given that tokio::select! randomly picks the next future to poll from.

Given something like this:

```rust
#[test]
fn ordering_is_hard() {
    eprintln!("first stderr");
    println!("this is stdout");
    eprintln!("second stderr");
    panic!("oh no");
}
```

Running this test with output capturing would basically give something like

`first stderr\nsecond stderr\nthis is stdout\n<panic output>` about 80% of the time, at least on my Linux machine. In fact in some cases the `this is stdout\n` would even appear within the `<panic output>` as it is written in multiple separate writes between the panic header, the panic message, and the potential backtrace. It was only correctly ordered infrequently. I noticed the same issue on Windows, so figured it would be best to fix this in the PR rather than get bugs reported on the libtest output being wrong.

> Also re Rust 1.74, stderr and stdout are just file descriptors (HANDLEs on Windows) so we should be able to convert them into OwnedFd/OwnedHandle and from there to File instances, unless I'm missing something. We can definitely bump up the MSRV to 1.72 right now, but based on my read of the docs that shouldn't be necessary.

Oh yes, that will probably work, I went for the "match std exactly" option initially just to prove out the solution without needing to worry about "minor" divergences making it not work for non-obvious reasons. I'll do that after break.

> Replacing windows with windows-sys is fine and was something I'd been wanting to do at some point, so thanks for doing that!

Excellent.

NobodyXu:
Maybe you can just set stdout and stderr to point to the same pipe?

That should make things a lot easier, isn't it?

Jake-Shadle:
That's what this change does.

Jake-Shadle:
> Also re Rust 1.74, stderr and stdout are just file descriptors (HANDLEs on Windows) so we should be able to convert them into OwnedFd/OwnedHandle and from there to File instances, unless I'm missing something. We can definitely bump up the MSRV to 1.72 right now, but based on my read of the docs that shouldn't be necessary.

Done.

sunshowers:
I made some changes to prep this to land, but I wasn't able to push to this PR. So I created a new PR https://github.com/nextest-rs/nextest/pull/1212 for this.

Thanks for all the work!

sunshowers:
I'm planning to release a version of nextest containing support for this within the next 1-2 days.

sunshowers:
All right, this is part of cargo-nextest 0.9.67 which will be out within 20 minutes.

PR #1159, opened 2023-12-11
[nextest-filering] Remove reliance on Incomplete

epage:
Incomplete specifically means that there isn't enough data to continue processing.
Its a bit of a hack to reuse this for error recovery. This will no longer work if/when moving to `winnow`.

At least judging by the tests, nothing fails when removing this so its good enough?

### DIFF
--- a/nextest-filtering/src/parsing.rs
+++ b/nextest-filtering/src/parsing.rs
@@ -22,7 +22,6 @@ use nom::{
     combinator::{eof, map, peek, recognize, value, verify},
     multi::{fold_many0, many0},
     sequence::{delimited, pair, preceded, terminated},
-    Slice,
 };
 use nom_tracable::tracable_parser;
 use std::{cell::RefCell, fmt};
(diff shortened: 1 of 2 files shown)


nextest-filtering/src/parsing.rs
```diff
@@ -333,11 +332,6 @@ fn parse_matcher_text(input: Span) -> IResult<Option<String>> {
     )(input.clone())
     {
         Ok((i, res)) => (i, res.flatten()),
-        Err(nom::Err::Incomplete(_)) => {
-            let i = input.slice(input.fragment().len()..);
-            // No need for error reporting, missing closing ')' will be detected after
```
epage, line 338: @sunshowers it seems like this is intentional but no tests are affected.  Any advice on how to hit this case?

sunshowers: Not entirely sure! I wasn't the primary author of this code. Will have to investigate to figure this out (may take a few days, I'm really busy sorry!)

epage: Fully understand!  Thanks for any help!

sunshowers: Yeah, I just spent some time looking at this and you're right.

sunshowers:
Happy to land this if you don't have any objections.

sunshowers:
I started writing another parser for #1172 and realized that it would probably end up conflicting with the work you're doing to port it to winnow. Let me know how you want to sequence things.

epage:
Feel free to merge this if you are ready for it.

If you feel you will be getting to the new parser in the next couple of days, go for it.  I'd like to focus on this but with some time off and other obligations it might be spotty and I'd rather not hold you up

sunshowers:
OK, I ended up falling quite sick so never did it sadly. I may pick this up some long weekend.

PR #1174, opened 2023-12-20
Fix default-config.toml comment on `retry` semantics

miguelff:
Did I understand wrong that after a retry if the test succeeds is marked as flakey?

If so, should the text in the default-config.toml be as suggested by this PR?


sunshowers:
Yes, thank you!

PR #1212, opened 2024-01-09
 Add support for combined test output (to land)

sunshowers:
This is a version of #1088 that's ready for landing.

The big changes compared to #1088 are:

* The code to try and manage a single large buffer is all gone, since we never use that functionality and the code was quite complex.
* We now track some of the other exceptional cases (exec-fail, no-capture) properly instead of trying to shoehorn it all into the stdout/stderr model.
* I found and fixed some bugs, particularly around custom test harnesses.


sunshowers:
(the last commit did not have the correct attribution, so I force-pushed one that did)

PR #1214, opened 2024-01-10
[nextest-filtering] Replace `nom` parser with `winnow`

epage:
@sunshowers expressed interest in `winnow` and I figured porting an existing project they were familiar with would make it easier when they try it out.

When i first switched from `nom`s pure `Fn(I) -> (I, O)` parser model to `Fn(&mut I) -> O`, I was concerned about it making things worse.  It does have the downside of requiring more lifetime annotations to disambiguate things but I feel like it has been proven time and again to simplify code.  In this case, it removed a lot of bookkeeping from the functions, making the intentional bookkeeping more obvious, and it allowed the removal of `RefCell` for error recovery.

I broke this up into a lot of commits
- Smaller nuggets to review
- Easier to call out the differences
- Easier for me to deal with merge conflicts.

I did put a little too much in the `unpeek`-ectomy commit towards the end.  The main downside is that we can't `git bisect` to one specific bookkeeping change.

I did this as 1 PR though because the decision to commit to this is all-or-nothing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,7 +138,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.48",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


sunshowers:
Wow, thanks! Will look very soon. (you want to run cargo xfmt to fix the lint issue)

sunshowers:
This is awesome! Thank you. I'll go ahead and land this.

epage:
As you use it, I welcome any feedback to help make it better!

PR #1269, opened 2024-01-31
[ci] try out ubicloud runners

sunshowers:


sunshowers:
Looks like it's mostly a wash for nextest:

* without cache enabled, it's roughly the same as GHA with cache enabled
* with cache enabled, pulling from it is pretty slow

Issue #1340, opened 2024-02-29
Proposed design for including extra files in the archive


rukai:
Maintainer edit, tracking TODOs:

- [x] Add optional depth param, and limit depth by default to a reasonable value (we've chosen 8): https://github.com/nextest-rs/nextest/pull/1411
- [x] Warn if a path isn't present (may allow this to be configured as required or skipped in the future) -- also warn if a depth = 0 is a directory, which is a similar situation: https://github.com/nextest-rs/nextest/pull/1451
- [x] Check that path is well-formed: https://github.com/nextest-rs/nextest/pull/1454
- [x] Add test for skipped file types (not file, dir or symlink): https://github.com/nextest-rs/nextest/pull/1451
- [x] For errors, indicate why we're deciding to include a file in the archive (using an enum): https://github.com/nextest-rs/nextest/pull/1454
- [x] Write docs: https://github.com/nextest-rs/nextest/commit/f74d29954614a5764f8f6465175197c63d4234f3

---

From reading the discussion in https://github.com/nextest-rs/nextest/issues/423 it sounds like a feature like this is welcome.
But I wanted to make a proposal for a design before I set about implementing it.

I propose that we add the following configuration options to the config file and they will behave as documented.
```toml
# `nextest archive` automatically includes any build output required by a standard build.
# However sometimes extra non-standard files are required.
# "archive-include-within-target" specifies additional paths that will be included in the archive.
#
# Example: archive-include-within-target = ["application-data", "data-from-some-dependency/file.txt"]
# In this example:
# * the directory and its contents at "target/application-data" will be included in the archive.
# * the file "target/data-from-some-dependency/file.txt" will be included in the archive.
archive-include-within-target = []

# "archive-include-within-target-profile" provides an additional list of paths within the
# current cargo profile to include.
#
# Example: archive-include-within-target-profile = ["application-data", "data-from-some-dependency/file.txt"]
# In this example when compiling with the release profile:
# * the directory and its contents at "target/release/application-data" will be included in the archive
# * the file "target/release/data-from-some-dependency/file.txt" will be included in the archive.
archive-include-within-target-profile = []
```

I believe that these 2 config options should cover all possible use cases.

Alternatively we could have a single config option and support `*` globbing for profile specific paths. e.g. `*/application-data`
But that would mean that both release and debug builds would be included.

This design intentionally does not intend to add CLI flags for this as it feels like a build configuration level problem and not something the user would want ad-hoc control over.

[...]

PR #1344, opened 2024-03-01
Add archive-include config

rukai:
Progress towards https://github.com/nextest-rs/nextest/issues/1340

This PR implements the design given by rain in https://github.com/nextest-rs/nextest/issues/1340#issuecomment-1972163611
However it only implements the smallest possible functionality that will the allow the full design to be implemented in a backwards compatible manner in the future.
This is because I would like to land a small part of the design to ensure I am on track and also because I cant afford the time to implement all of these features right now.

The full implementation will look like:
```toml
[profile.ci]
archive-include = [
    { path = "foo", relative-to = "target" },
    { path = "bar/*", relative-to = "target-profile" },
]
```

However this PR's config will only support:
```toml
[profile.ci]
archive-include = [
    { path = "foo", relative-to = "target" },
    { path = "bar", relative-to = "target" },
]
```
This leaves the following features for follow up work:
* `relative-to = "target-profile"`
    + This can easily be introduced later.
    + For now, the user can also set `{ path = "debug/foo", relative-to = "target" }, { path = "release/foo", relative-to = "target" }` for a hacky work around.
* `*` globbing
    + while adding it later would technically be a breaking change since `*` can be included in a valid file name on linux. Using `*` in file names is generally considered a very bad idea, so I'm not worried about breakage here. I am however happy to include globbing support in this PR if such breakage is unacceptable.

## `optional` config

It was also stated that by default the `nextest archive` command should fail when the path to include does not exist.
I considered the positives and negatives of this versus just ignoring missing paths:
* Positives
    + Will give slightly better diagnostics to what went wrong if they enter the wrong path
       + I dont believe this will be a huge improvement, since the user has already diagnosed that there is something missing from their archived target that they need to configure nextest to pick up. So if after configuring nextest to include the path they get the same issue due to missing artifacts, they will naturally inspect the archive and config to find out what is going wrong.
    + Will catch paths that no longer need to be included and can be removed from the config.
* Negatives
   + Will cause archiving to unexpectedly fail in these scenarios:
        - nextest config is workspace wide and only a single crate produces the build artifact
        - When a crate or its dependency is designed to only output the build artifact in only some cases.

Ultimately the negatives are not huge issues since adding `optional = true` is not that hard, its mostly just that it feels `optional = true` is the more reasonable default at which point having it configurable seems less valuable.
In conclusion, I dont know which is better, so I've taken the simpler option of not implementing the `optional` config, but just confirm that you do want fail by default and I'll go ahead and implement it in this PR.

## Questions

* Am I expected to implement custom diagnostics for the ArchiveInclude deserialization
* should append_path_recursive be given an optional recursion depth so it can be used instead of append_dir_one_level?
* Is there any extra documentation I need to add? Or is the automatically generated page at https://nexte.st/book/configuration.html sufficient?

### DIFF
--- a/integration-tests/tests/integration/main.rs
+++ b/integration-tests/tests/integration/main.rs
@@ -22,7 +22,7 @@
 //! cargo-nextest, except it isn't used as the actual test runner. We refer to it with
 //! `NEXTEST_BIN_EXE_cargo-nextest-dup`.
 
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use nextest_metadata::{BuildPlatform, NextestExitCode};
 use std::{fs::File, io::Write};
 
@@ -338,12 +338,66 @@ fn test_relocated_run() {
 }
 
 #[test]
-fn test_run_from_archive() {
+fn test_run_from_archive_with_no_includes() {
     set_env_vars();
 
+    let (_p1, archive_file) = create_archive("");
+    let (_p2, extracted_target) = run_archive(&archive_file);
+
+    let expected_path = extracted_target
+        .join("application-data")
+        .join("foo")
+        .join("bar")
+        .join("some_file.txt");
+
+    assert!(
+        !Utf8Path::new(&expected_path)
+            .join("application-data")
+            .exists(),
+        "stray files must not be included in the archive"
+    );
+}
+
+#[test]
+fn test_run_from_archive_with_includes() {
+    set_env_vars();
+
+    let config = r#"
+[profile.default]
+archive-include = [
+    { path = "application-data", relative-to = "target" },
+    { path = "file_that_does_not_exist.txt", relative-to = "target" },
+]"#;
+    let (_p1, archive_file) = create_archive(config);
+    let (_p2, extracted_target) = run_archive(&archive_file);
+
+    let expected_path = extracted_target
+        .join("application-data")
+        .join("foo")
+        .join("bar")
+        .join("some_file.txt");
+
+    let contents = std::fs::read_to_string(expected_path).expect("extra file written to archive");
+    assert_eq!(contents, "a test string");
+}
+
+fn create_archive(config_contents: &str) -> (TempProject, Utf8PathBuf) {
     let custom_target_dir = Utf8TempDir::new().unwrap();
     let custom_target_path = custom_target_dir.path();
     let p = TempProject::new_custom_target_dir(custom_target_path).unwrap();
+
+    let config_path = p.workspace_root().join(".config/nextest.toml");
+    std::fs::write(config_path, config_contents).unwrap();
+
+    // setup extra file to include
+    let test_dir = p
+        .target_dir()
+        .join("application-data")
+        .join("foo")
+        .join("bar");
+    let test_file = test_dir.join("some_file.txt");
+    std::fs::create_dir_all(test_dir).unwrap();
+    std::fs::write(test_file, "a test string").unwrap();
 
     let archive_file = p.temp_root().join("my-archive.tar.zst");
 
@@ -367,7 +421,14 @@ fn test_run_from_archive() {
     std::fs::remove_dir_all(p.workspace_root()).unwrap();
     std::fs::remove_dir_all(p.target_dir()).unwrap();
 
+    // project is included in return value to keep tempdirs alive
+    (p, archive_file)
+}
+
+fn run_archive(archive_file: &Utf8Path) -> (TempProject, Utf8PathBuf) {
     let p2 = TempProject::new().unwrap();
+    let extract_to = p2.workspace_root().join("extract_to");
+    std::fs::create_dir_all(&extract_to).unwrap();
 
     let output = CargoNextestCli::new()
         .args([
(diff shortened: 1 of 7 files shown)


sunshowers:
Thanks! Is this ready for review? If so I'll try and make time in a few days (really busy with other stuff this week)

sunshowers:
Planning to look at this tomorrow!

integration-tests/tests/integration/main.rs
```diff
@@ -338,13 +338,66 @@ fn test_relocated_run() {
...
+    let (_p2, extracted_target) = run_archive(&archive_file);
+
+    let expected_path = extracted_target
+        .join("application-data")
+        .join("foo")
+        .join("bar")
+        .join("some_file.txt");
+
+    assert!(
+        !Path::new(&expected_path).join("application-data").exists(),
```
sunshowers, line 354: Can be `Utf8Path`.

nextest-runner/src/config/config_impl.rs
```diff
@@ -702,6 +702,13 @@ impl<'cfg> NextestProfile<'cfg, FinalConfig> {
             .unwrap_or(self.default_profile.fail_fast)
     }
 
+    /// Returns the archive-include config for this profile.
+    pub fn archive_include(&self) -> &Vec<ArchiveInclude> {
```
sunshowers, line 706: `&[ArchiveInclude]`

nextest-runner/src/reuse_build/archiver.rs
```diff
@@ -114,16 +117,19 @@ struct Archiver<'a, W: Write> {
...
 impl<'a, W: Write> Archiver<'a, W> {
+    #[allow(clippy::too_many_arguments)]
     fn new(
         binary_list: &'a BinaryList,
         cargo_metadata: &'a str,
         path_mapper: &'a PathMapper,
         format: ArchiveFormat,
         compression_level: i32,
         writer: W,
+        profile: &NextestProfile<'_, FinalConfig>,
```
sunshowers, line 132: `writer` should be the last argument as it's an out param. I think `profile` should maybe be the first argument.

nextest-runner/src/reuse_build/archiver.rs
```diff
@@ -295,6 +316,46 @@ impl<'a, W: Write> Archiver<'a, W> {
...
 
+    fn append_path_recursive(
+        &mut self,
+        rel_path: &Utf8Path,
+        src_path: &Utf8Path,
+    ) -> Result<(), ArchiveCreateError> {
+        let entries =
+            src_path
+                .read_dir_utf8()
+                .map_err(|error| ArchiveCreateError::InputFileRead {
```
sunshowers, line 327: We should indicate here that the file read is user-specified and not something that nextest archives by default.

sunshowers (commented):
This looks awesome! Just a few minor comments.

I'll make these changes myself and land this. Thanks again!

sunshowers:
Paths being optional by default still concerns me. I think I'll change that to required in a followup, so that.

Regarding docs, we should add a section to https://nexte.st/book/reusing-builds.html. I'll try writing some docs up tomorrow, unless you want to give it a shot.

I'm hoping to release this within the next 1-2 days. I tend to release mornings or afternoons Pacific time because it's better for breakages to be discovered quickly.

sunshowers:
Regarding recursion depth -- yeah, I think `append_path_recursive` should have an `Option<usize>` depth param.

nextest-runner/src/reuse_build/archiver.rs
sunshowers: Will do this in a followup.

nextest-runner/src/config/archive_include.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright (c) The nextest Contributors
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use serde::Deserialize;
+
+/// Type for the archive-include key.
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct ArchiveInclude {
+    pub(crate) path: String,
```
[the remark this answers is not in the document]
sunshowers: 1 done.

sunshowers:
Having thought about it more, I think I'd rather just produce a warning if a path wasn't found. We're going to produce warnings for a few other things too.

sunshowers:
Writing down some TODOs here based on what I saw:

- [x] Add optional depth param, and limit depth by default to a reasonable value (we've chosen 8): https://github.com/nextest-rs/nextest/pull/1411
- [ ] Warn if a path isn't present (may allow this to be configured as required or skipped in the future) -- also include warn if a depth = 0 is a directory, which is similar
- [ ] Check that path is well-formed
- [ ] Add test for skipped file types (not file, dir or symlink)
- [ ] For errors, indicate why we're deciding to include a file in the archive (using an enum)
- [ ] Write docs

Getting these all sorted will take me a few hours, so the release will probably be later this week. If you want to take on some of them then I'd love that -- will aim to get any followups in quickly and get this out soon.

rukai:
@sunshowers thanks for getting this over the line!
I might be able to find time to take on one or two of those items on friday but if you get to them first thats fine by me.

sunshowers:
I fell pretty sick the last two days so won't be able to finish it up this week (need to catch up on work) -- feel free to tackle as many of the above as you want.

Issue #1346, opened 2024-03-01
Miri's memory usage has decreased. Thoughts on giving the default profile some parallelism?


saethlin:
https://github.com/nextest-rs/nextest/blob/ef47d019a22f238458699575e1144a12ce810dcf/nextest-runner/default-config.toml#L111-L113

Previously, I'd have just agreed with the statement in the default config. But I've landed a number of PRs over the past months, culminating with https://github.com/rust-lang/rust/pull/118336 which have all reduced the interpreter's memory growth over time. The interpreter is still not lightweight by any means; unit tests seem to consume no less than 250 MB and I rarely see them using more than 500 MB.

But I don't think suppressing parallelism is a clearly superior default anymore. I'd bet that most development environments can support `available_parallelism()` Miri jobs now, but I expect you want to be more cautious than that. Thoughts?

sunshowers:
Sure, sounds good to me. I don't remember actively making this decision -- I think I went along with something either @RalfJung or you said.

Do you want to be more cautious and increase the parallelism to (say) 4, or do you just want to remove the constraint entirely and let the settings from the default profile take over?

At some point I'd love to add arithmetic operations, so you can say something like `min(system_memory() / 2GiB, num_cpus())`, but that's a far-future thing.

saethlin:
Yeah that arithmetic would be ideal. If the only options are full parallelism or a fixed number, I'd prefer full parallelism. I suspect it will Just Work for nearly everyone, and I have little grasp of how many people are missing out on nextest parallelism because of this default, but I suspect that's the more common experience.

sunshowers:
Sure. Would you like to submit a PR for this? I'm planning to release a new version soon.

PR #1347, opened 2024-03-02
Have Miri use all the cores by default

saethlin:
I think this is all that's needed? Quite a slick system if so.
Closes https://github.com/nextest-rs/nextest/issues/1346


sunshowers:
Yep, that's all that's needed. Thanks!

PR #1348, opened 2024-03-02
[nextest-runner] add option hide-progress-bar to configuration file

joshleaves:
While configuring nextest for my own projects, I noticed that I couldn't control display of the progress bar from the `.config/nextest.toml` file and I would have to manually run the command with `--hide-progress-bar` every time, which didn't feel very intuitive.

This commit is here to fix this.

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -878,7 +878,10 @@ impl TestReporterOpts {
         if let Some(final_status_level) = self.final_status_level {
             builder.set_final_status_level(final_status_level.into());
         }
-        builder.set_hide_progress_bar(self.hide_progress_bar);
+        match self.hide_progress_bar {
+            true => builder.set_hide_progress_bar(Some(true)),
+            false => builder.set_hide_progress_bar(None),
+        };
         builder
     }
 }
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -67,6 +67,10 @@ success-output = "never"
 # to false.
 fail-fast = true
 
+# Hide the progress bar when running tests. For CI runs, consider setting this
+# to true
+hide-progress-bar = false
+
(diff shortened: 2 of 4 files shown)


sunshowers:
Thanks!

I don't think repository config is the right place for this -- this is pretty dependent on individual user preferences.

As documented [here](https://nexte.st/book/env-vars.html#environment-variables-nextest-reads), there's a `NEXTEST_HIDE_PROGRESS_BAR` environment variable which you can set to `1` to hide the progress bar. Hope that helps!

I'd like to add user-specific configuration at some point, but that does require some design work still.

joshleaves:
Oh right, never saw it that way. Thanks :)

Side note: I just tested and the environment variable should be set to `true`, not `1`. The documentation needs to be updated to reflect that.

For user-specific configuration, how about always reading from `$HOME/.nextest.toml` to read for properties that shouldn't be in repository config??

sunshowers:
> Side note: I just tested and the environment variable should be set to true, not 1. The documentation needs to be updated to reflect that.

Hmmmm, I don't like that -- for environment variables 1 should always be treated the same as true. We should be using `BoolishValueParser` if possible.

> For user-specific configuration, how about always reading from $HOME/.nextest.toml to read for properties that shouldn't be in repository config??

This is a matter of some controversy! Some folks like using the XDG directories, others want to put their configs in HOME. We'll have to do a survey of which

sunshowers:
I've fixed the `NEXTEST_HIDE_PROGRESS_BAR` issue in https://github.com/nextest-rs/nextest/commit/7aabada376f4d24325d2a37eb916c083906b2d62, thanks for flagging it! Once this is released, both `true` and `1` should work great.

sunshowers:
(And I've filed https://github.com/nextest-rs/nextest/issues/1419 for user-specific config.)

Thanks again for submitting the PR -- I'll close it now in favor of the above followups.

PR #1426, opened 2024-04-09
Fix link to "other features" from benchmarks.md

zaneduffield:
The link works in GitHub but is broken on the deployed website.


sunshowers:
Thank you!

Issue #1493, opened 2024-05-09
`cargo nextest run --workspace` fails with DLL missing if a macro lib exists 


06393993:
```
PS C:\src\rust-test> cargo nextest run --workspace
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
   Compiling proc-macro2 v1.0.82
   Compiling unicode-ident v1.0.12
   Compiling rust-test v0.1.0 (C:\src\rust-test\hello_world)
   Compiling quote v1.0.36
   Compiling syn v2.0.61
   Compiling hello_macro v0.1.0 (C:\src\rust-test\hello_macro)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.82s
error: creating test list failed

Caused by:
  for `hello_macro`, command `'C:\src\rust-test\target\debug\deps\hello_macro-621f285b4f25d626.exe' --list --format terse` exited with code 0xc0000135: The specified module could not be found. (os error 126)
--- stdout:

--- stderr:

---
```

I am using the pre-built cargo-next `0.9.70` from https://get.nexte.st/latest/windows. I am using the latest `cargo` `1.78.0`.
```
PS C:\src\rust-test> cargo nextest --version
cargo-nextest-nextest 0.9.70
PS C:\src\rust-test> cargo --version
cargo 1.78.0 (54d8815d0 2024-03-26)
```

Note that this issue doesn't exist on cargo 1.70.0 with the same `cargo-nextest` binary.

This can be reproduced by the following file structure:

```
PS C:\src\rust-test> tree /f
Folder PATH listing for volume ???
Volume serial number is EC09-0909
C:.
│   .gitignore
│   Cargo.lock
[...]

Issue #267, opened 2022-06-09
macos: dyld: Library not loaded: @rpath/libtest-77ee8c29c330e4a3.dylib


steveej:
*Note from maintainer:* [Here's an outline of the work required for this.](https://github.com/nextest-rs/nextest/issues/267#issuecomment-1152829695) Please dive in if you're interested!

---

we're having issues running `cargo nextest` in a nix-shell on a macos github actions runner.
we run the same steps on an ubuntu runner in parallel which succeeds, so this seems to be macos specific.

it may or may not have something to do that the tested source code is under `/tmp`, which on this macos runner has an implicit `/private/` prefix.

some references for context:

* [a script that runs `cargo nextest`](https://github.com/holochain/holochain/blob/cfb2943eeed6e49279ab2381efb3a0e46ce4877d/nix/pkgs/core.nix#L45)
* a recent [actions log](https://github.com/holochain/holochain/runs/6814468268?check_suite_focus=true#step:10:1543) of which the final output is pasted below

```
(...)
dyld: Library not loaded: @rpath/libtest-77ee8c29c330e4a3.dylib
  Referenced from: /private/tmp/holochain_repo/target/fast-test/deps/hdk_derive-73ec051829ad694a
  Reason: image not found
error: creating test list failed
Caused by:
  for `hdk_derive::proc-macro/hdk_derive`, running command `/private/tmp/holochain_repo/target/fast-test/deps/hdk_derive-73ec051829ad694a --list --format terse` failed
Caused by:
  command ["/private/tmp/holochain_repo/target/fast-test/deps/hdk_derive-73ec051829ad694a", "--list", "--format", "terse"] exited with code <signal 6>
Error: Process completed with exit code 104.
```

please let me know if i can provide further debug output to help us understand what's happening.

sunshowers:
OK, I know what's going on—we need to grab the value of "rustc --print sysroot" and set the sysroot path as part of LD_FALLBACK_LIBRARY_PATH properly. Will fix in the next couple of days or you're welcome to fix it as well.

steveeJ:
thanks for looking into this @sunshowers. 

> How is cargo installed?

it's provided by a nix-shell derivation which makes use of functions from here: https://github.com/oxalica/rust-overlay/.

> I'm guessing a rustup based installation sets particular environment variables that a nix install doesn't.

this is possible and could be easily resolved in the nix-shell setup as well.

[...]

PR #1499, opened 2024-05-13
Add rustc folder to dynamic linker search path

06393993:
... so that we can run cargo-nextest with proc-macro projects on Windows to fix https://github.com/nextest-rs/nextest/issues/1493.

This commit also adds integration tests for running cargo-nextest for proc-macro projects.


fixtures/nextest-tests/proc-macro-test/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
sunshowers, line 1: Is this an extra newline? Could be removed if not necessary.

nextest-metadata/src/test_list.rs
```diff
@@ -510,6 +510,27 @@ pub struct RustNonTestBinarySummary {
...
+    /// are supported.
+    ///
+    /// TODO: change to the `PlatformSummary` type once the CI picks up the latest cargo-nextest
+    /// binary.
+    #[serde(default)]
+    pub target: Option<PlatformSummary>,
+
+    /// The sysroot for the target platform.
+    ///
+    /// Empty if fialed to discover.
```
sunshowers, line 529:
```suggestion
    /// Empty if failed to discover.
```

nextest-metadata/src/test_list.rs
```diff
@@ -485,7 +485,7 @@ pub struct RustBuildMetaSummary {
 
     /// The target platforms used while compiling the Rust artifacts.
     #[serde(default)]
-    pub target_platforms: Vec<PlatformSummary>,
+    pub target_platforms: Vec<BuildPlatformsSummary>,
```
sunshowers, line 488: Ahhh, I see. I misunderstood what compatibility meant in this context, and I apologize for that. I think we do want to keep compatibility around in this case.

Also, this was my terrible oversight when I first put in `target_platforms`, I wish I'd have added this `BuildPlatformsSummary` struct at the time. I humbly apologize for the extra work you have to do here.

As far as handling this goes, I can think of three general directions.

1. Write out the old `target_platforms` field, and add a new field (say `targets`) which has the full context. Deprecate `target_platforms` and ask users to use the new field if it's present. (This may want to be an `Option<Vec<BuildPlatformsSummary>>`. (This is similar to `target_platform` below.)

   Hopefully, this adds enough indirection that will be the last time we'd need to do that.

2. Retain compatibility with the old fields in https://docs.rs/target-spec/3.1.0/src/target_spec/summaries.rs.html#33 by manually duplicating them, and add the `sysroot` as another field next to it.
    * One way to do this would be to use `serde(flatten)` -- but that requires internal buffering, and in general, serde has [numerous issues](https://github.com/serde-rs/serde/issues/1183) with internal buffering.
    * Another way would be to manually duplicate the fields. But this is a challenge for two reasons. One is that the struct is `#[non_exhaustive]` https://docs.rs/target-spec/3.1.0/src/target_spec/summaries.rs.html#32, and two is that new semantically-important fields could be added to target-spec in the future. (I'm the maintainer of target-spec as well, so I can make changes to it if they make sense.)
    
      The former we could probably add some kind of builder for, but the latter is uncomfortably easy to miss details for.
      
      We could have some kind of convoluted `PlatformSummaryBuilder` in `target-spec` that produces an error if not all fields were present, I guess. But I'm not convinced of the value of all that extra work.

3. Add another vector next to `target_platforms` called (say) `target_platforms_extra`, and promise that it's the same length as `target_platforms` and corresponds 1:1 to it (i.e. users can `zip` the two vectors).

I think between all of these options I like 1 the most. It is a little bit of extra work, but sets us up best for the future. (At some point in the future, we'll probably remove both `target_platform` and `target_platforms` with the appropriate warnings.)

nextest-runner/src/platform.rs
```diff
@@ -21,22 +23,52 @@ pub struct BuildPlatforms {
...
+        target: Option<TargetTriple>,
+    ) -> Result<Self, UnknownHostPlatform> {
+        let sysroot = reader.and_then(|mut reader| {
+            let mut line = Default::default();
+            match reader.read_line(&mut line) {
+                Ok(_) => {
+                    let original_line = line.as_str();
+                    let line = line.trim();
+                    if line.is_empty() {
+                        log::debug!(
```
sunshowers, line 49: I think this should be a warning -- not being able to find the sysroot is a bit weird.

nextest-runner/src/platform.rs
```diff
@@ -21,22 +23,52 @@ pub struct BuildPlatforms {
     /// In the future, this will become a list of target triples once multiple `--target` arguments
     /// are supported.
     pub target: Option<TargetTriple>,
+
+    /// The sysroot for the target platform.
+    ///
+    /// This is needed to run procedure macro binaries on Windows. See
+    /// [https://github.com/nextest-rs/nextest/issues/1493] for details. Empty if failed to detect.
```
sunshowers, line 30: Note that this is actually required on all platforms per #267, it's just that on other platforms rustup still sets the library path env vars. Could you update the comment to reflect that?

sunshowers (commented):
This looks really, really great. Thank you for all your hard work! Just a few comments but I think this is ready to go.

One general comment is that this won't yet work with archived builds, because we don't archive the sysroot. But I'd like to address that after getting a fix out to stop the bleeding.

fixtures/nextest-tests/proc-macro-test/src/lib.rs
06393993: This file contains only one newline character.

While I am not sure if `rustc` requires the source file to always end with new line separator(C/C++ somewhat requires this), I personally think it's good to follow that convention. For more rationale see https://unix.stackexchange.com/a/18789. But I don't feel very strong of it. I am Ok with leaving an actual empty file. WDYT?

nextest-runner/src/platform.rs
06393993: We've already added the warning log at [line 115 of `nextest-runner/src/list/rust_build_meta.rs`](https://github.com/nextest-rs/nextest/pull/1499/files/9c98c32f70611d86c088242b5862a6a48456b030#diff-5a11bb30b860bf0cc695f0d8140cdbe4b3d461bd8a542b335e7fe6e088476970R115) where we actually generate the `dylib_paths`. Let me know if we still want `log::warn!` here.

fixtures/nextest-tests/proc-macro-test/src/lib.rs
sunshowers: Ah. So I love trailing newlines, and I definitely agree that non-empty files should have them. But logically, an empty file has no lines, so there's no line to have a trailing newline appended to.

In other words, a file can be treated as a list of lines:

```
foo\n
bar\n
baz\n
```

and trailing newlines make sense because it's effectively part of the last line. But an empty file has no lines, so a trailing newline doesn't logically fit in.

nextest-runner/src/platform.rs
sunshowers: ahhh I see. I think in that case, the warning is fine. Could you document this in a comment?

nextest-runner/src/list/rust_build_meta.rs
```diff
@@ -110,6 +111,9 @@ impl RustBuildMeta<TestListState> {
         // FIXME/HELP WANTED: get the rustc sysroot library path here.
```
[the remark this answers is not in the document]
sunshowers: Oh I just noticed that you add the `bin` directory.

What about using `rustc --print target-libdir`? I believe that should work on all platforms.

cargo-nextest/src/dispatch.rs
```diff
@@ -1006,7 +1007,35 @@ impl BaseApp {
         // Next, read the build platforms.
         let build_platforms = match reuse_build.binaries_metadata() {
             Some(kind) => kind.binary_list.rust_build_meta.build_platforms.clone(),
-            None => discover_build_platforms(&cargo_configs, cargo_opts.target.as_deref())?,
+            None => {
+                let target_triple =
+                    discover_target_triple(&cargo_configs, cargo_opts.target.as_deref());
+                let mut args = vec!["--print", "sysroot"];
+                if let Some(target_triple) = &target_triple {
+                    args.extend(["--target", target_triple.platform.triple_str()]);
```
sunshowers, line 1015: Ahh so actually I realized reviewing this again -- we need to obtain the target libdirs for:

1. The host, unconditionally
2. The target, if present

So we have to call rustc up to twice. (And if we support multiple `--target` in the future, N+1 times.)

I think we'll have to change this code to achieve that effect.

nextest-metadata/src/test_list.rs
06393993: Done.

nextest-runner/src/platform.rs
06393993: Done.

nextest-runner/src/platform.rs
06393993: Done.

nextest-metadata/src/test_list.rs
06393993: Adopt approach 1, but name the field as `target_platforms_v2` to make it explicit that `target_platforms` is deprecated. WDYT?

nextest-metadata/src/test_list.rs
06393993: Done.

nextest-runner/src/platform.rs
```diff
@@ -21,22 +24,55 @@ pub struct BuildPlatforms {
     /// In the future, this will become a list of target triples once multiple `--target` arguments
     /// are supported.
     pub target: Option<TargetTriple>,
```
06393993, line 26: @sunshowers what's the plan for multiple targets here? Should we change this to a `Vec` here or should we just make `RustBuildMetaSummary::target_platforms_v2` a `Vec`?

06393993 (commented):
We probably can only merge this PR late this week, as I don't have much extra time working on this. The plan for the next step is:

* Create another commit that creates a builder for `BuildPlatforms` and `RustCli`
    * `BuildPlatformsBuilder` will make the ctor of `BuildPlatforms` better, so that we can create from
        * only target(useful for test and backward comaptible with the old version of `RustBuildMeta`);
        * target and host libdir without target libdir, and target, host libdir; and
        * target, host libdir and any number of the target libdir
    * `RustCli` to streamline the call to `rustc` as we now probably will call it multiple times.

* Integrate the new types with `BaseApp::new`. However I currently don't plan to actually support multiple targets here, as this is outside the scope of this PR.

nextest-runner/src/platform.rs
sunshowers: I'd just make `target_platforms_v2` a `Vec`. Actually supporting multiple targets is a lot more work, but let's make sure the relatively unchanging backend structures support this.

fixtures/nextest-tests/proc-macro-test/src/lib.rs
06393993: Done.

sunshowers:
I really, really appreciate all the work you've done here. Thank you for this kindness.

Totally understand if you don't have more time, there's just a fractal of complexity here. Please feel free to do as much as you can, and if you don't have time before this weekend, I'll probably pick it up on Saturday or Sunday and finish it up.

nextest-runner/src/platform.rs
```diff
@@ -50,19 +86,169 @@ impl BuildPlatforms {
         }
     }
 
-    /// Converts a target triple to a `String` that can be stored in the build-metadata.
+    /// Converts a target triple to a [`String`] that can be stored in the build-metadata.
     ///
-    /// cargo-nextest represents the host triple with `None` during runtime. However the
-    /// build-metadata might be used on a system with a different host triple. Therefore, the host
```
06393993, line 56: @sunshowers If `BuildPlatformsSummary` will be used on a different machine, does that mean we shouldn't serialize the target libdir and the host libdr to the summary and we should always detect both libdirs when creating the `BuildPlatforms` from `BuildPlatformsSummary`?

sunshowers: Ahh, good question. I don't expect this to be tackled in this PR, to be clear. But there are two broad strategies:

1. Go all-in: add the libdirs to the archive and use those rather than the ones on disk.
2. Discover the libdirs from the system while running them.

Given that we support running tests out of an archive without a Rust installation being present at all, I think overall 1 is probably the better approach -- if we can slim down the set of files we serialize to just the ones needed by proc-macro tests. But we may consider adding optional support for 2 as well.

I think we should still serialize these directories in the summary, but be intelligent about it while deserializing them.

nextest-runner/src/lib.rs
```diff
@@ -24,6 +24,7 @@ pub mod redact;
 pub mod reporter;
 pub mod reuse_build;
 pub mod runner;
+mod rustc_cli;
```
06393993, line 27: @sunshowers This crate(`nextest-runner`) probably is not the best place to add the new `rustc_cli` module which streamlines the `rustc --print target-libdir` command call, but:

* the current design of the `BuildPlatformBuilder` in the `cargo-nextest` requires both the host libdir and the target libdir string as an input; and
* the `runner_for_target` function in the `nextest-runner/tests/integration/target_runner.rs` file needs correct libdir strings to set up tests correctly

so we need to find a way to call `RustcCli` in both `cargo-nextest` and `nextest-runner`, and we end up with this solution. The alternatives are:

1. Do not set up or hard coded host libdir and target libdir for the `nextest-runner` integration tests, so that `nextest-runner` doesn't need `RustcCli`, and we can move it to the `cargo-nextest` crate; or
2. We can call `RustcCli` when we build `BuildPlatform`, so that the integration test of the `nextest-runner` crate doesn't explicitly depend on `RustcCli`, and we don't have to export it.
    cons:
        * we still need `RustcCli` to live in the `nextest-runner` crate
        * to make testing `BuildPlatform` easier I need to still keep the current interface that accepts string, which complicates the interface
3. Move `RustCli` to `cargo-nextest` and add a `dev-dependencies` entry to `nextest-runner` to depend on `cargo-nextest`.
    cons:
        * The dependency relationship becomes strange(may result in cyclic)
        *  We still need to expose `RustcCli` as a public interface

nextest-runner/src/list/rust_build_meta.rs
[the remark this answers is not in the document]
06393993: Done.

nextest-runner/src/platform.rs
```diff
@@ -5,64 +5,374 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct BuildPlatform {
+    /// The host platform.
+    pub host: Platform,
+
+    /// The host libdir.
+    pub host_libdir: Option<Utf8PathBuf>,
+
+    /// The target platform, if specified.
+    pub target: Option<BuildPlatformTarget>,
```
06393993, line 125: @sunshowers This design makes the builder complicate, e.g.:

```rust
        let mut builder = BuildPlatformBuilder::default();
        let mut target_builder = builder.set_target(TargetTriple::x86_64_unknown_linux_gnu());
        target_builder.set_libdir(Cursor::new("/fake/target/libdir"));
        target_builder.add();
        let build_platform = builder
            .build()
            .expect("should build with target successfully");
```

The alternative design is a flat struct
```rust
pub struct BuildPlatform {
    pub host: Platform,
    pub host_libdir: Option<Utf8PathBuf>,
    pub target_libdir: Option<Utf8PathBuf>,
    pub target_triple: Option<TargetTriple>,
}
```
and the previous build code can be rewritten as
```rust
        let mut builder = BuildPlatformBuilder::default();
        builder
            .set_target_triple(TargetTriple::x86_64_unknown_linux_gnu())
            .set_target_libdir(Cursor::new("/fake/target/libdir"));
        let build_platform = builder
            .build()
            .expect("should build with target successfully");
```

Which way do you think is better?

nextest-runner/src/platform.rs
06393993: Done with approach 1.

nextest-runner/src/target_runner.rs
```diff
@@ -27,11 +27,11 @@ impl TargetRunner {
     /// or via a `CARGO_TARGET_{TRIPLE}_RUNNER` environment variable
     pub fn new(
         configs: &CargoConfigs,
-        build_platforms: &BuildPlatforms,
```
06393993, line 30: @sunshowers I rename `BuildPlatforms` to `BuildPlatform` as now we decided to not include an array of targets in `BuildPlatform`. WDYT?

cargo-nextest/src/dispatch.rs
06393993: Done. Detect the host and target libdirs separately.

06393993:
There seems to be still something wrong with the path setting. I will try to reproduce the issue locally and fix.

nextest-runner/src/platform.rs
06393993: Done.

06393993:
Actually the CI "Test with latest nextest release" step fails because the CI upgrades to the latest rustup, and when using the latest nextest release, it suffers from the exact same issue this PR tries to fix. Will try to add the env suggested in https://github.com/nextest-rs/nextest/issues/1493#issuecomment-2106331574 to pass the CI. Once the latest nextest release includes this PR, we come up with another PR to revert the env change.

However, with the env change, this PR can't be tested in the CI, at least the integration test almost won't catch anything in this change, we have to delay until the next PR to revert the env change to check if this PR has any issue. But we pass the "Test with locally built nextest" step in https://github.com/nextest-rs/nextest/actions/runs/9136713176/job/25125843092 without the env change, and this should give us some confidence.

nextest-runner/src/target_runner.rs
sunshowers: Ah, I've always imagined the plural in `BuildPlatforms` to consist of "host + target". So I think I'd prefer to keep it that way.

nextest-runner/src/platform.rs
```diff
@@ -5,64 +5,374 @@
...
-/// A representation of host and target platforms.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct BuildPlatforms {
-    /// The host platform.
-    pub host: Platform,
+fn read_first_line_as_path(mut reader: impl io::BufRead) -> Option<Utf8PathBuf> {
+    // We will print warn logs later when we are adding the path to the dynamic linker search paths,
+    // so we don't print the warn log here to avoid spammy log.
+    let mut line = Default::default();
+    match reader.read_line(&mut line) {
```
sunshowers, line 20: Maybe `reader.lines().next()`?

nextest-runner/src/platform.rs
```diff
@@ -5,64 +5,374 @@
...
+            parent: self,
+            triple,
+        }
+    }
+
+    /// Set the host libdir to the built [`BuildPlatform`].
     ///
+    /// Used to set the dynamic linker search path when running the test executables. If we fail to
+    /// parse the input, the libdir won't be set.
+    pub fn set_host_libdir(&mut self, reader: impl io::BufRead) {
```
sunshowers, line 61: I'd rather this just take a `Utf8PathBuf`, seems a bit too implementation-dependent to take the stdout of rustc here.

nextest-runner/src/platform.rs
sunshowers: Well -- the flat struct, as your proposed it, allows for the invalid state where `target_libdir` is `Some(_)` but `target_triple` is `None`. So I'd strongly prefer the `BuildPlatformTarget` pattern you have.

I'm not quite sure the builder pattern is pulling its weight though. Can the setters just live directly on the `BuildPlatforms` and `BuildPlatformTarget` structs?

nextest-runner/src/platform.rs
06393993: Changed to `set_host_libdir_from_rustc_output` for setting from the output. If we just want to set it, just use mutate the struct. WDYT?

06393993:
> Looks generally great, just a few more comments. Could you let me know within the next day or so whether you'll have time to get this finished up? If not then I can address them myself and land this.
> 
> Thanks again for all your work.

I do have time to address any further comments in the next week as I don't expect any big changes ahead.

sunshowers (commented):
Okay, just a couple more changes to go and then we can get this out there 🥳

nextest-metadata/src/test_list.rs
```diff
@@ -510,6 +516,26 @@ pub struct RustNonTestBinarySummary {
...
+#[serde(rename_all = "kebab-case")]
+pub struct BuildPlatformsSummary {
+    /// The target platform, if specified.
+    pub target: PlatformSummary,
+
+    /// The libdir for the host platform.
+    ///
+    /// Empty if failed to discover.
+    #[serde(default)]
+    pub host_libdir: Option<Utf8PathBuf>,
```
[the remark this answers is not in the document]
06393993: > also is independent justification to remove the "v2" stuff -- we're now clearly storing the host and target platforms separately, rather than smooshing them together.

IIUC, this new `platforms` field still makes both `target_platform` and `target_platforms` deprecated. Is that correct?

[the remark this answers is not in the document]
06393993: Does this implicitly implies that `BuildPlatforms::host` should respect the new `HostPlatformSummary::platform`?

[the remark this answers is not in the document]
06393993: In addition, what should be the interface of `from_summary` be like? `from_summary(summary: BuildPlatformsSummary) -> BuildPlatforms` or `from_summary(summary: BuildPlatformsSummary) -> Vec<BuildPlatforms>`? In the former case of `from_summary(summary: BuildPlatformsSummary) -> BuildPlatforms`, should we change `BuildPlatforms::target` from `Option<BuildPlatformsTarget>` to `Vec<BuildPlatformsTarget>`?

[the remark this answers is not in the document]
sunshowers: > IIUC, this new platforms field still makes both target_platform and target_platforms deprecated. Is that correct?

Yeah.



> Does this implicitly imply that `BuildPlatforms::host` should respect the new `HostPlatformSummary::platform`?

Yep, while performing deserialization.



> In addition, what should be the interface of `from_summary` be like? `from_summary(summary: BuildPlatformsSummary) -> Result<BuildPlatforms>` or `from_summary(summary: BuildPlatformsSummary) -> Result<Vec<BuildPlatforms>>`?

The former.

> In the former case of `from_summary(summary: BuildPlatformsSummary) -> Result<BuildPlatforms>`, should we change `BuildPlatforms::target` from `Option<BuildPlatformsTarget>` to `Vec<BuildPlatformsTarget>`?

Just error out if there's more than one.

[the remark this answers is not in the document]
06393993: > Just error out if there's more than one.

Which error should I return? `RustBuildMetaParseError::PlatformDeserializeError` or `RustBuildMetaParseError::UnknownHostPlatform` or should I create another enum variant? In case of using existing enum variant, how do I generate the source `target_spec::Error` or should I change the variant to be another enum which includes a `MultipleTargets` variant?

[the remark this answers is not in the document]
sunshowers: Another variant on RustBuildMetaParseError is fine.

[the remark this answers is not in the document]
06393993: Done.

* Introduce new types `HostPlatformSummary` and `TargetPlatformSummary`
* Deserialize `BuildPlatforms::host` from `HostPlatformSummary::platform`
* Introduce a new  `RustBuildMetaParseError` variant, which is returned if the argument of `from_summary(summary: BuildPlatformsSummary) -> Result<BuildPlatforms>` has more than one entries in the `targets` field.

nextest-metadata/src/test_list.rs
```diff
@@ -510,6 +516,45 @@ pub struct RustNonTestBinarySummary {
...
+
+/// Serialized representation of the host and the target platform.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct BuildPlatformsSummary {
+    /// The target platform used while compiling the Rust artifacts.
+    pub host: HostPlatformSummary,
+
+    /// The host platform used while compiling the Rust artifacts.
+    pub targets: Vec<TargetPlatformSummary>,
```
sunshowers, line 555:
```suggestion
    /// The host platform used while compiling the Rust artifacts.
    pub host: HostPlatformSummary,

    /// The target platforms used while compiling the Rust artifacts.
    ///
    /// With current versions of nextest, this will contain at most one element.
    pub targets: Vec<TargetPlatformSummary>,
```

nextest-runner/src/cargo_config/target_triple.rs
```diff
@@ -24,6 +24,18 @@ pub struct TargetTriple {
 }
 
 impl TargetTriple {
+    /// Create an x86_64-unknown-linux-gnu [`TargetTriple`]. Useful for testing.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the underlying implementation fail to parse the `"x86_64-unknown-linux-gnu"`
+    /// triple string.
+    pub fn x86_64_unknown_linux_gnu() -> Self {
```
sunshowers, line 33: Can this be `cfg(test)`?

sunshowers (approved):
Thanks for all your work. I'll land this and fix up the last couple of things.

nextest-runner/src/cargo_config/target_triple.rs
06393993: Yes, it can.

sunshowers: It can -- addressing this and other followups in comments.

cwfitzgerald:
Appreciate you both for getting this through!

PR #1528, opened 2024-05-28
[site] switch to mkdocs-material

sunshowers:
TODO: 

- [ ] Replace remaining examples in image or HTML form with ANSI-formatted blocks
- [x] Add a sitemap
- [x] Add redirects
- [x] Update build process


sunshowers:
This is better enough that I'm just going to land this, and fix up the remaining bits later.

PR #1638, opened 2024-08-18
[meta] update tokio to 1.39.3

sunshowers:


.github/workflows/docs.yml
```diff
@@ -37,6 +40,7 @@ jobs:
           mv site/output target/gh-pages
           mv target/doc target/gh-pages/rustdoc
       - name: Publish
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
sunshowers, line 43: Added this so that PRs sent from forks don't fail.

Issue #1483, opened 2024-05-06
Add support for `cargo test`s --skip and --exact command line flags


svix-jplatte:
Hi, I was just looking for an issue about `--skip` in nextest and found #29 which is saying that these are supported, but using different syntax. This is problematic for my use case since I have a test script that runs `cargo nextest run` if available, but falls back to regular `cargo test` if nextest is not installed. I also think that for interactive usage, `--skip` and `--exact` are much easier to use than the filter expression syntax.

Would you consider adding support for these flags? The implementation could just instantiate the appropriate filter expressions internally, so it should be really simple.

sunshowers:
Hi -- I was hesitant to add this at first, but over the last few months I've changed my mind. I'm happy to accept a PR for this. Some notes:

* `--skip` and `--exact` should only be accepted after `cargo nextest run --` -- the point after `--` is where we accept test binary args. See https://github.com/nextest-rs/nextest/blob/02a0cf0cb7d388c4cb709b2c9612a0b2b6bda800/cargo-nextest/src/dispatch.rs#L606-L629.
* Just like the existing support, filter expressions must be ANDed with them. i.e. if both a filter expression and `--skip`/`--exact` is passed in, both should be matched for a test to be run.
* I'm not exactly sure of the semantics of `--skip` and `--exact` especially when they're passed in together, so a full description of the semantics, with tests would be wonderful.

svix-jplatte:
I was able to get it to work, see #1653. It's in draft state though because I broke a test and couldn't really figure out how to add a new one either. And it's quite likely I didn't solve it in the best possible way.

sunshowers:
Released as part of cargo-nextest 0.9.80, which will be out within the next 15-20 minutes. Thanks again for your hard work!

sunshowers:
Also added documentation: https://nexte.st/docs/running/#--skip-and---exact

svix-jplatte:
Thanks!

PR #1653, opened 2024-08-21
WIP: Add support for --skip and --exact test binary arguments

svix-jplatte:
My half-finished attempt at #1483. It works, but clearly is not acceptable as-is because it adds a `todo!()` to one test, and doesn't add any tests for the new behavior.

Any hints as to how to get this into a mergable shape would be appreciated.

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -42,7 +42,7 @@ use nextest_runner::{
     show_config::{ShowNextestVersion, ShowTestGroupSettings, ShowTestGroups, ShowTestGroupsMode},
     signal::SignalHandlerKind,
     target_runner::{PlatformRunner, TargetRunner},
-    test_filter::{FilterBound, RunIgnored, TestFilterBuilder},
+    test_filter::{FilterBound, RunIgnored, TestFilterBuilder, TestFilterPatterns},
     write_str::WriteStr,
     RustcCli,
 };
@@ -590,11 +590,17 @@ struct TestBuildFilter {
     #[arg(long)]
     ignore_default_filter: bool,
 
-    /// Test name filters
+    /// Test name filters.
     #[arg(help_heading = None, name = "FILTERS")]
     pre_double_dash_filters: Vec<String>,
 
-    /// Test name filters and emulated test binary arguments (partially supported)
+    /// Test name filters and emulated test binary arguments.
+    ///
+    /// Arguments supported:{n}
+    /// - --ignored:         Only run ignored tests{n}
+    /// - --include-ignored: Run both ignored and non-ignored tests{n}
+    /// - --skip PATTERN:    Skip tests that match the pattern{n}
+    /// - --exact PATTERN:   Only run tests that exactly match the pattern
     #[arg(help_heading = None, value_name = "FILTERS_AND_ARGS", last = true)]
     filters: Vec<String>,
 }
@@ -648,13 +654,13 @@ impl TestBuildFilter {
     fn make_test_filter_builder(&self, filter_exprs: Vec<Filterset>) -> Result<TestFilterBuilder> {
         // Merge the test binary args into the patterns.
         let mut run_ignored = self.run_ignored.map(Into::into);
-        let mut patterns = self.pre_double_dash_filters.clone();
+        let mut patterns = TestFilterPatterns::new(self.pre_double_dash_filters.clone());
         self.merge_test_binary_args(&mut run_ignored, &mut patterns)?;
 
         Ok(TestFilterBuilder::new(
             run_ignored.unwrap_or_default(),
             self.partition.clone(),
-            &patterns,
+            patterns,
             filter_exprs,
         )?)
     }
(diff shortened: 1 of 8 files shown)


cargo-nextest/src/dispatch.rs
```diff
@@ -597,38 +598,53 @@ impl TestBuildFilter {
...
+                read_trailing_filters = true;
+            } else if arg == "--skip" {
+                let skip_arg = it.next().ok_or_else(|| {
+                    ExpectedError::test_binary_args_parse_error(
+                        "missing required argument",
+                        vec![arg.clone()],
+                    )
+                })?;
+
+                match FilteringExpr::parse(format!("!test({skip_arg})"), graph) {
```
sunshowers, line 627: I don't think this is quite right set-logic wise. The general selection logic here is that the set of tests run is: (UNION of all tests selected by regular arguments) INTERSECTION (UNION of all of tests selected by filtersets).

Based on this, here's what I think the behavior should be:

* `--exact` should be treated as part of the regular arguments (`patterns`).

  For example: `cargo nextest run -- a b -E 'test(c)' -E 'test(=d)' --exact e` should select tests that (have "a" in their name UNION have "b" in their name UNION named exactly "e") INTERSECTION (have "c" in their name UNION named exactly "d").

  This case fits nicely into the existing logic for regular arguments (but it should be treated as part of regular arguments, not filtersets).

* `--skip` should be applied at the end after everything.

  For example, `cargo nextest run -- a b -E 'test(c)' -E 'test(=d)' --skip e` should select tests that (have "a" in their name UNION have "b" in their name) INTERSECTION (have "c" in their name UNION named exactly "d") INTERSECTION (not have "e" in their name). This needs to be handled separately, as a third thing to intersect against.

While it's definitely possible to use filtersets to implement this logic, I think the requirement to pass in a graph here makes it somewhat problematic so I'd prefer just using regular arguments.

So here's what I would do:

1. Expand `patterns` so that rather than being a `Vec<String>` it is a `Vec<Pattern>`, where `Pattern` is an enum with `Contains` and `Exact` variants.
2. Also add a separate `skip_patterns`.
3. Add some set logic tests for all of these cases (I can help out with this if you get stuck, just doing 1 and 2 is okay and I can pick up 3.)

cargo-nextest/src/dispatch.rs
```diff
@@ -2318,7 +2340,7 @@ mod tests {
         fn get_test_filter_builder(cmd: &str) -> Result<TestFilterBuilder> {
             let app = TestCli::try_parse_from(shell_words::split(cmd).expect("valid command line"))
                 .unwrap_or_else(|_| panic!("{cmd} should have successfully parsed"));
-            app.build_filter.make_test_filter_builder(vec![])
+            app.build_filter.make_test_filter_builder(vec![], todo!())
```
sunshowers, line 2343: Ahh this is annoying for sure. The reason package graphs are needed by the filterset logic is that they're used for package queries. But if you switch the logic to not using filtersets, you should no longer need this.

sunshowers (commented):
Thanks for the PR and your work on this!

cargo-nextest/src/dispatch.rs
svix-jplatte: I force-pushed an incomplete version of this. I'm not clear on why `exact` would be added to the existing `patterns` and `skip` would be its own list. It seems to me like the different sorts of patterns need to be picked apart in `TestFilterBuilder::new` anyways, so might as well use one `Vec` for each (potentially bundled into a `struct`), or put them all in one `Vec`?

I also just saw this needs rebasing, will do that when the above question is answered, alongside fixing the missing docs and filtering logic.

sunshowers: Sure, you can use a separate Vec for exact patterns. The important thing is to get the set logic right.

sunshowers:
I'm going to pick this up and finish it off. Thanks for your work!

nextest-runner/src/test_filter.rs
```diff
@@ -63,30 +63,240 @@ enum TestFilterExprs {
...
+        /// General patterns.
+        patterns: Vec<String>,
+
+        /// Patterns passed in via `--exact`.
+        exact_patterns: HashSet<String>,
+
+        /// Patterns passed in via `--skip`.
+        skip_patterns: Vec<String>,
+    },
+}
```
sunshowers, line 87: I encapsulated the pattern logic into this lattice-like structure.

nextest-runner/src/test_filter.rs
```diff
@@ -63,30 +63,240 @@ enum TestFilterExprs {
...
+        skip_pattern_matcher: Box<AhoCorasick>,
+    },
+    Patterns {
         patterns: Vec<String>,
-        matcher: Box<AhoCorasick>,
+        exact_patterns: HashSet<String>,
+        skip_patterns: Vec<String>,
+        pattern_matcher: Box<AhoCorasick>,
+        skip_pattern_matcher: Box<AhoCorasick>,
     },
```
sunshowers, line 210: This is the resolved form with matchers for all the substring patterns. I separated `All` here mostly for convenience.

cargo-nextest/src/dispatch.rs
sunshowers: Having written the code -- agreed, a separate Vec is better. Thanks! :D

sunshowers:
Thanks again for all your hard work on this! Excited to get it into the hands of folks.

PR #1682, opened 2024-09-04
buildomat job with colors

sunshowers:


sunshowers:
Landed in https://github.com/nextest-rs/nextest/commit/dbd1ff381a872b07ca67acfbc477a07b2b2937c4.

PR #1707, opened 2024-09-12
Add support for silencing only one of the outputs of a test

### LINKED ISSUE #1688: Allow only showing stdout or stderr when using `--success-output` and `--failure-output`
I'm writing a file parser and when I get a new version of the format I like to add the samples to my tests.
However, this means that the tests will start hitting todo!() and panic!() and this pollutes the output of the tests:
```
        FAIL [   0.004s] ubiart_toolkit::tape tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd

--- STDOUT:              ubiart_toolkit::tape tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd ---

running 1 test
test tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd ... FAILED

failures:

---- tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd ----
test panicked: not yet implemented


failures:
    tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s


--- STDERR:              ubiart_toolkit::tape tape_parse_wiiu2015::base/tape.ckd/f054654bfd32299c39542c768be25c6e.tape.ckd ---
thread '<unnamed>' panicked at ubiart_toolkit/src/cooked/tape/mod.rs:2087:9:
not yet implemented
```
In these cases stdout has no useful additional information compared to stderr, so I would like to disable it.
I was thinking of the following:
--failure-output immediate -> outputs both stdout and stderr (current case)
--failure-output stdout=never -> does not output to stdout, uses default for stderr
--failure-output stdout=never,stderr=immediate -> does not output to stdout, immediate output for stderr
Another option is adding more flags.

Kriskras99:
The old syntax is still supported, the new complete syntax is:
 - "{value}":                         where value is one of [immediate, immediate-final, final, never]
 - "{stream}={value}":                where {stream} is one of [stdout, stderr]
 - "{stream}={value},{stream=value}": where the two {stream} keys cannot be the same

For tests where the output is combined (currently only used for libtest-json output), only the stdout value is used.

There is a small regression where the CLI won't suggest values that are close to the incorrect input, i.e. if you input 'imediate' it won't suggest 'immediate'.

I've not added tests yet, I first wanted to make sure that the implementation is acceptable for inclusion into nextest.

Closes #1688

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -32,7 +32,8 @@ use nextest_runner::{
     redact::Redactor,
     reporter::{
         heuristic_extract_description, highlight_end, structured, DescriptionKind,
-        FinalStatusLevel, StatusLevel, TestOutputDisplay, TestReporterBuilder,
+        FinalStatusLevel, StatusLevel, TestOutputDisplay, TestOutputDisplayStreams,
+        TestReporterBuilder,
     },
     reuse_build::{archive_to_file, ArchiveReporter, PathMapper, ReuseBuildInfo},
     runner::{
@@ -55,6 +56,7 @@ use std::{
     env::VarError,
     fmt,
     io::{Cursor, Write},
+    str::FromStr,
     sync::Arc,
 };
 use swrite::{swrite, SWrite};
@@ -871,7 +873,10 @@ enum MessageFormat {
 #[derive(Debug, Default, Args)]
 #[command(next_help_heading = "Reporter options")]
 struct TestReporterOpts {
-    /// Output stdout and stderr on failure
+    /// Output stdout and/or stderr on failure
+    ///
+    /// Takes the form of: '{value}' or 'stdout={value}' or 'stdout={value},stderr={value}'
+    /// where {value} is one of: 'immediate', 'immediate-final', 'final', 'never'
     #[arg(
         long,
         value_enum,
@@ -879,9 +884,12 @@ struct TestReporterOpts {
         value_name = "WHEN",
         env = "NEXTEST_FAILURE_OUTPUT",
     )]
-    failure_output: Option<TestOutputDisplayOpt>,
+    failure_output: Option<TestOutputDisplayStreamsOpt>,
 
-    /// Output stdout and stderr on success
+    /// Output stdout and/or stderr on success
+    ///
+    /// Takes the form of: '{value}' or 'stdout={value}' or 'stdout={value},stderr={value}'
+    /// where {value} is one of: 'immediate', 'immediate-final', 'final', 'never'
     #[arg(
         long,
         value_enum,
@@ -889,7 +897,7 @@ struct TestReporterOpts {
         value_name = "WHEN",
         env = "NEXTEST_SUCCESS_OUTPUT",
     )]
-    success_output: Option<TestOutputDisplayOpt>,
+    success_output: Option<TestOutputDisplayStreamsOpt>,
 
     // status_level does not conflict with --no-capture because pass vs skip still makes sense.
     /// Test statuses to output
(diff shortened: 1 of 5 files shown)


sunshowers:
Thanks for the PR! I'm currently quite sick but will look when I'm better.

Kriskras99:
No worries, get well soon!

cargo-nextest/src/dispatch.rs
```diff
@@ -963,6 +971,68 @@ impl TestReporterOpts {
...
+                (Some((stream @ "stdout" | stream @ "stderr", Err(_))), _) => return Err(format!("\n  unrecognized setting for {stream}: [possible values: immediate, immediate-final, final, never]")),
+                (_, Some((stream @ "stdout" | stream @ "stderr", Err(_)))) => return Err(format!("\n  unrecognized setting for {stream}: [possible values: immediate, immediate-final, final, never]")),
+                (Some(("stdout", _)), Some(("stdout", _))) => return Err("\n  stdout specified twice".to_string()),
+                (Some(("stderr", _)), Some(("stderr", _))) => return Err("\n  stderr specified twice".to_string()),
+                (Some((stream, _)), Some(("stdout" | "stderr", _))) => return Err(format!("\n  unrecognized output stream '{stream}': [possible values: stdout, stderr]")),
+                (Some(("stdout" | "stderr", _)), Some((stream, _))) => return Err(format!("\n  unrecognized output stream '{stream}': [possible values: stdout, stderr]")),
+                (_, _) => return Err("\n  [possible values: immediate, immediate-final, final, never], or specify one or both output streams: stdout={}, stderr={}, stdout={},stderr={}".to_string()),
+            }
```
sunshowers, line 1006: Hmm, can this be shortened? This feels excessive.

I'd probably do something like:

* split with `,` if possible, getting 1 or 2 entries from this (more than 2 should error out)
* if one entry, try parsing via `<stream>=<setting>` or then just `<setting>`
* if two entries, always try parsing as `<stream>=<setting>`, then at the end ensure that the streams are different

nextest-runner/src/reporter/displayer.rs
```diff
@@ -29,12 +29,172 @@ use serde::Deserialize;
...
+    pub fn create_never() -> Self {
+        Self {
+            stdout: Some(TestOutputDisplay::Never),
+            stderr: Some(TestOutputDisplay::Never),
+        }
+    }
+
+    /// Which output streams should be output immediately
+    ///
+    /// # Returns
```
sunshowers, line 81: We generally don't use a `# Returns` header here.

Also I'd rather this return a `DisplayOutput` rather than `Option<DisplayOutput>`, to avoid representing the invalid state where `DisplayOutput` is set to `false` for both.

nextest-runner/src/reporter/displayer.rs
```diff
@@ -29,12 +29,172 @@ use serde::Deserialize;
...
+                "stdout" => {
+                    if stdout.is_some() {
+                        return Err(A::Error::duplicate_field("stdout"));
+                    }
+                    stdout = Some(TestOutputDisplay::from_str(value).map_err(|_| {
+                        A::Error::invalid_value(
+                            serde::de::Unexpected::Str(value),
+                            &"'never', 'immediate', 'immediate-final', or 'final'",
+                        )
+                    })?);
```
sunshowers, line 167: As the warnings suggest, for a manual deserialize impl it would be good to have tests for all the error cases. (I know you haven't written tests yet, just wanted to mention this.)

sunshowers (commented):
Yeah I really like the overall direction. Thanks!

cargo-nextest/src/dispatch.rs
Kriskras99: It can be reduced to the two first lines and the last line, but that would result in significantly less helpful error messages.
I also don't see a way to reduce it without degrading the error messaging.

I could do a priority inversion by first checking if the value is a valid TestOutputDisplayOpt, which should be the most common case. And only then dealing with commas and equal signs.

sunshowers: Yeah, trying to parse it first that way makes sense.

ArhanChaudhary:
Hello! Any update on this?

Kriskras99:
No update unfortunately. It still needs tests added. It also has some merge conflicts that need to be resolved.

I won't get to it anytime soon, feel free to take over the PR.

ArhanChaudhary:
I've began an attempt to resolve the merge conflicts and port this PR to main. I can't figure out how to get a `display_error` so my code doesn't compile, but this should hopefully still move the ball down the field! See https://github.com/ArhanChaudhary-forks/nextest/commit/013807c18d670235f359c7533f0d1e4bd494ec15

sunshowers:
Thanks @ArhanChaudhary -- would you be willing to submit a PR even with your WIP? Happy to have a look.

ArhanChaudhary:
Done!

PR #1798, opened 2024-10-15
Set environment variables specified via `cargo::rustc-env` in the build script

chrjabs:
As discussed in #1787, `cargo test` does this, so `nextest` should too.

I added warnings in case the build script output file is not found, this should usually have been created when cargo executes the build script.
I also decided to package the build script output _file_ in the same step as the build script output _directory_.
Let me know if you prefer creating a separate step for this.

I tried to approach this with minimal changes, maybe it would be nicer to store the path to the `$OUT_DIR` _parent_, so that getting the path of the output file is a tad simpler, but that would have required me touching more code.
Let me know if you have suggestions for implementing any of this differently.

Side note: trying to run tests with `cargo run -p cargo-nextest -- nextest run --all-features` failed with tracing complaining that it failed to set a global subscriber since one is already set, so I ran tests without the `--all-features` flag.

### DIFF
--- a/fixtures/nextest-tests/with-build-script/src/lib.rs
+++ b/fixtures/nextest-tests/with-build-script/src/lib.rs
@@ -12,4 +12,17 @@ mod tests {
         let contents = std::fs::read(&path).expect("test file exists in OUT_DIR");
         assert_eq!(contents, b"test-contents");
     }
+
+    #[test]
+    fn test_build_script_vars_set() {
+        // Since the build script wrote `cargo::rustc-env` instructions, these variables are
+        // expected to be set by nextest
+        #[cfg(new_format)]
+        {
+            let val = std::env::var("BUILD_SCRIPT_NEW_FMT").expect("BUILD_SCRIPT_NEW_FMT is valid");
+            assert_eq!(val, "new_val");
+        }
+        let val = std::env::var("BUILD_SCRIPT_OLD_FMT").expect("BUILD_SCRIPT_OLD_FMT is valid");
+        assert_eq!(val, "old_val");
+    }
 }
--- a/nextest-runner/src/test_command.rs
+++ b/nextest-runner/src/test_command.rs
@@ -14,7 +14,10 @@ use once_cell::sync::Lazy;
 use std::{
     collections::{BTreeSet, HashMap},
     ffi::{OsStr, OsString},
+    fs::File,
+    io::{BufRead, BufReader},
 };
+use tracing::warn;
 
 mod imp;
 pub use imp::{Child, Output};
@@ -74,7 +77,8 @@ impl TestCommand {
         {
             // Convert the output directory to an absolute path.
             let out_dir = lctx.rust_build_meta.target_directory.join(out_dir);
-            cmd.env("OUT_DIR", out_dir);
+            cmd.env("OUT_DIR", &out_dir);
+            apply_build_script_env(&mut cmd, out_dir);
         }
 
         // Expose paths to non-test binaries at runtime so that relocated paths work.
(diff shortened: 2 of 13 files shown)


chrjabs:
Ah, I didn't realize you're testing against `1.75`, which doesn't support the new `cargo::` syntax yet. Any ideas for how to circumvent this? I could make the `with-build-script` test crate depend on [rustversion](https://docs.rs/rustversion/latest/rustversion/) to only include the new syntax test conditionally.

chrjabs:
I changed the test to be dependent on the `rustversion` crate now, but I kept it a separate commit for now, since this requires a new dependency for the testing fixtures, and I'm not sure if this is fine. If yes, the commits should be squashed, otherwise we'll need another fix.

sunshowers:
Thanks! Can this use a very simple version check, similar to:

https://github.com/camino-rs/camino/blob/467b33692ac2224f6ccfb6b4e96d7dfe2ea09ec7/build.rs#L32-L34

I don't think it's necessary to pull down a dependency here.

nextest-runner/src/test_command.rs
```diff
@@ -191,6 +195,34 @@ fn apply_package_env(cmd: &mut std::process::Command, package: &PackageMetadata<
...
+        };
+        // `cargo::rustc-env` is the official syntax since `cargo` 1.77, `cargo:rustc-env` is
+        // supported for backwards compatibility
+        let Some(key_val) = line
+            .strip_prefix("cargo::rustc-env=")
+            .or_else(|| line.strip_prefix("cargo:rustc-env="))
+        else {
+            continue;
+        };
+        let split = key_val.find('=').unwrap();
```
sunshowers, line 221: Could you update this to not unwrap and instead warn?

fixtures/nextest-tests/with-build-script/src/lib.rs
```diff
@@ -12,4 +12,17 @@ mod tests {
         let contents = std::fs::read(&path).expect("test file exists in OUT_DIR");
         assert_eq!(contents, b"test-contents");
     }
+
+    #[test]
+    fn test_build_script_vars_set() {
+        // Sinc the build script wrote `cargo::rustc-env` instructions, these variables are
```
sunshowers, line 18:
```suggestion
        // Since the build script wrote `cargo::rustc-env` instructions, these variables are
```

sunshowers (commented):
Looks generally great, thanks! Just a couple of comments.

nextest-runner/src/test_command.rs
```diff
@@ -191,6 +195,34 @@ fn apply_package_env(cmd: &mut std::process::Command, package: &PackageMetadata<
     );
 }
 
+/// Applies environment variables spcified by the build script via `cargo::rustc-env`
+fn apply_build_script_env(cmd: &mut std::process::Command, out_dir: Utf8PathBuf) {
```
sunshowers, line 199: Could you also add a unit test for this? Just testing out some of these situations.

chrjabs:
Now also added documentation, where I noted that cargo actually discourages using these variables at runtime. I hadn't noticed that before.

sunshowers (approved):
Looks great, thanks!

site/src/docs/configuration/env-vars.md
```diff
@@ -152,6 +152,9 @@ Nextest also sets these environment variables at runtime, matching the behavior
 
 Additionally, environment variables specified in [the `[env]` section of `.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html#env) are also set.
 
+Since `cargo test` also sets variables specified in the build script via [`cargo::rustc-env`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env) at test runtime, `nextest` does the same.
+However, note that [`cargo` discourages using these variables at runtime](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env).
```
sunshowers, line 156: Thanks for adding this! I generally add documentation at the time the release happens, so that the nexte.st site is up-to-date. I'll remove this for now and will re-add it when I'm getting ready to release 0.9.82 (hopefully in the next 1-2 days).

sunshowers:
> Side note: trying to run tests with cargo run -p cargo-nextest -- nextest run --all-features failed with tracing complaining that it failed to set a global subscriber since one is already set, so I ran tests without the --all-features flag.

Ah good catch, I'll try and fix this.

sunshowers:
Released this as part of cargo-nextest 0.9.82. Thanks again!

> Ah good catch, I'll try and fix this.

Fixed in https://github.com/nextest-rs/nextest/commit/11623d5ce06fd63ff3803499e594400876bac502.

PR #1835, opened 2024-10-29
[nextest-runner] demo of a hang

sunshowers:
Repro (just to keep it as minimal as possible + allow attaching a Tokio console)

```
cargo run -p cargo-nextest --all-features -- nextest run -p nextest-filtering -j1
```

### DIFF
--- a/nextest-runner/src/runner.rs
+++ b/nextest-runner/src/runner.rs
@@ -1082,6 +1082,9 @@ impl<'a> TestRunnerInner<'a> {
                     () = sleep, if !child_fds.is_done() => {
                         break true;
                     }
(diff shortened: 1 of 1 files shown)


nextest-runner/src/runner.rs
```diff
@@ -1082,6 +1082,9 @@ impl<'a> TestRunnerInner<'a> {
                     () = sleep, if !child_fds.is_done() => {
                         break true;
                     }
+                    _ = forward_receiver.recv() => {
```
sunshowers, line 1085: This is happening because the above branches get disabled but this one doesn't. The fix is to also add `!child_fds.is_done()` to this branch so the `else` is hit.

PR #1874, opened 2024-11-15
feat: support hyphen values for --build-jobs

mattsse:
cargo --jobs supports negative values: https://github.com/rust-lang/cargo/issues/9217

currently this fails:

```
cargo nextest run --build-jobs -1                                                                                                                                                                                                   
error: unexpected argument '-1' found

  tip: to pass '-1' as a value, use '-- -1'
```

because clap doesn't support this by default and requires `allow_hyphen_values = true`

https://github.com/clap-rs/clap/issues/1245


sunshowers:
Thank you!

PR #1918, opened 2024-11-29
more prep work

sunshowers:
Been splitting and landing chunks of work as atomic commits.


sunshowers:
The test coverage issue is https://github.com/taiki-e/cargo-llvm-cov/issues/400.

PR #1919, opened 2024-11-29
[meta] update MSRV to Rust 1.81, switch lints from allow to expect

sunshowers:
This caught a few outdated lints and a few that should have been `cfg(test)`.


glandium:
This change didn't update the README, which still says 1.75 is the MSRV

sunshowers:
@glandium thanks, fixed in https://github.com/nextest-rs/nextest/commit/8af1b7486a5b22e8631bea4608b5df48032a3aa6.

PR #1926, opened 2024-11-30
feat(max-fail): introduce --max-fail runner option

AJamesyD:
# TL;DR

Introduce `--max-fail` flag. Inspired by [pytest's own `--maxfail` flag](https://docs.pytest.org/en/7.1.x/how-to/failures.html#stopping-after-the-first-or-n-failures).

# Motivation

The `--max-fail` option provides a flexible middle-ground between the efficiency of `--fail-fast` and the comprehensive coverage of `--no-fail-fast`.

It allows users to uncover multiple issues without running the entire test suite, which is particularly valuable for tightening the write/test/debug iteration loop when running integration tests and/or large test suites.

# Implementation

## This PR

Replace all references to `fail_fast: bool` (except those in the profile config) with:

```rust
// Some(1) -> Equivalent to fail_fast = true
// Some(n) -> End run after n test failures
// None -> Equivalent to fail_fast = false / --no-fail-fast
self.max_fail: Option<usize>
```

If the maintainers believe it would be more readable and/or maintainable to introduce an Enum for `max_fail` I'm all ears. Would appreciate any other feedback you have as well.

## Potential Future Work

- Expose `max_fail` option in profile.
  - Potentially deprecate `fail_fast` option to minimize confusion / exposed surface area to same underlying interface.

# Open Questions
- Above note about creating a dedicated Enum for `max_fail`
- If we're not relying on negating `run_statuses.last_status().result.is_success()`, should `self.run_stats.exec_failed` and `self.run_stats.timed_out` count towards`max_fail`?


cargo-nextest/src/dispatch.rs
```diff
@@ -845,9 +845,24 @@ pub struct TestRunnerOpts {
...
+        name = "no-fail-fast",
+        conflicts_with = "no-run",
+        overrides_with = "fail-fast"
+    )]
     no_fail_fast: bool,
 
+    /// Number of tests that can fail before exiting test run
+    #[arg(
+        long,
+        name = "max-fail",
```
sunshowers, line 859: * It would be interesting to also support a value here that's equivalent to `--no-fail-fast`, e.g. `--max-fail=all` or `--max-fail=disabled`.

* Could you make this `require_equals = true`? For newer options we've been preferring that.

cargo-nextest/src/dispatch.rs
```diff
@@ -845,9 +845,24 @@ pub struct TestRunnerOpts {
...
+        value_name = "N",
+        conflicts_with_all = &["no-run", "no-fail-fast"],
+        overrides_with = "fail-fast"
```
sunshowers, line 862: Hmm is there a reason this overrides `--fail-fast`? I would expect this to conflict with `--fail-fast`, not override it. (Unlike `--no-fail-fast` which is a clear negation.)

cargo-nextest/src/dispatch.rs
```diff
@@ -2420,6 +2437,7 @@ mod tests {
                 "cargo nextest run --no-run --no-fail-fast",
                 ArgumentConflict,
             ),
+            ("cargo nextest run --no-run --max-fail 3", ArgumentConflict),
```
sunshowers, line 2440: Nice -- thanks for doing this!

nextest-runner/src/reporter/displayer.rs
```diff
@@ -1982,14 +1982,15 @@ fn write_final_warnings(
             if cancel_status == Some(CancelReason::TestFailure) {
                 writeln!(
                     writer,
-                    "{}: {}/{} {} {} not run due to {} (run with {} to run all tests)",
+                    "{}: {}/{} {} {} not run due to {} (run with {} to run all tests, or run with {})",
```
sunshowers, line 1985: Could you reword this to:

`specify --no-fail-fast to run all tests, or --max-fail`

?

nextest-runner/src/runner.rs
```diff
@@ -2078,7 +2078,9 @@ where
                 self.run_stats.on_test_finished(&run_statuses);
 
                 // should this run be cancelled because of a failure?
-                let fail_cancel = self.fail_fast && !run_statuses.last_status().result.is_success();
+                let fail_cancel = self
+                    .max_fail
+                    .map_or(false, |mf| self.run_stats.failed >= mf);
```
sunshowers, line 2083: hmm, so `failed` is one of several kinds of test errors -- there's also timed-out and success. I would add a `non_success_count` method to `RunStats` that sums them up and use that here.

site/src/docs/running.md
```diff
@@ -187,6 +187,9 @@ cargo nextest run -E 'platform(host)'
 `--no-fail-fast`
 : Do not exit the test run on the first failure. Most useful for CI scenarios.
 
+`--max-fail`
+: Number of tests that can fail before aborting the test run. Useful for uncovering multiple issues without having to run the whole test suite. Mutually exclusive with `--no-fail-fast`
```
sunshowers, line 191: A few suggest changes here:

1. Move `--max-fail` to above `--no-fail-fast`, and write it out as `--max-fail=N`.
2. I'd just drop the "mutually exclusive" bit, it's something people can discover for themselves :) Instead, to `--no-fail-fast` you can add something like "Equivalent to `--max-fail=1`."
3. Could you tag this with `<!-- md:version 0.9.86 -->` like `--run-ignored only` below? (We should have a tag for "unreleased" at some point but just saying the next version is okay for now).

sunshowers (commented):
Thank you for the PR! I agree that this idea makes a lot of sense.

nextest-runner/src/runner.rs
sunshowers: Yeah, sorry, I meant timed-out and exec errors. Thanks for catching that!

cargo-nextest/src/dispatch.rs
sunshowers: Hmm, so `--fail-fast` and `--no-fail-fast` overriding each other is part of a long-standing CLI convention that in case of that kind of binary choice, the last argument wins. I feel like `--max-fail` is slightly different from them, though I do see the argument. Ultimately it's a judgment call -- if we change our mind in the future, going from conflicts to overrides is easy while going from overrides to conflicts is harder.

nextest-runner/src/max_fail.rs
```diff
@@ -0,0 +1,38 @@
+use std::{fmt, str::FromStr};
+
+use crate::errors::MaxFailParseError;
+
+/// Type for the max-fail flag
```
sunshowers, line 5: Could you put this file inside config? That's where this usually lives (and would align with future work to make config support max-fail as well).

nextest-runner/src/runner.rs
```diff
@@ -1618,26 +1618,28 @@ pub struct RunStats {
...
-        if self.setup_scripts_failed > 0
-            || self.setup_scripts_exec_failed > 0
-            || self.setup_scripts_timed_out > 0
-        {
+        if self.failed_setup_script_count() > 0 {
             FinalRunStats::Failed(RunStatsFailureKind::SetupScript)
         } else if self.setup_scripts_initial_count > self.setup_scripts_finished_count {
             FinalRunStats::Cancelled(RunStatsFailureKind::SetupScript)
-        } else if self.failed > 0 || self.exec_failed > 0 || self.timed_out > 0 {
+        } else if self.failed_count() > 0 {
```
sunshowers, line 1642: Wonderful, thanks for this cleanup!

sunshowers (approved):
Thanks!

nextest-runner/src/runner.rs
```diff
@@ -2078,7 +2080,9 @@ where
                 self.run_stats.on_test_finished(&run_statuses);
 
                 // should this run be cancelled because of a failure?
-                let fail_cancel = self.fail_fast && !run_statuses.last_status().result.is_success();
+                let fail_cancel = self
+                    .max_fail
+                    .map_or(false, |mf| self.run_stats.failed_count() >= mf);
```
sunshowers, line 2085: Oh actually, one change required here -- I think there might be an off-by-one issue here.

What if `--max-fail=0` is passed in? I would expect this to mean "cancel test run after the first failure". But that seems to be `--max-fail=1`, and `--max-fail=0` would I think always trigger this condition.

Looking at [pytest](https://docs.pytest.org/en/7.1.x/how-to/failures.html#stopping-after-the-first-or-n-failures), though, it seems like they mean `--max-fail=2` to mean "cancel test run after 2 failures".

The pytest source code seems to suggest that [they treat `--max-fail=0` as `--no-fail-fast`](https://github.com/pytest-dev/pytest/blob/9d4f36d87dae9a968fb527e2cb87e8a507b0beb3/src/_pytest/main.py#L659C13-L661).

So we have a few options:

1. Do what pytest does. Remove `--max-fail=all`, use `--max-fail=0` to indicate `--no-fail-fast`, and keep everything else the same.
2. Use `>` rather than `>=` here, so shift meaning by 1.
3. Make `--max-fail=0` the same as 1.
4. Ban `--max-fail=0`, require that the minimum be 1.

Out of these, 1 and 4 are the most appealing. I think 1 makes the most sense just for similarity with pytest.

sunshowers (changes requested):
Could you update your PR with 1 or 4 (I'd lean towards 1 but your call), and also add a test for this?

nextest-runner/src/errors.rs
```diff
@@ -554,6 +554,22 @@ pub enum ToolConfigFileParseError {
     },
 }
 
+/// Error returned while parsing a [`MaxFail`](crate::config::MaxFail) input.
+#[derive(Clone, Debug, Error)]
+#[error("unrecognized value for max-fail: {input}\n(hint: expected either an integer or \"all\")")]
```
sunshowers, line 559: Thanks! Could this say "positive integer"?

sunshowers:
Looks great, thanks!

PR #2018, opened 2024-12-26
[DNM] Add documentation for hypothetical pre-timeout scripts

benesch:
In the spirit of documentation driven development, this commit adds docs for the pre-timeout scripts proposed in #1906.

I'll start prototyping an implementation soon. In the meantime, I figured we could start shaking out the major design questions by reviewing what's proposed in the documentation here.

cc @criccomini

### DIFF
--- a/site/src/docs/configuration/scripts.md
+++ b/site/src/docs/configuration/scripts.md
@@ -3,25 +3,34 @@ icon: material/ray-start-arrow
 status: experimental
 ---
 
-# Setup scripts
+# Scripts
 
 <!-- md:version 0.9.59 -->
 
-!!! experimental "Experimental: This feature is not yet stable"
+!!! experimental "Experimental: Setup scripts are not yet stable"
 
     - **Enable with:** Add `experimental = ["setup-scripts"]` to `.config/nextest.toml`
     - **Tracking issue:** [#978](https://github.com/nextest-rs/nextest/issues/978)
 
-Nextest supports running _setup scripts_ before tests are run. Setup scripts can be scoped to
-particular tests via [filtersets](../filtersets/index.md).
+!!! experimental "Experimental: Pre-timeout scripts are not yet stable"
 
-Setup scripts are configured in two parts: _defining scripts_, and _setting up rules_ for when they should be executed.
+    - **Enable with:** Add `experimental = ["pre-timeout-scripts"]` to `.config/nextest.toml`
+    - **Tracking issue:** [#978](https://github.com/nextest-rs/nextest/issues/978)
+
+Nextest supports running _scripts_ when certain events occur during a test run. Scripts can be scoped to particular tests via [filtersets](../filtersets/index.md).
+
+Nextest currently recognizes two types of scripts:
+
+  * _Setup scripts_, which are executed at the start of a test run.
+  * _Pre-timeout scripts_, which are executed before nextest begins terminating a test that has exceeded its timeout. 
+
+Scripts are configured in two parts: _defining scripts_, and _setting up rules_ for when they should be executed.
 
 ## Defining scripts
 
-Setup scripts are defined using the top-level `script` configuration. For example, to define a script named "my-script", which runs `my-script.sh`:
+Scripts are defined using the top-level `script` configuration. For example, to define a script named "my-script", which runs `my-script.sh`:
 
-```toml title="Setup script definition in <code>.config/nextest.toml</code>"
+```toml title="Script definition in <code>.config/nextest.toml</code>"
 [script.my-script]
 command = 'my-script.sh'
 ```
@@ -36,16 +45,16 @@ command = 'script.sh -c "Hello, world!"'
 command = ['script.sh', '-c', 'Hello, world!']
 ```
 
-Setup scripts can have the following configuration options attached to them:
+Scripts can have the following configuration options attached to them:
 
-- **`slow-timeout`**: Mark a setup script [as slow](../features/slow-tests.md) or [terminate it](../features/slow-tests.md#terminating-tests-after-a-timeout), using the same configuration as for tests. By default, setup scripts are not marked as slow or terminated (this is different from the slow timeout for tests).
-- **`leak-timeout`**: Mark setup scripts [leaky](../features/leaky-tests.md) after a timeout, using the same configuration as for tests. By default, the leak timeout is 100ms.
+- **`slow-timeout`**: Mark a script [as slow](../features/slow-tests.md) or [terminate it](../features/slow-tests.md#terminating-tests-after-a-timeout), using the same configuration as for tests. By default, scripts are not marked as slow or terminated (this is different from the slow timeout for tests).
+- **`leak-timeout`**: Mark scripts [leaky](../features/leaky-tests.md) after a timeout, using the same configuration as for tests. By default, the leak timeout is 100ms.
 - **`capture-stdout`**: `true` if the script's standard output should be captured, `false` if not. By default, this is `false`.
 - **`capture-stderr`**: `true` if the script's standard error should be captured, `false` if not. By default, this is `false`.
 
 ### Example
 
-```toml title="Advanced setup script definition"
+```toml title="Advanced script definition"
 [script.db-generate]
 command = 'cargo run -p db-generate'
 slow-timeout = { period = "60s", terminate-after = 2 }
@@ -56,7 +65,7 @@ capture-stderr = false
 
 ## Setting up rules
 
-In configuration, you can create rules for when to use scripts on a per-profile basis. This is done via the `profile.<profile-name>.scripts` array. For example, you can set up a script that generates a database if tests from the `db-tests` package, or any packages that depend on it, are run.
+In configuration, you can create rules for when to use scripts on a per-profile basis. This is done via the `profile.<profile-name>.scripts` array. For example, you can configure a setup script that generates a database if tests from the `db-tests` package, or any packages that depend on it, are run.
 
 ```toml title="Basic rules"
 [[profile.default.scripts]]
@@ -66,7 +75,7 @@ setup = 'db-generate'
 
 (This example uses the `rdeps` [filterset](../filtersets/index.md) predicate.)
 
-Setup scripts can also filter based on platform, using the rules listed in [_Specifying platforms_](../configuration/specifying-platforms.md):
+Scripts can also filter based on platform, using the rules listed in [_Specifying platforms_](../configuration/specifying-platforms.md):
 
 ```toml title="Platform-specific rules"
 [[profile.default.scripts]]
@@ -82,13 +91,32 @@ filter = 'test(/^script_tests::/)'
 setup = ['script1', 'script2']
 ```
 
+Executing pre-timeout scripts follows the same pattern. For example, you can configure a pre-timeout script for every test that contains `slow` in its name.
+
+```toml title="Basic pre-timeout rules"
+[[profile.default.scripts]]
+filter = 'test(slow)'
+pre-timeout = 'capture-backtrace'
+```
+
+A single rule can specify any number of setup scripts and any number of pre-timeout scripts.
+
+```toml title="Combination rules"
+[[profile.default.scripts]]
+filter = 'test(slow)'
+setup = ['setup-1', 'setup-2']
+pre-timeout = ['pre-timeout-1', 'pre-timeout-2']
+```
+
 ## Script execution
+
+### Setup scripts
 
 A given setup script _S_ is only executed if the current profile has at least one rule where the `filter` and `platform` predicates match the current execution environment, and the setup script _S_ is listed in `setup`.
 
 Setup scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any setup script exits with a non-zero exit code, the entire test run is terminated.
 
-### Environment variables
+#### Environment variables
 
 Setup scripts can define environment variables that will be exposed to tests that match the script. This is done by writing to the `$NEXTEST_ENV` environment variable from within the script.
 
@@ -125,6 +153,18 @@ fn my_env_test() {
     assert_eq!(std::env::var("MY_ENV_VAR"), Ok("Hello, world!".to_string()));
 }
 ```
+
+### Pre-timeout scripts
+
+A given pre-timeout script _S_ is executed when the current profile has at least one rule where the `platform` predicates match the current execution environment, the script _S_ is listed in `pre-timeout`, and a test matching the `filter` has reached its configured timeout.
+
+Pre-timeout scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any pre-timeout script exits with a non-zero exit code, an error is logged but the test run continues.
+
+Nextest sets the following environment variables when executing a pre-timeout script:
+
+  * **`NEXTEST_PRE_TIMEOUT_TEST_PID`**: the ID of the process running the test.
+  * **`NEXTEST_PRE_TIMEOUT_TEST_NAME`**: the name of the running test.
+  * **`NEXTEST_PRE_TIMEOUT_TEST_BINARY`**: the name of the binary running the test.
 
 ## Setup scripts in JUnit output
 


site/src/docs/configuration/scripts.md
```diff
@@ -126,6 +154,18 @@ fn my_env_test() {
...
+
+A given pre-timeout script _S_ is executed when the current profile has at least one rule where the `platform` predicates match the current execution environment, the script _S_ is listed in `pre-timeout`, and a test matching the `filter` has reached its configured timeout.
+
+Pre-timeout scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any pre-timeout script exits with a non-zero exit code, an error is logged but the test run continues.
+
+Nextest sets the following environment variables when executing a pre-timeout script:
+
+  * **`NEXTEST_PRE_TIMEOUT_TEST_PID`**: the ID of the process running the test.
+  * **`NEXTEST_PRE_TIMEOUT_TEST_NAME`**: the name of the running test.
+  * **`NEXTEST_PRE_TIMEOUT_TEST_BINARY`**: the name of the binary running the test.
```
sunshowers, line 167: I'd call the last one `BINARY_ID`. Maybe also pass in the components of the binary ID as separate variables for convenience.

site/src/docs/configuration/scripts.md
```diff
@@ -126,6 +154,18 @@ fn my_env_test() {
 }
 ```
 
+### Pre-timeout scripts
+
+A given pre-timeout script _S_ is executed when the current profile has at least one rule where the `platform` predicates match the current execution environment, the script _S_ is listed in `pre-timeout`, and a test matching the `filter` has reached its configured timeout.
```
sunshowers, line 159: One of the things it would be useful to add here is a practical use case that's outlined. For setup scripts, database setup or preparing artifacts for individual test use are illustrative examples. For pre-timeout scripts, it would be good to mention grabbing stack traces and logs. That kind of use case would also help shape their functionality.

It would also be useful to cover some more of the points [from my post](https://github.com/nextest-rs/nextest/discussions/1906#discussioncomment-11468794):

> Should the script be responsible for killing the test, or should nextest do so (after waiting for the script to exit)? I'd lean towards the latter.

Covering this would be useful -- I'm still leaning towards killing the process group anyway, but saying it explicitly would be nice.

> Do we still want to allow a grace period (SIGTERM before SIGKILL) in this case?

This is worth mentioning as well. Pre-timeout scripts add a bunch of states to the state machine, because each a unit of work now has a sidecar process involved as well. One approach is to add the pre-timeout script to the process group (Unix) / job object (Windows) -- that would provide fairly clear semantics, I think. The script and the test then live and die together -- the code that waits for the child process to exit can now wait on both processes.

> What should the cwd of the script process be?

This should just be the same as the cwd of the test, I'm quite sure.

> How should information about the script being run be communicated to the user?

The states presented to the user blow up a bit here.

One of the things that's worth thinking about, I think, is passing through stdout and stderr -- unlike setup scripts, pre-timeout scripts aren't executed serially, so capturing stdout/stderr generally makes a lot of sense.

But! There's also a compelling use case for pre-timeout scripts: to put you into an interactive debugger, effectively acting as a break point. In that case, you do want to pass through stdout and stderr, and you want to keep processing existing tests in the background, but (I think) not start new ones. Even if we don't solve it here, it's worth keeping that case in mind.

> Does nextest need to create a directory for the script to write logs and such out to? Even if not necessary, is it a good to have?

I really would like this, personally. That information can then be attached in all sorts of ways, e.g. to JUnit reports.

I think it would also be useful to have a protocol for the pre-timeout script to give information back to nextest -- for example, the list of files written out and/or to include. This can just be a file the script writes to, similar to `NEXTEST_ENV` for setup scripts.

---

The exact details for much of this can be outlined in an architecture doc (this is more usage-oriented), but a quick summary of a lot of this would be useful.

sunshowers (commented):
Thank you for writing this up!

benesch:
Thanks for this great feedback, @sunshowers! Will take an another spin through the docs and start working in answers to the questions you mentioned. I've got some long haul flights coming up this week and next, and this should make for a perfect plane project.

site/src/docs/configuration/scripts.md
```diff
@@ -126,6 +154,18 @@ fn my_env_test() {
...
+
+Pre-timeout scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any pre-timeout script exits with a non-zero exit code, an error is logged but the test run continues.
```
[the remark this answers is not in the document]
criccomini: I'm partial to just one script. Users can wrap multiple scripts themselves if need be.

site/src/docs/scripts/pre-timeout.md
```diff
@@ -0,0 +1,69 @@
...
+The pre-timeout script is not responsible for terminating the test process, but it is permissible for it to do so.
+
+Nextest executes pre-timeout scripts with the same working directory as the test and sets the following variables in the script's environment:
+
+* **`NEXTEST_PRE_TIMEOUT_TEST_PID`**: the ID of the process running the test.
+* **`NEXTEST_PRE_TIMEOUT_TEST_NAME`**: the name of the running test.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID`**: the ID of the binary in which the test is located.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_PACKAGE_NAME`**: the package name component of the binary ID.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_NAME`**: the name component of the binary ID, if known.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_KIND`**: the kind component of the binary ID, if known.
```
benesch, line 67: @sunshowers would love your input on whether these are the right names for the `_BINARY_ID_*` environment variables.

site/src/docs/scripts/pre-timeout.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Pre-timeout script execution
+
+A given pre-timeout script _S_ is executed when the current profile has at least one rule where the `platform` predicates match the current execution environment, the script _S_ is listed in `pre-timeout`, and a test matching the `filter` has reached its configured timeout.
+
+Pre-timeout scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any pre-timeout script exits with a non-zero exit code, an error is logged but the test run continues.
+
+Nextest will proceed with graceful termination of the test only once the pre-timeout script terminates. See [_How nextest terminates tests_](#defining-pre-timeout-scripts). If the pre-timeout script itself is slow, nextest will apply the same termination protocol to the pre-timeout script.
+
+The pre-timeout script is not responsible for terminating the test process, but it is permissible for it to do so.
```
benesch, line 58: Beginning to address this question:

> > Should the script be responsible for killing the test, or should nextest do so (after waiting for the script to exit)? I'd lean towards the latter.
> 
> Covering this would be useful -- I'm still leaning towards killing the process group anyway, but saying it explicitly would be nice.

and this one:

> > Do we still want to allow a grace period (SIGTERM before SIGKILL) in this case?
> 
> This is worth mentioning as well. Pre-timeout scripts add a bunch of states to the state machine, because each a unit of work now has a sidecar process involved as well. One approach is to add the pre-timeout script to the process group (Unix) / job object (Windows) -- that would provide fairly clear semantics, I think. The script and the test then live and die together -- the code that waits for the child process to exit can now wait on both processes.

site/src/docs/scripts/pre-timeout.md
```diff
@@ -0,0 +1,69 @@
...
+
+Nextest executes pre-timeout scripts with the same working directory as the test and sets the following variables in the script's environment:
```
benesch, line 60: Added a clause here specifying the CWD:

> > What should the cwd of the script process be?
> 
> This should just be the same as the cwd of the test, I'm quite sure.

site/src/docs/scripts/pre-timeout.md
```diff
@@ -0,0 +1,69 @@
...
+```toml title="Script definition in <code>.config/nextest.toml</code>"
+[script.pre-timeout.my-script]
+command = 'my-script.sh'
+```
+
+See [_Defining scripts_](index.md#defining-scripts) for options that are common to all scripts.
+
+Pre-timeout scripts do not support additional configuration options.
+
+Notably, pre-timeout scripts always capture stdout and stderr. Support for not capturing stdout and stderr may be added in the future in order to support use cases like interactive debugging of a hung test.
```
benesch, line 34: Starting to address this:

> > How should information about the script being run be communicated to the user?
> 
> The states presented to the user blow up a bit here.
> 
> One of the things that's worth thinking about, I think, is passing through stdout and stderr -- unlike setup scripts, pre-timeout scripts aren't executed serially, so capturing stdout/stderr generally makes a lot of sense.
> 
> But! There's also a compelling use case for pre-timeout scripts: to put you into an interactive debugger, effectively acting as a break point. In that case, you do want to pass through stdout and stderr, and you want to keep processing existing tests in the background, but (I think) not start new ones. Even if we don't solve it here, it's worth keeping that case in mind.

I was thinking to keep things simple for the initial implementation and force that stdout/stderr are captured. @sunshowers does that make sense to you?

> How should information about the script being run be communicated to the user?

Flagging that I haven't forgotten about this! I think I'll need to get a little further along in the implementation before I can answer this question, though. Happy to take any advice you have in the meantime, though, @sunshowers, if there's a particular UX you're imagining!

site/src/docs/scripts/pre-timeout.md
```diff
@@ -0,0 +1,69 @@
...
+Nextest executes pre-timeout scripts with the same working directory as the test and sets the following variables in the script's environment:
+
+* **`NEXTEST_PRE_TIMEOUT_TEST_PID`**: the ID of the process running the test.
+* **`NEXTEST_PRE_TIMEOUT_TEST_NAME`**: the name of the running test.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID`**: the ID of the binary in which the test is located.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_PACKAGE_NAME`**: the package name component of the binary ID.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_NAME`**: the name component of the binary ID, if known.
+* **`NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_KIND`**: the kind component of the binary ID, if known.
+
+<!-- TODO: a protocol for writing script logs to a file and telling nextest to attach them to JUnit reports? -->
```
benesch, line 69: > > Does nextest need to create a directory for the script to write logs and such out to? Even if not necessary, is it a good to have?
> 
> I really would like this, personally. That information can then be attached in all sorts of ways, e.g. to JUnit reports.
> 
> I think it would also be useful to have a protocol for the pre-timeout script to give information back to nextest -- for example, the list of files written out and/or to include. This can just be a file the script writes to, similar to `NEXTEST_ENV` for setup scripts.
> 
> The exact details for much of this can be outlined in an architecture doc (this is more usage-oriented), but a quick summary of a lot of this would be useful.

@sunshowers are you comfortable deferring this to future work? I'd like to try to keep the initial implementation as focused as possible, and this bit seems a hunk of work that's easy to split off.

site/src/docs/configuration/scripts.md
benesch: Sounds good! I pushed up a revision that begins to tackle these questions. I also split out individual discussions for each of your questions, since this discussion is now on an orphaned hunk of code.

benesch (commented):
Finally found the time to get back to this! @sunshowers @criccomini thanks for all the time so far—have a bunch more questions for y'all!

My internet access is quite spotty for the week, but I have a fair bit of free time. I'll start working on the first draft of an implementation of what I've proposed here while I'm offline, and I'll check in on the discussion here as my internet access permits!

site/src/docs/scripts/pre-timeout.md
benesch: I think I may need to get further into the code to really understand the complexity here, but it's not immediately obvious to me why we need to couple the script and test lifetimes together or put the processes in the same pgrp. Naively it seems straightforward to just apply the graceful termination logic to the pre-timeout script and then move on to apply the graceful termination logic to the test.

site/src/docs/configuration/scripts.md
```diff
@@ -82,13 +91,32 @@ filter = 'test(/^script_tests::/)'
...
 ## Script execution
 
+### Setup scripts
+
 A given setup script _S_ is only executed if the current profile has at least one rule where the `filter` and `platform` predicates match the current execution environment, and the setup script _S_ is listed in `setup`.
 
 Setup scripts are executed serially, in the order they are defined (_not_ the order they're specified in the rules). If any setup script exits with a non-zero exit code, the entire test run is terminated.
 
-### Environment variables
+#### Environment variables
```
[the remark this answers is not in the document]
sunshowers: Thanks!

site/src/docs/configuration/scripts.md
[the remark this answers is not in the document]
sunshowers: Nextest already has precedence rules for overrides (first matching override wins), so we can just use that.

site/src/docs/scripts/pre-timeout.md
sunshowers: I think I'd match the [filterset DSL](https://nexte.st/docs/filtersets/reference/). I'd change:

* `NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_PACKAGE_NAME` -> `NEXTEST_PRE_TIMEOUT_TEST_PACKAGE`
* `NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_NAME` -> `NEXTEST_PRE_TIMEOUT_TEST_BINARY`
* `NEXTEST_PRE_TIMEOUT_TEST_BINARY_ID_KIND` -> `NEXTEST_PRE_TIMEOUT_TEST_BINARY_KIND`. I think `BINARY_KIND` here rather than `KIND` because the latter is a bit confusing.

site/src/docs/scripts/pre-timeout.md
sunshowers: It's definitely possible to do that, but my concern is just that it blows up the state machine even further. The state machine for an individual test already has on the order of 40-50 states today once you consider all the combinations of `select` branches that are or aren't done at any moment.

Adding a "start script -> read stdout/stderr from script -> wait on script + test to exit -> check for the script leaking handles" already blows it up quite a bit, though the benefits outweigh the complexity.

Adding "SIGTERM to script -> maybe SIGKILL -> script done -> SIGTERM to test -> maybe SIGKILL -> test done" adds even more states to the state machine, and I'm not sure this pays its way compared to killing the script and the test at the same time.

But, I agree that this decision can be made based on how complex the implementation gets.

site/src/docs/scripts/pre-timeout.md
sunshowers: > I was thinking to keep things simple for the initial implementation and force that stdout/stderr are captured. @sunshowers does that make sense to you?

Yeah, this is ok, keep it simple for now.

> Flagging that I haven't forgotten about this! I think I'll need to get a little further along in the implementation before I can answer this question, though. Happy to take any advice you have in the meantime, though, @sunshowers, if there's a particular UX you're imagining!

So there's several points to surface user information here.

**At the moment the test has timed out:**

* Today we print out a `TERMINATING` line: https://github.com/nextest-rs/nextest/blob/e595f3f34f3c063a3bfeed8e024fa2cc6777afa6/nextest-runner/src/reporter/displayer/imp.rs#L347-L364
* If there's a pre-timeout script I'd add "running pre-timeout script foo against test" on a second line. We have a format for this already, used for Windows abort codes (it's aligned so that the `-` is in the same column as `]`) https://github.com/nextest-rs/nextest/blob/e595f3f34f3c063a3bfeed8e024fa2cc6777afa6/nextest-runner/src/reporter/displayer/snapshots/nextest_runner__reporter__displayer__imp__windows_tests__stack_violation_code.snap#L6

**Once the script is done executing:**

We should print the test's output, as well as the script's output, with the two being clearly distinguished from each other.

**In information queries:** (i.e. `t`, or ctrl-t/SIGINFO where available):

We should say something like "terminating due to timeout" -> "running pre-timeout script foo", and carry around all the details for foo just like we would for any other unit.

To do this we'd need to expand the terminating state to carry this information, I think: https://github.com/nextest-rs/nextest/blob/e595f3f34f3c063a3bfeed8e024fa2cc6777afa6/nextest-runner/src/reporter/events.rs#L1023-L1045

Maybe in an `Option<Box<T>>` field.

site/src/docs/scripts/pre-timeout.md
sunshowers: Yeah, fine to defer to the future. I just want to make sure we don't box ourselves into making this impossible in the future.

sunshowers:
Thanks for the wait! I've gone through and answered all the questions. I think the design is in generally good shape, and you're asking all the right questions :D

nextest-runner/src/runner/executor.rs
```diff
@@ -573,6 +574,269 @@ impl<'a> ExecutorContext<'a> {
...
+    #[instrument(level = "debug", skip(self, resp_tx, req_rx))]
+    async fn run_pre_timeout_script_inner<'b>(
+        &self,
+        script: PreTimeoutScriptPacket<'a, 'b>,
+        stopwatch: &mut StopwatchStart,
+        resp_tx: &UnboundedSender<ExecutorEvent<'a>>,
+        req_rx: &mut UnboundedReceiver<RunUnitRequest<'a>>,
+    ) -> Result<InternalPreTimeoutScriptExecuteStatus, ChildStartError> {
+        // XXX: this function is 95% the same as run_setup_script_inner. Do we
+        // need to try to factor out a shared helper method?
```
benesch, line 620: @sunshowers would love your quick take on this question.

site/src/docs/scripts/pre-timeout.md
benesch: Makes sense! I ended up with a slightly different approach than you proposed. Instead of adding an optional pre-timeout field to the `Terminating` state, I added a new top-level `UnitState::PreTimeout` state, which itself contains a `script_state: UnitState` field. This allows the pre-timeout script itself to transition between the standard unit states (running, slow, terminating, exited) with minimal code duplication.

Here's how the output looks in this initial implementation. 

Basic statuses:

```
$ ~/Sites/nextest/target/debug/cargo-nextest nextest run --no-fail-fast
info: experimental features enabled: setup-scripts, pre-timeout-scripts
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.00s
────────────
 Nextest run ID d0f8b05a-5a82-4fa7-9f3c-5cccd3cbccbb with nextest profile: default
    Starting 2 tests across 1 binary
             [ 00:00:00]                                                                                                                                                                                                                   0/2:             SETUP [      1/1] my-script: ./setup.sh
setting up!
  SETUP PASS [   0.015s] my-script: ./setup.sh
        SLOW [>  1.000s] nextest-play tests::it_no_works
        SLOW [>  1.000s] nextest-play tests::it_works
 TERMINATING [>  2.000s] nextest-play tests::it_no_works
      PRETMT             nextest-play tests::it_works
     TIMEOUT [   2.007s] nextest-play tests::it_no_works
──── STDOUT:             nextest-play tests::it_no_works

running 1 test

 PRETMT SLOW [>  1.000s] nextest-play tests::it_works
 PRETMT PASS [   1.233s] nextest-play tests::it_works
──── STDOUT:             gdb-dump: ./pre-timeout.sh
pre-timeout sample stdout
──── STDERR:             gdb-dump: ./pre-timeout.sh
pre-timeout sample stderr

 TERMINATING [>  2.000s] nextest-play tests::it_works
     TIMEOUT [   3.240s] nextest-play tests::it_works
──── STDOUT:             nextest-play tests::it_works

running 1 test
test sample stdout
──── STDERR:             nextest-play tests::it_works
test sample stderr

────────────
     Summary [   3.257s] 2 tests run: 0 passed, 2 timed out, 0 skipped
     TIMEOUT [   2.007s] nextest-play tests::it_no_works
     TIMEOUT [   3.240s] nextest-play tests::it_works
error: test run failed
```

A test with an activated pre-timeout script transitions like so:

  * Running (not printed)
  * Slow (printed as `SLOW`)
  * Pre-timeout script starting (printed as `PRETMT`)
  * Pre-timeout script itself slow (printed as `PRETMT SLOW`)
  * Pre-timeout script pass/fail (printed as `PRETMT PASS`/`PRETMT FAIL`, followed by the pre-timeout script's stdout/stderr)
  * Terminating (printed as `TERMINATING`)
  * Exit (printed as `TIMEOUT`, followed by the test's stdout/stderr)

If you request status while the pre-timeout script is executing, you see something like this:

```
* 1/1:    nextest-play tests::it_works
  status: test running for 2.493s as PID 41610 (marked slow after 1.000s)
  note:   test has reached timeout, pre-timeout script is running:
      status: script running for 0.485s as PID 41641
      stdout:
        pre-timeout sample stdout
      stderr:
        pre-timeout sample stderr
```

I'm by no means married to any of this. It's just a result of my initial attempt to balance "clear to end users" and "straightforward to implement."

benesch (commented):
As promised, I've whipped up an initial implementation! The code here is largely feature complete and works well in my manual testing.

I've not yet written any automated tests. I figured it would be more efficient to defer writing tests until after an initial few rounds of reviews. I didn't want to burn cycles locking in behavior that we wind up wanting to change.

@sunshowers let me know if there's anything I can do to make this easier to review. The PR's gotten fairly large. I've been pushing up my raw commits so that you can use GitHub to skip over commits you've already reviewed. If you'd prefer, I can start squashing/splitting my raw commits into a tidy series of semantic changes; the downside, of course, is that I'd have to force-push the history and you wouldn't as easily be able to see what changed after every force-push.

sunshowers:
Thanks! FYI this is on my queue to look at this weekend -- have a lot of stuff at work this week. I like the pretimeout state :)

benesch:
> FYI this is on my queue to look at this weekend -- have a lot of stuff at work this week.

No rush on my end at all—take as long as you need!

> I like the pretimeout state :)

🎉 !


> It's currently ok (I was expecting the impl to be this large), but I think before landing it would be good to get into a series of smaller commits.

Sounds good! I'm happy to polish up the commit series whenever works best for you.

sunshowers:
Still in the queue to review -- higher priority stuff came up but I'm hoping to carve out some time this weekend.

benesch:
No problem at all, @sunshowers, and thanks for the update!

benesch:
@sunshowers no pressure, but just wanted to check in here!

nextest-runner/src/runner/executor.rs
sunshowers: 95% the same -- what are the differences? Looking at the function itself, yes -- I'd love to see some common code factored out.

nextest-runner/src/runner/executor.rs
```diff
@@ -1026,6 +1313,55 @@ impl<'a> SetupScriptPacket<'a> {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(super) struct PreTimeoutScriptPacket<'a, 'b> {
+    script_id: ScriptId,
+    test: TestPacket<'a>,
+    test_pid: u32,
+    test_stopwatch: &'b StopwatchStart,
```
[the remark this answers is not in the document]
benesch: Yeah, for sure. I didn't love the extra lifetime parameter either.

nextest-runner/src/config/scripts.rs
```diff
@@ -243,9 +243,111 @@ impl<'profile> SetupScriptExecuteData<'profile> {
...
+/// Represents a to-be-run pre-timeout script command with a certain set of arguments.
+pub(crate) struct PreTimeoutScriptCommand {
+    /// The command to be run.
+    command: std::process::Command,
+    /// Double-spawn context.
+    double_spawn: Option<DoubleSpawnContext>,
+}
+
+impl PreTimeoutScriptCommand {
+    /// Creates a new `PreTimeoutScriptCommand` for a setup script.
```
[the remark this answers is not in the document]
benesch: Whoops, thanks! Will fix.

nextest-runner/src/reporter/displayer.rs
```diff
@@ -1608,6 +1675,42 @@ impl<'a> TestReporterImpl<'a> {
...
+                // Print test status.
+                print_running(*pid, *time_taken, Some(*slow_after))?;
+
+                // Print note about running pre-timeout script.
+                write!(
+                    writer,
+                    "{}:   test has reached timeout, pre-timeout script is running:",
+                    "note".style(self.styles.count)
+                )?;
+
```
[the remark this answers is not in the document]
benesch: For sure. Planning to write a full suite of tests the next moment I have to look at this. Probably this weekend.

benesch:
My turn to be a bit behind: busy starting a new job! Hoping to have time to get back to this shortly but almost certainly not before the end of the week.

sunshowers:
I know things have been a bit behind here, but wanted to let you know that I ended up using a lot of this implementation for the new [wrapper script](https://nexte.st/docs/configuration/wrapper-scripts/) support. So thank you for implementing this!

Hopefully this implementation becomes relatively small as a result.

benesch:
Oh, that's great to hear, @sunshowers! Sorry that I wound up being too busy to push this over the finish line.

benesch:
I actually wonder if wrapper scripts are all we need here? For SlateDB, the wrapper script could be responsible for handling the SIGTERM and invoking `gdb` before exiting.

sunshowers:
Yeah wrapper scripts can definitely fulfill this role! But ease of use is an important consideration:

* Nextest manages timeouts but has no clear way of communicating that to the wrapper script
* Writing a shell script that does this with a trap instruction is possible but not easy
* There isn't a good story on Windows
* Wrapper scripts are a power tool that are easy to get wrong

So overall, explicit support for pre-timeout scripts is still valuable.

benesch:
> * Writing a shell script that does this with a trap instruction is possible but not easy
> * Wrapper scripts are a power tool that are easy to get wrong

Indeed! I rigged this up for SlateDB just now and it is a bit fiddly: https://github.com/slatedb/slatedb/pull/363/files#diff-8b5d6d952cf5fddb1cedc6bb8c4aebd2048ac395f30b43e80f082d7d23a58d64

> So overall, explicit support for pre-timeout scripts is still valuable.

Makes sense. Realistically I don't think I'm going to be able to make the time to continue this work in the coming months. Is it worth leaving the PR open in case you or someone else has time to pick it up? Or better to just close it out until someone else needs it?

sunshowers:
Let's leave this open, I'll try and finish it up some time soon if you don't get to it. I feel like around 50% of this landed already.

benesch:
Sounds good—thanks, @sunshowers!

PR #2063, opened 2025-01-11
feat(config): add max-fail option to configuration file

### LINKED ISSUE #1944: Add configuration for max-fail
`--max-fail` was added in https://github.com/nextest-rs/nextest/pull/1926. We should also add configuration for it, similar to how `fail-fast = true/false` currently works.

Need to figure out how `fail-fast` would conflict with `max-fail`.

Jayllyz:
Closes #1944


nextest-runner/default-config.toml
```diff
@@ -66,6 +66,9 @@ final-status-level = "flaky"
 # Can be overridden through the `--failure-output` option.
 failure-output = "immediate"
 
+# "max-fail" defines the maximum number of test failures that are allowed before test execution is stopped.
+max-fail = 1
```
sunshowers, line 70: So I think this should actually be part of the `fail-fast` field. This field currently supports true or false, but it should also support

```toml
fail-fast = { max-fail = 1 }
```

This addresses issues like precedence, and matches the CLI behavior where `--fail-fast` and `--max-fail=1` cannot be specified together.

nextest-runner/src/config/max_fail.rs
```diff
@@ -83,4 +130,126 @@ mod tests {
             MaxFail::from_str(input).expect_err(&format!("expected input '{input}' to fail"));
         }
     }
+
+    #[test_case(
```
sunshowers, line 134: These tests are fantastic, thank you!

sunshowers (commented):
Thank you for the PR!

nextest-runner/default-config.toml
Jayllyz: Makes sense, Im going to try to do it

nextest-runner/src/config/fail_fast.rs
```diff
@@ -0,0 +1,224 @@
+use crate::config::max_fail::MaxFail;
```
sunshowers, line 1:
```suggestion
// Copyright (c) The nextest Contributors
// SPDX-License-Identifier: MIT OR Apache-2.0

use crate::config::max_fail::MaxFail;
```

sunshowers (approved):
Looks great, thank you! I have a couple minor changes I'll take care of.

sunshowers:
I've combined max-fail and fail-fast into a single struct, and added tests for error messages (i.e. not just that there are errors).

sunshowers:
This is now out in nextest 0.9.89. Thanks for your contribution!

Issue #2067, opened 2025-01-14
Bug: error: writing to archive: caused by: unsupported parameter


Leandros:
### Description of the issue

**Description:**

Archival will error out trying to create the zstd encoder if `libzstd` is not build with multithread support enabled (which is disabled by default, as seen in the Zstandard [README.md here](https://github.com/facebook/zstd/blob/80af41e08a630946a75a5cda9e4cdf192247f20a/lib/README.md?plain=1#L22-L26)).

**Steps to reproduce:**
1. Checkout the latest stable release (`cargo-nextest-0.9.87`)
2. Build `libzstd` with the default configuration from source with: `make lib`
3. Build `cargo-nextest` with: `cargo build  --release --locked`
4. Run `cargo-nextest nextest archive -p nextest-runner --archive-file test.tar.zst`

### Expected outcome

Archive succeeds, despite `libzstd` not supporting multithreaded processing.

### Actual result

I receive the following error:

```
multithread: Custom { kind: Other, error: "Unsupported parameter" }
error: error creating archive `test.tar.zst`

Caused by:
  error writing to archive

Caused by:
  Unsupported parameter
```

### Nextest version

```text
cargo-nextest 0.9.87 (fee71ad7b 2024-12-17)
release: 0.9.87
commit-hash: fee71ad7ba08be06519cf9eeb495c6fce0a2e888
commit-date: 2024-12-17
host: x86_64-unknown-linux-gnu
```

### Additional context

I have debugged this, and tracked it down to 

```rust
impl<'a, W: Write> Archiver<'a, W> {
    #[expect(clippy::too_many_arguments)]
    fn new(
    // ...
    ) -> Result<Self, ArchiveCreateError> {
        let buf_writer = BufWriter::new(writer);
        let builder = match format {
            ArchiveFormat::TarZst => {
                // ...
                encoder
                    .multithread(get_num_cpus() as u32)
                    .inspect_err(|err| eprintln!("multithread: {err:?}")) // <-- line added to pin it down which parameter was not supported
                    .map_err(ArchiveCreateError::OutputArchiveIo)?;
                tar::Builder::new(encoder)
            }
        };
    // ...
    }
// ...
}
```

By itself, the error message was too generic and couldn't pin the error down. I further recommend to improve your error handling to make it easier to track those bugs down.

The following patch fixes the issue:

```patch
diff --git a/nextest-runner/src/reuse_build/archiver.rs b/nextest-runner/src/reuse_build/archiver.rs
[...]

PR #2068, opened 2025-01-14
fix: support for non-multithread libzstd (#2067)

Leandros:
Fix for the bug report in #2067

### DIFF
--- a/nextest-runner/src/reuse_build/archiver.rs
+++ b/nextest-runner/src/reuse_build/archiver.rs
@@ -217,9 +217,12 @@ impl<'a, W: Write> Archiver<'a, W> {
                 encoder
                     .include_checksum(true)
                     .map_err(ArchiveCreateError::OutputArchiveIo)?;
-                encoder
-                    .multithread(get_num_cpus() as u32)
-                    .map_err(ArchiveCreateError::OutputArchiveIo)?;
+                if let Err(err) = encoder.multithread(get_num_cpus() as u32) {
+                    tracing::warn!(
+                        ?err,
+                        "libzstd compiled without multithreading. defaulting to single-thread."
+                    );
+                }
                 tar::Builder::new(encoder)
             }
         };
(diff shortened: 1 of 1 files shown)


nextest-runner/src/reuse_build/archiver.rs
```diff
@@ -217,9 +217,7 @@ impl<'a, W: Write> Archiver<'a, W> {
                 encoder
                     .include_checksum(true)
                     .map_err(ArchiveCreateError::OutputArchiveIo)?;
-                encoder
-                    .multithread(get_num_cpus() as u32)
-                    .map_err(ArchiveCreateError::OutputArchiveIo)?;
+                let _err_ignored = encoder.multithread(get_num_cpus() as u32);
```
sunshowers, line 220: Thank you for debugging this and for the PR!

Could you add a warning about this rather than ignoring the error? Just using `tracing::warn!` would be fine. (If you don't like `warn!` then `info!` is ok.)

Leandros: Sure! I wasn't to sure which methodology is best for giving a warning. I'll add it asap.

Leandros: Updated, and rebased with `main` as well.

Leandros: @sunshowers: Updated and rebased with `main`, as well.

nextest-runner/src/reuse_build/archiver.rs
```diff
@@ -217,9 +217,12 @@ impl<'a, W: Write> Archiver<'a, W> {
                 encoder
                     .include_checksum(true)
                     .map_err(ArchiveCreateError::OutputArchiveIo)?;
-                encoder
-                    .multithread(get_num_cpus() as u32)
-                    .map_err(ArchiveCreateError::OutputArchiveIo)?;
+                if let Err(err) = encoder.multithread(get_num_cpus() as u32) {
+                    tracing::warn!(
+                        ?err,
+                        "libzstd compiled without multithreading. defaulting to single-thread."
```
sunshowers, line 223: To match prevailing standards:

```suggestion
                        "libzstd compiled without multithreading, defaulting to single-thread"
```

sunshowers (approved):

PR #2109, opened 2025-01-30
chore: Upgrade to Winnow 0.7

epage:
With 0.7, `ErrMode` is optional.  It looks like `cut_err` isn't used but I held off on migrating from `ModalResult` to `winnow::Result` and `ModalParser` to `Parser` in case there is a future possibility of using `ErrMode`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4138,9 +4138,9 @@ checksum = "589f6da84c646204747d1270a2a5661ea66ed1cced2631d546fdfb155959f9ec"
 
 [[package]]
 name = "winnow"
-version = "0.6.24"
+version = "0.6.26"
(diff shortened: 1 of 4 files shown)


sunshowers:
Thanks! Yeah, I think preserving the ability to do backtracks and cuts is useful.

(btw a small bit of doc feedback, https://docs.rs/winnow/latest/winnow/error/enum.ErrMode.html could make it a little clearer that "modal parsing" means, specifically, that more than one kind of error can happen during parsing. Had to read https://docs.rs/winnow/latest/winnow/_tutorial/chapter_7/index.html#error-cuts to understand that.)

epage:
Docs PR is up at winnow-rs/winnow#733.  If you could let me know if that does a sufficient job clarifying things, that'd be great!

PR #2145, opened 2025-02-12
Add support for global and group slot numbers

sunshowers:
Per https://github.com/nextest-rs/nextest/discussions/2142 -- add support for global and group slots.

This should address a class of problems around things like port number allocations. For example, if you want to reserve blocks of 5 ports for tests, you could pick an arbitrary starting port in the ephemeral port range (e.g. 46210) and then do `starting_port + 5 * NEXTEST_TEST_GROUP_SLOT`. 

These slots are numbers increasing from 0, are unique for the lifetime of the test within the nextest run (or within the group), and are compact (numbers are freed as tests finish, and the smallest possible number is always assigned to new tests).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1000,10 +1000,11 @@ dependencies = [
 
 [[package]]
 name = "future-queue"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0d1aa68ae1a41111c2ebdfec3c73479b55751b0bdceb8e1e3f7a05f3cf6b2e6"
+checksum = "47cdf4a7eef4808ffa1e5c47dbf37124dfe33a7acc34e8568c5d5359b365a8cb"
 dependencies = [
+ "debug-ignore",
  "fnv",
  "futures-util",
  "pin-project-lite",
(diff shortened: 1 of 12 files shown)


sondrelg:
Thanks again for adding this @sunshowers!

I tried cloning and checking out 9977b2a; building a release version, then running the following test:

```rust
#[test]
fn l() {
    println!("{}", std::env::var("NEXTEST_TEST_GROUP").unwrap());
    println!("{}", std::env::var("NEXTEST_TEST_GROUP_SLOT").unwrap());
}
```

with `./nextest/target/release/cargo-nextest nextest run tests::l`

Unexpectedly, the test fails, since the env var is undefined:

```rust
──── STDERR:             print tests::l
thread 'tests::l' panicked at src/lib.rs:142:60:
called `Result::unwrap()` on an `Err` value: NotPresent

  Cancelling due to test failure
────────────
     Summary [   0.011s] 1 test run: 0 passed, 1 failed, 35 skipped
        FAIL [   0.009s] print tests::l
error: test run failed
```

Am I doing something wrong?

sunshowers:
Thanks for testing it out! This suggests that merely removing the GROUP variables from the environment isn't enough, we should probably set them to values which indicate that the test is not in a group.

PR #2154, opened 2025-02-16
[WIP] option for store dir relative to the target dir

sunshowers:
Still to figure out:

- [ ] Write tests for parsing configuration.
- [ ] Write tests which set configuration and ensure that output gets written to the right spot.
- [ ] How should remapped archive directories be handled? Archive target directories are typically temporary directories which will be deleted at the end of the run. Using that as the store directory seems wrong. In that case, we may wish to use another location instead.


eduardorittner:
Hi! I've been writing some tests for this in https://github.com/eduardorittner/nextest/tree/store-dir-relative. I'm not sure how to best integrate these changes? Open a PR from my branch to upstream/store-dir-relative? You just copy the commits over? 

Also, regarding your third point, do you mean that when store.dir's full path is a temporary directory we should choose another directory? But how can nextest detect when it's a temporary directory? Is there a mechanism for this?

sunshowers:
Thanks for picking this up. Let me rebase on to main and then you can submit your work as dependent on it.

> But how can nextest detect when it's a temporary directory?

Only concerned about temp dirs nextest itself creates in this context, which can be tracked internally.

sunshowers:
Rebased this on main.

PR #2212, opened 2025-03-07
Update documentation about Miri

saethlin:
We forgot about this in https://github.com/nextest-rs/nextest/pull/1347 :facepalm: 

This also mirrors https://github.com/rust-lang/miri/pull/4221, which removes the Miri version of this note.


sunshowers:
Thanks!

PR #2374, opened 2025-06-02
Add support for ": bench" suffix in test list parsing

gaetschwartz:


sunshowers:
Thanks for the pull request. Can this be fixed in iai_callgrind instead? `: bench` is not what Rust's (unstable) benchmark support produces.

gaetschwartz:
Ideally, it can yes ! I'll see with them if it's something they would be inclined to change.

sunshowers:
Thanks. FWIW criterion used to produce ': bench' as well, and I was able to work with them to update that to ': benchmark'.

gaetschwartz:
Since it seems to not be an isolated case, it may be a idea to still allow : bench here, unless you just want to stick as close as possible to the cargo sepcs ?

sunshowers:
> Since it seems to not be an isolated case, it may be a idea to still allow : bench here, unless you just want to stick as close as possible to the cargo sepcs ?

I would definitely like to stay as close to Cargo as possible. I'm fine diverging if there's a compelling reason to do so (e.g. the [process-per-test model](https://nexte.st/docs/design/why-process-per-test/) which is a pretty large divergence), but in this case IMHO I don't think there is.

PR #2426, opened 2025-06-25
[site] update CI action `baptiste0928/cargo-install`: v1 -> v3

jflessau:
v1 and v2 of [baptiste0928/cargo-install](https://github.com/baptiste0928/cargo-install) use a [deprecated cache API](https://github.com/baptiste0928/cargo-install/issues/32#issuecomment-2979676159).


sunshowers:
thanks!

PR #2442, opened 2025-07-05
Add global-timeout option

roblabla:
The global-timeout option gives an upper bound to how long a test run can last. When this limit is reached, the run is cancelled (similar to sending a signal to the process) and the results we got so far are displayed.

This is an implementation of the feature discussed in https://github.com/nextest-rs/nextest/discussions/1898.

### DIFF
--- a/nextest-runner/default-config.toml
+++ b/nextest-runner/default-config.toml
@@ -97,6 +97,11 @@ slow-timeout = { period = "60s" }
 # See <https://nexte.st/docs/features/leaky-tests> for more information.
 leak-timeout = "100ms"
 
+# Stop all tests after the configured global timeout.
+# Defaults to 30 year, which is a large enough value to feel "infinite" without
+# running into overflows on various platforms.
+global-timeout = "30y"
+
 # `nextest archive` automatically includes any build output required by a standard build.
 # However sometimes extra non-standard files are required.
 # To address this, "archive.include" specifies additional paths that will be included in the archive.
--- a/nextest-runner/src/config/global_timeout.rs
+++ b/nextest-runner/src/config/global_timeout.rs
@@ -1,0 +1,120 @@
+// Copyright (c) The nextest Contributors
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use serde::{Deserialize, Deserializer};
+use std::time::Duration;
+
+/// Type for the global-timeout config key.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct GlobalTimeout {
+    pub(crate) period: Duration,
+}
+
+impl<'de> Deserialize<'de> for GlobalTimeout {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Ok(GlobalTimeout {
+            period: humantime_serde::deserialize(deserializer)?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::{
+        NextestConfig,
+        test_helpers::{build_platforms, temp_workspace},
+    };
+    use camino_tempfile::tempdir;
+    use indoc::indoc;
+    use nextest_filtering::ParseContext;
+    use test_case::test_case;
+
+    #[test_case(
+        "",
+        Ok(GlobalTimeout { period: Duration::from_secs(946728000) }),
+        None
+
+        ; "empty config is expected to use the hardcoded values"
+    )]
+    #[test_case(
+        indoc! {r#"
+            [profile.default]
+            global-timeout = "30s"
+        "#},
+        Ok(GlobalTimeout { period: Duration::from_secs(30) }),
+        None
+
+        ; "overrides the default profile"
+    )]
+    #[test_case(
+        indoc! {r#"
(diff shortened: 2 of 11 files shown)


nextest-runner/src/runner/dispatcher.rs
```diff
@@ -285,6 +294,10 @@ where
                             // A test failure has caused cancellation to begin.
                             self.broadcast_request(RunUnitRequest::OtherCancel);
                         }
+                        CancelEvent::Timeout => {
+                            // A test failure has caused cancellation to begin.
```
PaulDance: This comment should probably be updated.

roblabla: Thanks, fixed.

nextest-runner/src/config/global_timeout.rs
```diff
@@ -0,0 +1,165 @@
...
+use serde::{Deserialize, de::IntoDeserializer};
+use std::{fmt, time::Duration};
+
+/// Type for the global-timeout config key.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct GlobalTimeout {
+    #[serde(with = "humantime_serde")]
+    pub(crate) period: Duration,
+}
```
sunshowers: Is there a reason to accept a table with `period` here? That makes more sense for `slow-timeout` which takes a `terminate-after`.

I think `global-timeout` can just accept a scalar `Duration` given that there isn't a great use case for marking an entire test run slow, just one that times out.

nextest-runner/src/runner/dispatcher.rs
```diff
@@ -106,6 +112,9 @@ where
 
         loop {
             let internal_event = tokio::select! {
+                _ = tokio::time::sleep_until(self.timeout_at.into()) => {
```
sunshowers: This should use a [`PausableSleep`](https://github.com/nextest-rs/nextest/blob/7483c7730e2b62e0b50488e3dc3962cb9dbdd9d5/nextest-runner/src/time/pausable_sleep.rs) so that the timeout isn't taking effect while the test run is paused.

nextest-runner/src/runner/dispatcher.rs
```diff
@@ -850,6 +866,7 @@ enum InternalEvent<'a> {
     Signal(SignalEvent),
     Input(InputEvent),
     ReportCancel,
+    Timeout,
```
sunshowers: Let's call this `GlobalTimeout`

nextest-runner/src/runner/dispatcher.rs
```diff
@@ -883,6 +900,7 @@ enum InfoEvent {
 enum CancelEvent {
     Report,
     TestFailure,
+    Timeout,
```
sunshowers: This as well.

nextest-runner/src/runner/dispatcher.rs
```diff
@@ -919,7 +937,9 @@ mod tests {
             vec![],
             0,
             MaxFail::All,
+            Instant::now() + Duration::from_secs(86400 * 365 * 30),
```
sunshowers: Could this reuse `far_future` from `pausable_sleep.rs`?

nextest-runner/default-config.toml
```diff
@@ -97,6 +97,9 @@ slow-timeout = { period = "60s" }
 # See <https://nexte.st/docs/features/leaky-tests> for more information.
 leak-timeout = "100ms"
 
+# Stop all tests after the configured global timeout.
+global-timeout = "30y"
```
sunshowers: Could this explain why it's 30 years? The explanation in `far_future` in `pausable_sleep.rs` can be copied or adapted here.

nextest-runner/src/runner/dispatcher.rs
roblabla: When should it be paused/resumed? In `handle_signal_event`'s `SignalEvent::JobControl(JobControlEvent::Stop)` and `Continue` ?

roblabla: Actually realized I can do it as a reaction to `HandleEventResponse::JobControl` messages in the same function that does the sleep.

nextest-runner/src/runner/dispatcher.rs
roblabla: Since I swapped the implementation to use PausableSleep, I changed it to pass a Duration since that's what PausableSleep needs. This means I can't directly use far_future, since it returns an `Instant`. Not sure how to proceed - I could make a version of PausableStream that takes an Instant, but that feels "wrong" (if you give a deadline, it doesn't make much sense to extend the deadline when you pause).

sunshowers: what about adding a far_future_duration function?

roblabla: Done.

sunshowers:
Looks fantastic, thank you so much! Will try and get this out later today or tomorrow.

sunshowers:
Now out in cargo-nextest 0.9.100.

sunshowers:
(Note I did make a change here, to actively terminate tests -- https://github.com/nextest-rs/nextest/pull/2446. This matches the `timeout` command.)

PR #2459, opened 2025-07-11
chore(deps): update dependencies

reneleonhardt:
### Updated
- Update cargo dependencies
- Update uv dependencies
- Remove unused dependencies
- Clean Cargo.toml (add `edition` in nextest-derive)
- Bump MSRV to 1.86 for cargo_metadata 0.21.0 (except mkdocs.yml)

### Not Updated
- toml 0.9 (breaking changes)


reneleonhardt:
For me, removing license-file in nextest-tests actually passed test_cargo_env_vars, is that possible? 😄

reneleonhardt:
Can you update uv.lock with renovate?

reneleonhardt:
I added a workaround to build-site on macOS.

Maybe you could add a build job to CI if `site/uv.lock` has been changed to make sure it would still work after merging / before releasing.
For example with https://github.com/dorny/paths-filter#example if you don't want to create a new workflow.

sunshowers:
Thank you for the PR. Unfortunately this does far too many things to be accepted in one PR, and is incorrect in some ways (idna_adapter should not be updated).

Is there a higher level goal you'd like to achieve here?

reneleonhardt:
Good catch, I fixed idna_adapter to 1.0.0 again and double-checked, only patch and minor updates except duct.

I saw the open pull requests, the goal was to update cargo dependencies as stated in the description.

The only thing more is uv.lock which had outdated dependencies when I looked after finishing cargo, and the macOS build fix when I tried to verify the new uv.lock .
I can remove both if it's too much for you.

reneleonhardt:
For reference, this would have been the dependabot.yml I created but didn't push, maybe you can configure renovate to update missing pieces so nobody has to do it manually:
```yaml
version: 2
updates:
  - package-ecosystem: cargo
    directories:
      - /
      - cargo-nextest
      - fixture-data
      - fixtures/nextest-metadata-msrv
      - fixtures/nextest-tests
      - fixtures/nextest-tests/cdylib/cdylib-example
      - fixtures/nextest-tests/cdylib/cdylib-link
      - fixtures/nextest-tests/derive
      - fixtures/nextest-tests/dylib-test
      - fixtures/nextest-tests/proc-macro-test
      - fixtures/nextest-tests/with-build-script
      - integration-tests
      - internal-test
      - nextest-filtering
      - nextest-filtering/fuzz
      - nextest-metadata
      - nextest-runner
      - workspace-hack
    schedule:
      interval: weekly
  - package-ecosystem: github-actions
    directory: /
    schedule:
      interval: weekly
  - package-ecosystem: uv
    directory: site
    schedule:
      interval: weekly
```
Of course you can remove directories which should never be updated for test purposes, and you could group updates etc.

reneleonhardt:
What would be the big catch for enabling IDNA support again and allowing updates, will domain parsing containing only ASCII code slow down to a halt?

https://crates.io/crates/idna_adapter#user-content-turning-off-idna-support
> Using this option is not recommended, but to make the idna crate not actually support IDNA,

sunshowers:
Appreciate the PR, but I'm afraid this just adds risk for not much benefit. If you'd like to remove unused dependencies in a separate PR then that would be welcome.

> What would be the big catch for enabling IDNA support again and allowing updates, will domain parsing containing only ASCII code slow down to a halt?

IDNA/punycode support adds unnecessary attack surface.

PR #2462, opened 2025-07-11
[meta] set far-future durations to 1 year instead

sunshowers:
A possible but unlikely cause behind https://github.com/nextest-rs/nextest/discussions/2458 -- set these durations to be 1 year long to be half of Tokio's max timer wheel resolution. Hopefully no one will leave a copy of nextest suspended for a whole year.


sunshowers:
Not required -- was a Cargo 1.88 bug (#2463).

Issue #2463, opened 2025-07-12
nextest 0.9.100 hanging in some configurations


sunshowers:
Nextest 0.9.100 appears to be hanging in some configurations:

* On the Arch Linux build server, as documented in the discussion
* In one illumos CI run so far.

### Discussed in https://github.com/nextest-rs/nextest/discussions/2458

<div type='discussions-op-text'>

<sup>Originally posted by **orhun** July 11, 2025</sup>
Hey!

I'm maintaining [`cargo-nextest`](https://archlinux.org/packages/extra/x86_64/cargo-nextest/) for Arch Linux and we are currently running tests like this after building the package:

```shell
build() {
  cd "$pkgname-$pkgver"
  cargo build --release --frozen --package "$pkgname" --no-default-features --features default-no-update
}

check() {
  cd "$pkgname-$pkgver"
  cargo run --package cargo-nextest -- nextest run -- --skip test_version_info
}
```

I adopted this from the GitHub workflows from this repository a while ago.

However, with the last release (`0.9.100`), running the tests take a long time and get stuck after a while, maybe due to the new global timeout feature?

More specifically, after this point:

```
        SLOW [>1620.000s] nextest-runner::integration basic::test_retries::retry_overrides_ignored
        SLOW [>1620.000s] nextest-runner::integration basic::test_retries::retry_overrides_obeyed
        SLOW [>1620.000s] nextest-runner::integration basic::test_run
        SLOW [>1620.000s] nextest-runner::integration basic::test_run_ignored
        SLOW [>1620.000s] nextest-runner::integration basic::test_string_filters_without_filter_expr
        SLOW [>1620.000s] nextest-runner::integration basic::test_termination
        SLOW [>1620.000s] nextest-runner::integration target_runner::test_listing_with_target_runner
        SLOW [>1620.000s] nextest-runner::integration target_runner::test_run_with_target_runn
```

So I have a couple of questions:

- What do you suggest we do as packagers to avoid long-running tests like this? Is there a filter command or something similar to skip these?
- Is my way of running tests correct & can it be improved?

Thank you :)</div>

sunshowers:
Some further investigation:

I haven't been able to reproduce this despite running the nextest test suite in a loop on both Linux and illumos for the last 2-3 hours. This hasn't reproduced in illumos CI either. I've added a per-test timeout in CI so that hopefully next time it happens there are more logs.

[...]

PR #2470, opened 2025-07-12
[nextest-runner] add --verbose --verbose to cargo command execution + run tests

sunshowers:
See #2463 -- this might be a hang in Cargo similar to
rust-lang/cargo#15032. Try running cargo with
`--verbose --verbose`.


sunshowers:
Not required -- was a Cargo 1.88 bug (https://github.com/nextest-rs/nextest/issues/2463).

PR #2484, opened 2025-07-22
[meta] update cc to 1.2.30

sunshowers:
Includes https://github.com/rust-lang/cc-rs/pull/1496 for illumos.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -449,9 +449,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.25"
+version = "1.2.30"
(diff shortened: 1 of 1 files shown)


sunshowers:
Chatted with iximeow (the author of the cc fix), and it looks like this is likely not urgent to get out as a fix.

Within nextest, cc is used as a dependency for:

* openssl
* ring
* zstd-sys

The first two are only used in self-updates, while the last one is used for archives (and the C library works with in-memory buffers so shouldn't poke at errno). We should still land this, though, so it gets pulled into the next nextest release.

PR #2514, opened 2025-08-24
Make `--reporter libtest-json-plus` output test results immediately

### LINKED ISSUE #1256: JSON output is not emitted until tests end
With `cargo test -- -Z unstable-options --format json`, I seem to receive JSON lines as soon as a corresponding event is emitted. With `cargo nextest run --message-format libtest-json --color never --failure-output immediate --success-output immediate`, or pretty much any other combination I try, JSON is output only after all tests are complete, before the final summary. Is my observation correct? IS there a way to receive any form of machine readable output as tests progress? In my use case, I am not interested with the console feedback or human readable messages at all, I need only any form of machine readable output, but I need it to be available around the time of a corresponding test event, and not at the very end. Is it possible to satisfy my use case with currently available state of `nextest`? I figured out that one reason for such behavior could be that otherwise it would not be possible to avoid interleaved messages of different formats, but maybe it would be worth to have an option to suppress the original, human readable messages, and have immediate JSON lines instead?

dnbln:
Given that it is possible to distinguish between different suites in the structured reporter through the `nextest` property, this patch outputs test results as soon as they happen, when using `--reporter libtest-json-plus`.

`--reporter libtest-json` output is unaffected.

Maybe fixes #1256 ?

### DIFF
--- a/nextest-runner/src/reporter/error_description.rs
+++ b/nextest-runner/src/reporter/error_description.rs
@@ -340,7 +340,7 @@ fn heuristic_error_str(stderr: &[u8]) -> Option<ByteSubslice<'_>> {
 
 // This regex works for the default panic handler for Rust -- other panic handlers may not work,
 // which is why this is heuristic.
-static PANICKED_AT_REGEX_STR: &str = "^thread '([^']+)' panicked at ";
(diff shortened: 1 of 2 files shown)


sunshowers:
Looks good, thank you!

sunshowers:
This is now out in `cargo-nextest 0.9.103`. Thanks!

PR #2577, opened 2025-09-14
Always forward --config args to cargo invocations

### LINKED ISSUE #2566: Bug: `--config` arguments not forwarded to `cargo metadata`
### Description of the issue

**Description**

When invoking nextest with `--config` arguments, those aren't forwarded when invoking `cargo metadata`.

In my case, I pass `--config .cargo/patched-cargo.toml` where `patched-cargo.toml` contains a `[patch.crates-io]` section. Invoking `cargo metadata` regenerates `Cargo.lock`, referencing unpatched dependencies, and leading to build failures.

### Expected outcome

I expected the build (`cargo test --no-run`) to succeed and the tests to be run.

### Actual result

The build (`cargo test --no-run`) fails because incorrect/unpatched dependencies are compiled against.

The build failure is preceded by the following unexpected and unwanted `Cargo.lock` regeneration.

```
    Updating crates.io index
     Locking 23 packages to latest compatible versions
      Adding backoff v0.4.0
            ⋮
      Adding windows_x86_64_msvc v0.42.2
warning: Patch `tokio v1.999.0 (/home/<redacted>/_patched/tokio)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

### Nextest version

```text
cargo-nextest 0.9.103 (d2e7b879f 2025-08-25)
release: 0.9.103
commit-hash: d2e7b879fb79975e8b47a8e3ce569b651e6381c0
commit-date: 2025-08-25
host: x86_64-unknown-linux-gnu
```

### Additional context

**Possible solution**

As far as I can tell, this can be solved by adding the config options in `CargoCli::add_generic_cargo_options` rather than `CargoCli::add_options` (see below). The test suite passes with that change, but it may have unintended effects.

```patch
diff --git a/cargo-nextest/src/cargo_cli.rs b/cargo-nextest/src/cargo_cli.rs
index 41854fb1..e89a6013 100644
--- a/cargo-nextest/src/cargo_cli.rs
+++ b/cargo-nextest/src/cargo_cli.rs
@@ -356,7 +356,6 @@ impl<'a> CargoCli<'a> {
         if options.future_incompat_report {
             self.add_arg("--future-incompat-report");
         }
-        self.add_args(options.config.iter().flat_map(|s| ["--config", s.as_str()]));
         self.add_args(
             options
                 .unstable_flags
@@ -368,7 +367,7 @@ impl<'a> CargoCli<'a> {
     }

     /// Add Cargo options that are common to all commands.
-    pub(crate) fn add_generic_cargo_options(&mut self, options: &CargoOptions) -> &mut Self {
+    pub(crate) fn add_generic_cargo_options(&mut self, options: &'a CargoOptions) -> &mut Self {
         // ---
         // Manifest options
         // ---
@@ -391,6 +390,8 @@ impl<'a> CargoCli<'a> {
             self.stderr_null = true;
         }

+        self.add_args(options.config.iter().flat_map(|s| ["--config", s.as_str()]));
+
         self
     }
```

benschulz:
Resolves #2566.

### DIFF
--- a/cargo-nextest/src/cargo_cli.rs
+++ b/cargo-nextest/src/cargo_cli.rs
@@ -356,7 +356,6 @@ impl<'a> CargoCli<'a> {
         if options.future_incompat_report {
             self.add_arg("--future-incompat-report");
         }
-        self.add_args(options.config.iter().flat_map(|s| ["--config", s.as_str()]));
         self.add_args(
             options
                 .unstable_flags
@@ -368,7 +367,7 @@ impl<'a> CargoCli<'a> {
     }
 
     /// Add Cargo options that are common to all commands.
-    pub(crate) fn add_generic_cargo_options(&mut self, options: &CargoOptions) -> &mut Self {
+    pub(crate) fn add_generic_cargo_options(&mut self, options: &'a CargoOptions) -> &mut Self {
         // ---
         // Manifest options
         // ---
@@ -390,6 +389,7 @@ impl<'a> CargoCli<'a> {
         if options.cargo_quiet > 1 {
             self.stderr_null = true;
         }
+        self.add_args(options.config.iter().flat_map(|s| ["--config", s.as_str()]));
 
         self
     }


sunshowers (approved):

sunshowers:
Thank you!

PR #2595, opened 2025-09-16
feat: cargo profiles inheritance

zokhcat:


zokhcat:
@sunshowers could you please take a look at this and give a review when you have a chance?

nextest-runner/src/config/core/imp.rs
```diff
@@ -874,6 +884,7 @@ pub struct EarlyProfile<'cfg> {
     store_dir: Utf8PathBuf,
     default_profile: &'cfg DefaultProfileImpl,
     custom_profile: Option<&'cfg CustomProfileImpl>,
+    inheritance_chain: Vec<&'cfg CustomProfileImpl>,
```
sunshowers, line 887: nice!

nextest-runner/src/config/core/imp.rs
```diff
@@ -984,11 +997,24 @@ impl<'cfg> EvaluatableProfile<'cfg> {
...
+        
+        // Walk up inheritance chain
+        for parent in &self.inheritance_chain {
+            if let Some(retries) = parent.retries {
+                return retries;
+            }
+        }
+        
+        // Fall back to default
+        self.default_profile.retries
```
sunshowers, line 1017: This looks good -- you'll want to replicate this logic for all of the other config options here (can you extract this into a common function?)

nextest-runner/src/config/core/imp.rs
```diff
@@ -1124,6 +1150,71 @@ pub(in crate::config) struct NextestConfigImpl {
...
+        chain: &mut Vec<&'cfg CustomProfileImpl>,
+    ) -> Result<(), ProfileNotFound> {
+        if visited.contains(profile_name) {
+            return Err(ProfileNotFound::new(
+                profile_name,
+                self.all_profiles().collect::<Vec<_>>(),
+            ));
+        }
+        
+        visited.insert(profile_name.to_string());
```
sunshowers, line 1175: Do you need the `to_string` here? Or can you store references to the profile name?

You might run into [variance](https://lifetime-variance.sunshowers.io/) issues with storing a reference, but I think you can have another lifetime parameter for that.

nextest-runner/src/config/core/imp.rs
```diff
@@ -1124,6 +1150,71 @@ pub(in crate::config) struct NextestConfigImpl {
...
+                self.resolve_profile_chain_recursive(parent_name, visited, chain)?;
+            }
+            chain.push(profile);
+        }
+    
+        Ok(())
+    }
+
+    fn check_inheritance_cycles(&self) -> Result<(), ConfigParseError> {
+        let mut graph = Graph::<String, (), Directed>::new();
```
sunshowers, line 1189: same here, can you store references?

nextest-runner/src/config/core/imp.rs
```diff
@@ -1124,6 +1150,71 @@ pub(in crate::config) struct NextestConfigImpl {
...
+                }
+            }
+        }
+
+        match toposort(&graph, None) {
+            Ok(_) => Ok(()),
+            Err(cycle) => {
+                let cycle_profile = graph[cycle.node_id()].clone();
+                Err(ConfigParseError::new(
+                    "Inheritance cycle detected in profile configuration",
```
sunshowers, line 1210: We use lowercase for errors:

```suggestion
                    "inheritance cycle detected in profile configuration",
```

nextest-runner/src/config/core/imp.rs
```diff
@@ -1124,6 +1150,71 @@ pub(in crate::config) struct NextestConfigImpl {
...
+                    None,
+                    ConfigParseErrorKind::InheritanceCycle(cycle_profile),
```
sunshowers, line 1212: Could we use https://docs.rs/petgraph/latest/petgraph/algo/fn.kosaraju_scc.html to detect all cycles here? Each SCC returned by the algorithm is a cycle -- it would be nice to print out all such cycles here.

nextest-runner/src/config/core/imp.rs
```diff
@@ -1337,6 +1428,8 @@ pub(in crate::config) struct CustomProfileImpl {
     junit: JunitImpl,
     #[serde(default)]
     archive: Option<ArchiveConfig>,
+    #[serde(default)]
+    inherit: Option<String>
```
sunshowers, line 1432: Need a trailing comma (did rustfmt not catch this?)

nextest-runner/src/config/core/imp.rs
```diff
@@ -1124,6 +1150,71 @@ pub(in crate::config) struct NextestConfigImpl {
...
+            if let Some(parent_name) = &profile.inherit {
+                self.resolve_profile_chain_recursive(parent_name, visited, chain)?;
+            }
+            chain.push(profile);
+        }
+    
+        Ok(())
+    }
+
+    fn check_inheritance_cycles(&self) -> Result<(), ConfigParseError> {
```
sunshowers, line 1188: We need some tests, both for inheritance and for cycles.

sunshowers (commented):
Thanks for working on this!

asder8215:
I just noticed there was already a PR created to resolve the inherits issue. Is it alright if I continue with this @zokhcat @sunshowers?

sunshowers:
Sure, I'm fine with that.

nextest-runner/src/config/core/imp.rs
```diff
@@ -792,7 +793,15 @@ impl NextestConfig {
...
     fn make_profile(&self, name: &str) -> Result<EarlyProfile<'_>, ProfileNotFound> {
+        // Check for cycles first
+        self.inner.check_inheritance_cycles().unwrap();
+        
         let custom_profile = self.inner.get_profile(name)?;
+        let inheritance_chain = if let Some(_) = custom_profile {
+            self.inner.resolve_profile_chain(name)?
+        } else {
+            Vec::new()
+        };
```
asder8215, line 804: Hi @sunshowers, I was taking a look at this part of the code and was wondering if this needed to be resolved as a `Vec<&'cfg CustomProfileImpl>`? Earlier we checked for `self.inner.check_inheritance_cycles()`, which if it passes through that, that would indicate that there are no cycles detected and we can retrieve the root ancestor in the chain, right?

Would it be appropriate to store this as an `Option<&'cfg CustomProfileImpl>`?

asder8215: Oh wait nevermind. I'm assuming within `EvaluatableProfile` it is possible to inherit certain configs that are not seen at the end of the chain of `CustomProfileImpl`?

Like, if I had a profile called "foo" that inherits from a profile called "bar" which inherits from a profile called "baz".

The "bar" profile could have a filled in "slow_timeout" field while "baz" doesn't have a "slow_timeout" field. "foo" would still inherit the "slow_timeout" field from the "baz" profile, is what I'm understanding?

nextest-runner/src/config/core/imp.rs
asder8215: I noticed in the functions they all have the same or similar format of:
```
self.custom_profile
            .and_then(|profile| profile.${some_field})
            .unwrap_or(&self.default_profile.${some_field})
```
On top of that we'd have to include inheritance chain logic to go up the chain for any parent `CustomProfileImpl` that possibly includes that field.

Would it be appropriate to encapsulate all of this into a declarative macro?

sunshowers:
This work has since landed. Thanks for starting it and getting it over the line, both of you!

Issue #526, opened 2022-09-13
[feature] progress in testing log


jiangying000:
currently nextest supports progress bar, but I did not find a way to enable a log like below, it is an example made by hand:

    Finished test [unoptimized + debuginfo] target(s) in 6m 01s
    Starting 500 tests across 161 binaries (17 skipped)
    [1/500]    PASS [   0.009s]                                        bcs-ext tests::test_bcs_enum_compact
    [2/500]     PASS [   0.009s]                                        bcs-ext tests::test_serialize
      ....many lines
    [400/500]    PASS [   0.232s]                starcoin-service-registry::basc test_launch_and_shutdown
     [401/500]     PASS [   0.234s]                starcoin-service-registry::basc test_basic

so that when a open a part of CI log ( like the one on github, github give us tail of log when testing is running) we can see current testing progress

sunshowers:
I agree, this would be valuable. Let me see what I can do here.

sunshowers:
I'm going to call this a progress counter, as distinct from the progress bar.

## When should the counter be displayed?

I think it would make sense to display it whenever the progress bar isn't displayed. Specifically, we'd want to add a CLI flag called `--show-progress`, with values `bar`, `counter` or `none`. If `--show-progress` is not specified, then:

1. If this is a CI environment, set it to `counter`.
2. If standard error is a terminal, set it to `bar`.
3. Otherwise, set it to `counter`.

If `--show-progress bar` is specified and standard error is not a terminal, then this implies `none`.

We'll also want to deprecate `--hide-progress-bar`, make it mean `--show-progress none`, and make it conflict with `--show-progress` so both of those can't be specified at the same time.

## Where should the progress counter be shown?

I think the best place to show the progress would be to the right of the timer: e.g.

```
        PASS [   0.232s 400/500]                starcoin-service-registry::basc test_launch_and_shutdown
```

One thing to be careful of here is padding. This is what it should look like:

```
        PASS [   0.009s  99/500]                                        bcs-ext tests::test_serialize
        PASS [   0.232s 100/500]                starcoin-service-registry::basc test_launch_and_shutdown
```

## What should be displayed?

I think the progress counter should be kept really simple for now. Only the number of completed tests and the total number of tests to run should be displayed. We could potentially show failed tests here but that'll be too noisy I think.

Would you like to implement this, @jiangying000? I think it should maybe be 1.5 days of work or so.

jiangying000:
[...]

PR #2618, opened 2025-09-29
Add a counter to the status line with each test

bobrik:
This adds a counter to each status line, so that one doesn't have to guess how many tests are remaining when the progress bar is not visible (think in CI):

image

It's been a long requested feature: #526. As a casual observer of test runs with 5000+ tests, this would be nice to have.

As is, it shows the counter unconditionally. In #526 @sunshowers suggested a more complicated heuristic (see https://github.com/nextest-rs/nextest/issues/526#issuecomment-1254332115), which I can probably take a stab at next, if it still makes sense.

### DIFF
--- a/nextest-runner/src/reporter/displayer/imp.rs
+++ b/nextest-runner/src/reporter/displayer/imp.rs
@@ -19,10 +19,12 @@ use crate::{
     cargo_config::CargoConfigs,
     config::{elements::LeakTimeoutResult, overrides::CompiledDefaultFilter, scripts::ScriptId},
     errors::WriteEventError,
-    helpers::{DisplayScriptInstance, DisplayTestInstance, plural},
+    helpers::{
+        DisplayCounterIndex, DisplayScriptInstance, DisplayTestInstance, decimal_char_width, plural,
+    },
     list::{TestInstance, TestInstanceId},
     reporter::{
-        displayer::{formatters::DisplayHhMmSs, progress::TerminalProgress},
+        displayer::{ShowProgress, formatters::DisplayHhMmSs, progress::TerminalProgress},
         events::*,
         helpers::Styles,
         imp::ReporterStderr,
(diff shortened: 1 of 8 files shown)


sunshowers:
Thanks. I think the more complex heuristic described in https://github.com/nextest-rs/nextest/issues/526#issuecomment-1254332115 is required here. I hope it's not too much of a burden.

Looking at the output, I'd also put the counter after the time taken.

sunshowers:
Thanks, hoping to look at this soon (currently dealing with a hard deadline at my workplace)

nextest-runner/src/reporter/displayer/imp.rs
```diff
@@ -1224,11 +1261,19 @@ impl<'a> DisplayReporterImpl<'a> {
             }
         };
 
-        // Print the time taken and the name of the test.
+        write!(
```
sunshowers, line 1271: A couple of comments here:

1. This looks too similar to stress test counters (try out `--stress-count 5` or similar). I think we'll somehow need to make it clear that this isn't the stress test counter. Using parentheses might work:

```
        PASS [   1.860s] [1/10] ( 40/300) integration-tests::integration test_relocated_run
```
1. I think the duration should stay where it is, and that the counter should be displayed after that as well as after the stress test counter.

nextest-runner/src/reporter/displayer/formatters.rs
```diff
@@ -62,6 +62,20 @@ impl fmt::Display for DisplayHhMmSs {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "[{:>width$} / {:>width$}] ",
+            self.0.finished_count,
+            self.0.initial_run_count,
+            width = (self.0.initial_run_count.ilog10() as usize + 1).max(1),
+        )
+    }
+}
```
sunshowers: No need for the extra spaces around the `/`.

nextest-runner/src/reporter/displayer/imp.rs
bobrik: Sure thing, updated.

Issue #2620, opened 2025-10-01
Bug: On Windows tests can not breakaway from the job-object created by nextest


Guiguiprim:
### Description of the issue

I'm unable to test some code that use Windows API `AssignProcessToJobObject()` on Windows 7.

- On Windows 7, processes can only have one job-object
- `CREATE_BREAKAWAY_FROM_JOB` should be used during process creation
-  `JOB_OBJECT_LIMIT_BREAKAWAY_OK` is required to do so

`nextest` is using job-objects but without `JOB_OBJECT_LIMIT_BREAKAWAY_OK`

**Steps to reproduce:**
1.  call `CreateProcess()` with the `CREATE_BREAKAWAY_FROM_JOB`  flag

### Expected outcome

The tests using the `AssignProcessToJobObject()` API pass.

### Actual result

The tests fail with `ACCESS_DENIED`.

### Nextest version

```text
cargo-nextest 0.9.100-rc.2
```

### Additional context

The patch should be simple enough:

Replacing

```rust
let job = Job::create().ok();
```

by

```rust
let job = Job::create_with_limit_info(ExtendedLimitInfo::new().limit_breakaway_ok()).ok();
```

I'm happy to do it, if you do not see any reason not to.

PR #2621, opened 2025-10-01
[nextest-runner] Add "breakaway OK" to Windows job object limits

Guiguiprim:
This is the minimalist fix for https://github.com/nextest-rs/nextest/issues/2620.

With this changes my use-case works.

I only added `JOB_OBJECT_LIMIT_BREAKAWAY_OK`. `JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK` is closely related, I didn't need it but was wondering if I should also add it in good measure.

### DIFF
--- a/nextest-runner/src/runner/executor.rs
+++ b/nextest-runner/src/runner/executor.rs
@@ -369,7 +369,7 @@ impl<'a> ExecutorContext<'a> {
 
         // If creating a job fails, we might be on an old system. Ignore this -- job objects are a
         // best-effort thing.
-        let job = super::os::Job::create().ok();
+        let job = super::os::create_job().ok();
 
         // The --no-capture CLI argument overrides the config.
         if self.capture_strategy != CaptureStrategy::None {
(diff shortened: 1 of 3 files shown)


sunshowers:
Thanks -- so part of the point of job objects is that when nextest goes to terminate parent processes, child processes also get terminated. How does this change impact that?

Guiguiprim:
This change allow using `CREATE_BREAKAWAY_FROM_JOB`  in `CreateProcess()` calls. 

- Before: this would result in a `ACCESS_DENIED` and such code would be impossible to test using `nextest`
- After: the process creation can succeed and the created process will indeed no be part of the job-object created by `nextest` losing all the associated behaviour. But I believe this is fine, if a crate is using this flag, it is trying to do advanced process management and I would expect it to correctly handle process termination (by for example creating itself a job-object with similar attributes or any other mechanism). Starting from Windows 8, processes can be part of a job-object hierarchy making the flag unnecessary in most case.

sunshowers:
Gotcha, thanks! That seems reasonable then, it's similar to a test creating  a Unix process group for its own child processes.

sunshowers:
Released in cargo-nextest 0.9.105, should be available within 20 minutes.

PR #2622, opened 2025-10-01
Make it easier to use nextest as a lib (as opposed to through CLI)

TobiasvdVen:
See feature request for details: https://github.com/nextest-rs/nextest/discussions/2619

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -781,7 +781,8 @@ impl From<RunIgnoredOpt> for RunIgnored {
 }
 
 impl CargoOptions {
-    fn compute_binary_list(
+    /// Invoke 'cargo test --no-run' to compile test binaries and produce a list of them
+    pub fn compute_binary_list(
         &self,
         graph: &PackageGraph,
         manifest_path: Option<&Utf8Path>,
@@ -797,6 +798,7 @@ impl CargoOptions {
         cargo_cli.add_options(self);
 
         let expression = cargo_cli.to_expression();
+
         let output = expression
             .stdout_capture()
             .unchecked()
@@ -2405,7 +2407,8 @@ pub enum BuildPlatformsOutputFormat {
     Triple,
 }
 
-fn acquire_graph_data(
+/// Invoke 'cargo metadata', the result may be parsed with `PackageGraph::from_json`
+pub fn acquire_graph_data(
     manifest_path: Option<&Utf8Path>,
     target_dir: Option<&Utf8Path>,
     cargo_opts: &CargoOptions,
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -798,7 +798,6 @@ impl CargoOptions {
         cargo_cli.add_options(self);
 
         let expression = cargo_cli.to_expression();
-
         let output = expression
             .stdout_capture()
             .unchecked()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -300,31 +300,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bon"
-version = "3.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2529c31017402be841eb45892278a6c21a000c0a17643af326c73a73f83f0fb"
-dependencies = [
- "bon-macros",
- "rustversion",
-]
-
-[[package]]
-name = "bon-macros"
-version = "3.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d82020dadcb845a345591863adb65d74fa8dc5c18a0b6d408470e13b7adc7005"
-dependencies = [
- "darling",
- "ident_case",
- "prettyplease",
- "proc-macro2",
- "quote",
- "rustversion",
- "syn",
-]
-
-[[package]]
 name = "borsh"
 version = "1.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -395,7 +370,6 @@ dependencies = [
 name = "cargo-nextest"
 version = "0.9.104"
 dependencies = [
- "bon",
  "camino",
  "camino-tempfile",
  "cfg-if",
@@ -825,41 +799,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
 dependencies = [
  "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
-name = "darling"
-version = "0.21.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9cdf337090841a411e2a7f3deb9187445851f91b309c0c0a29e05f74a00a48c0"
-dependencies = [
- "darling_core",
- "darling_macro",
-]
-
-[[package]]
-name = "darling_core"
-version = "0.21.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1247195ecd7e3c85f83c8d2a366e4210d588e802133e1e355180a9870b517ea4"
-dependencies = [
- "fnv",
- "ident_case",
- "proc-macro2",
- "quote",
- "strsim",
- "syn",
-]
-
-[[package]]
-name = "darling_macro"
-version = "0.21.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d38308df82d1080de0afee5d069fa14b0326a88c14f15c5ccda35b4a6c414c81"
-dependencies = [
- "darling_core",
  "quote",
  "syn",
 ]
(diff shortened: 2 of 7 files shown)


TobiasvdVen:
I see the `CargoOptions::new()` function is causing a drop in coverage, do you have a recommended approach here? I could add a test of some form, but it would be rather trivial for what is merely a constructor.

Not sure why the `CI/ Lint (ubuntu-latest) ` step is failing, it seems to also be the case on other PRs, so perhaps not related to the changes?

sunshowers:
Ahh, unfortunately this isn't a correct approach, because the public API for cargo-nextest is defined as just the CLI. We do patch releases for nextest version upgrades based on that, and making functions from cargo-nextest public would necessitate a breaking version upgrade each time the APIs changed.

I would be fine moving bits into nextest-runner as needed.

The big new function is also not a good idea. I don't know why that needs to be a builder -- can the fields just be `pub`?

TobiasvdVen:
Ah, I see, I'll move these required pieces out of `cargo-nextest` and into `nextest-runner` instead. 

Regarding the `new()` function, I will remove it and make the fields `pub`, as suggested. My personal intuition was that a `new()` function would be a less invasive change. And I was too lazy to write a whole builder pattern without relying on a proc-macro :)

TobiasvdVen:
Let me know if this update matches your expectations. A few notes:

I started by trying to pull `OutputContext` to `nextest-runner`, but after that lead to a few more things needing to move, it seemed the better route might be to remove the dependency on `OutputContext` in `CargoOptions::compute_binary_list` and `acquire_graph_data`. I did this under the presumption that the `OutputContext` was only used to set the `--color` argument, and these functions seemed to capture the stdout of the command, thus it would be irrelevant. It's possible I missed something and that the `OutputContext` does need to be included.

The next largest change was the need to introduce new error types for `compute_binary_list` and `acquire_graph_data` to return: `CreateBinaryListError` (modeled after `CreateTestListError`) and `CargoMetadataError`, respectively. The latter includes a `TargetTripleError` as a kind, which is why the error handling in `cargo-nextest/src/errors.rs` now has a function `display_target_triple_error_to_stderr`. Maybe there's a better way to avoid duplicating the `TargetTripleError` handling that I didn't see.

It can be hard to tell in the diff of moved code, so I'll add some comments where the `OutputContext` was removed.

cargo-nextest/src/dispatch.rs
```diff
@@ -1588,7 +1552,6 @@ impl BaseApp {
             None => Arc::new(self.cargo_opts.compute_binary_list(
                 self.graph(),
                 self.manifest_path.as_deref(),
-                self.output,
```
TobiasvdVen, line 1591: Removal of `OutputContext`, was passed into `compute_binary_list`.

cargo-nextest/src/dispatch.rs
```diff
@@ -2000,7 +1963,7 @@ impl ShowConfigCommand {
         match self {
             Self::Version {} => {
                 let mut cargo_cli =
-                    CargoCli::new("locate-project", manifest_path.as_deref(), output);
+                    CargoCli::new("locate-project", manifest_path.as_deref());
```
TobiasvdVen: Removal of `OutputContext`, was passed into `CargoCli::new`, where its only usage was in `CargoCli::to_expression` to add the `--color` argument.

cargo-nextest/src/errors.rs
```diff
@@ -741,14 +696,7 @@ impl ExpectedError {
...
             Self::TargetTripleError { err } => {
-                if let Some(report) = err.source_report() {
-                    // Display the miette report if available.
-                    error!(target: "cargo_nextest::no_heading", "{report:?}");
-                    None
-                } else {
-                    error!("{err}");
-                    err.source()
-                }
+                display_target_triple_error_to_stderr(err)
```
TobiasvdVen, line 699: The original printing of `TargetTripleError` to stderr.

cargo-nextest/src/errors.rs
```diff
@@ -485,20 +433,27 @@ impl ExpectedError {
...
+                    } => {
+                        error!(
+                            "command `{}` failed with {}",
+                            command.style(styles.bold),
+                            exit_status
+                        );
+                        None
+                    }
+                    CargoMetadataError::TargetTriple { err } => {
+                        display_target_triple_error_to_stderr(err)
```
TobiasvdVen, line 454: The new case of printing of `TargetTripleError` to stderr.

nextest-runner/src/cargo_cli.rs
```diff
@@ -198,47 +202,86 @@ pub(crate) struct CargoOptions {
         group = "cargo-opts",
         help_heading = "Other Cargo options"
     )]
-    unstable_flags: Vec<String>,
+    pub unstable_flags: Vec<String>,
 }
 
+impl CargoOptions {
+    /// Invoke 'cargo test --no-run' to compile test binaries and produce a list of them
+    pub fn compute_binary_list(
```
TobiasvdVen, line 210: `OutputContext` was removed as a parameter here, it was previously passed on into `CargoCli::new`.

nextest-runner/src/cargo_cli.rs
```diff
@@ -426,6 +471,50 @@ impl<'a> CargoCli<'a> {
     }
 }
 
+/// Invoke 'cargo metadata', the result may be parsed with `PackageGraph::from_json`
+pub fn acquire_graph_data(
```
TobiasvdVen, line 475: `OutputContext` was removed as a parameter here, it was previously passed on into `CargoCli::new`.

sunshowers:
Thanks, hoping to look at this soon (currently dealing with a hard deadline at my workplace)

cargo-nextest/src/dispatch.rs
sunshowers: Where is the `--color` argument passed in now?

cargo-nextest/src/errors.rs
sunshowers: What change was made here? Not quite clear about it.

nextest-runner/src/errors.rs
```diff
@@ -954,6 +954,130 @@ pub enum FromMessagesError {
...
+        exit_code: Option<i32>,
+    ) -> Self {
+        Self::CommandFail {
+            command: shell_words::join(command),
+            exit_code,
+        }
+    }
+}
+
+/// An error that occurs while gather cargo metadata.
```
sunshowers:
```suggestion
/// An error that occurs while gathering cargo metadata.
```

cargo-nextest/src/dispatch.rs
TobiasvdVen: Well, it isn't 😄 

The way I went about this is noticing that the `--color` argument seemed unnecessary, as all usages of `to_expression` (the only place `--color` was added) are in places that don't display their output - it is captured and only used internally (specifically in `compute_binary_list` and `acquire_graph_data`).

There is this comment in `CargoCli`, which may or may not be relevant: `// --color is handled by runner`. 

I'm not entirely sure what it's intended to communicate, but I think it's explaining why there is no `color` field on the `CargoCli` struct. Maybe if we're no longer "internally" adding the `--color` argument it should become a field of `CargoCli`, so that users can set it? I'm not sure if `CargoCli` is intended for use beyond what it is currently used for, though, in which case it could be superfluous.

My concern would be that I missed some way in which it was relevant. Presumably it was there for a reason, at least at one point, but I was unable to identify one.

cargo-nextest/src/errors.rs
TobiasvdVen: Right, so there was pre-existent handling of the `TargetTriple` error kind in this big error-handling `match`. 

One of the new error types (`CargoMetadataError`) I introduced also has a `TargetTriple` error kind as one of its sub-kinds. So now there were 2 places in the `match` that needed to handle the same error kind (`TargetTriple`). Hence the introduction of a function to avoid duplicating. 

Hopefully that clarifies, let me know if not.

sunshowers: Ah gotcha!

cargo-nextest/src/dispatch.rs
sunshowers: Well -- when the user runs `cargo nextest run`, we run `cargo test --no-run` to build all the tests -- we want to forward the `--color` argument to that underlying `cargo test` execution. Totally fine with it being factored out to be easier to do, but we need to maintain this property.

TobiasvdVen: Ah, I went and had another look, and I misinterpreted things before.

I thought the output of `cargo test --no-run` (in `compute_binary_list`) was being captured, in the sense that it was only used for parsing to a list of binaries. But its output is also sent to stdout, so `--color` is relevant. 

My bad, I'll have to take a different approach.

sunshowers: Thanks -- stdout (machine-readable output) is captured, but stderr (human-readable output) is passed through.

sunshowers:
Looks like a bad merge happened -- feel free to rebase locally, or delete + recreate if that's more convenient.

TobiasvdVen:
> Looks like a bad merge happened -- feel free to rebase locally, or delete + recreate if that's more convenient.

Whoops, sorry for the intrusion. I'll just close this for now.

PR #2670, opened 2025-10-18
Introduce test filter flag to nextest archive command

clundin55:
Filtersets can be used to reduce the amount of binaries archived in a test archive.

NOTE: Test binaries are not executed to collect a test list. This prevents filtering by test.

See feature request: https://github.com/nextest-rs/nextest/discussions/2648

### DIFF
--- a/cargo-nextest/src/dispatch.rs
+++ b/cargo-nextest/src/dispatch.rs
@@ -1618,10 +1618,7 @@ impl App {
     fn new(base: BaseApp, build_filter: TestBuildFilter) -> Result<Self> {
         check_experimental_filtering(base.output);
 
-        Ok(Self {
-            base,
-            build_filter,
-        })
+        Ok(Self { base, build_filter })
     }
 
     fn build_test_list(


integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +842,29 @@ archive.include = [
         .expect_err("archive should have failed");
 }
 
+#[test]
+fn test_archive_with_build_filters() {
+    set_env_vars();
+
+    let filters = ["all()", "none()", "package(cdylib-example)"];
+    let expected_file_counts = [27, 11, 12];
```
clundin55: This is brittle and it seems the file count varies per platform. I'll rewrite this once I have feedback on the implementation.

cargo-nextest/src/dispatch.rs
```diff
@@ -1797,6 +1702,122 @@ impl App {
...
+        let test_filter_builder = self.build_filter.make_test_filter_builder(filter_exprs)?;
+        let ecx = profile.filterset_ecx();
+
+        // Apply filterset to `RustTestArtifact` list.
+        let test_artifacts: BTreeSet<_> = test_artifacts
+            .iter()
+            .filter(|test_artifact| {
+                matches!(
+                    test_filter_builder.filter_binary_match(test_artifact, &ecx, filter_bound),
+                    FilterBinaryMatch::Definite | FilterBinaryMatch::Possible
```
sunshowers: Rather than this I'd just reject `test` predicates entirely, returning a `FiltersetParseErrors` that contains all the errors (so you get a nice display). You could do it either at parsing time or as a pass afterwards (have a `check_no_test_predicates` method or similar).

integration-tests/tests/integration/main.rs
sunshowers: This does vary by platform but doesn't change much. Instead can you scan the archive to ensure that the expected binaries are available and the others aren't?

clundin55: Modified the test to check which files are present in the archive.

cargo-nextest/src/dispatch.rs
clundin55: Makes sense! I've modified the parser to introduce an error when the filterset kind is `FiltersetKind:TestArchive`.

I've added a test case that verifies an error is displayed if a `test` predicate is used.

sunshowers: Can `FilterBinaryMatch::Possible` ever be produced with this check in place? I think not, right? Could you add a `debug_assert` saying that it can't?

integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +842,113 @@ archive.include = [
...
+        create_archive_with_args("", false, "", &["-E", filter], false).expect("archive succeeded");
+    let file = File::open(archive_file).unwrap();
+    let decoder = zstd::stream::read::Decoder::new(file).unwrap();
+    let mut archive = tar::Archive::new(decoder);
+    let paths = archive
+        .entries()
+        .unwrap()
+        .map(|e| e.unwrap().path().unwrap().into_owned())
+        .collect::<Vec<_>>();
+    check_archive_contents(paths);
```
sunshowers: Naming the arg the same as the function is a bit confusing -- I'd just use `cb` for "callback" here.

integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +842,113 @@ archive.include = [
...
+                        .any(|folder| folder.file_name() == Some(OsStr::new("deps"))))
+                    .any(|path| {
+                        let file_name = path.file_name().unwrap();
+                        file_name
+                            .to_ascii_lowercase()
+                            .to_str()
+                            .unwrap()
+                            .contains(file)
+                    })
+            );
```
sunshowers, line 887: Include the name of the file you're checking in the assert message -- here, and in the other sections as appropriate.

integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +842,113 @@ archive.include = [
...
+                .to_str()
+                .unwrap()
+                .contains(expected_package_test_file)
+        }));
+        assert!(!paths.iter().any(|path| {
+            path.file_name()
+                .unwrap()
+                .to_str()
+                .unwrap()
+                .contains(filtered_test)
```
sunshowers: Use `unwrap_or_else` + `panic!` here with a descriptive error message.

integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +842,113 @@ archive.include = [
...
+        "nextest_tests",
+        "other",
+        "other",
+        "proc_macro_test",
+        "segfault",
+        "with_build_script",
+        "wrapper",
+        "nextest_tests",
+        "other",
+    ];
```
sunshowers: Can you use the list of fixtures as described here? That way when new fixtures get added this list won't have to be updated. https://github.com/nextest-rs/nextest/blob/main/fixture-data/src/nextest_tests.rs

integration-tests/tests/integration/snapshots/integration__archive_unsupported_build_filter.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: integration-tests/tests/integration/main.rs
+expression: output.stderr_as_str()
+---
+  error: predicate not allowed in `test-archive-filter` expressions
+   ╭────
+ 1 │ test(sample_test)
+   ·      ─────┬─────
+   ·           ╰── this predicate causes an unsupported expression
```
sunshowers: Could you reword this as "test predicates are not supported while archiving"? You may have to update the other messages as well instead of saying "this predicate causes {}".

sunshowers (commented):
Just a few more minor comments, thanks!

integration-tests/tests/integration/snapshots/integration__archive_unsupported_build_filter.snap
clundin55: Tweaked the wording a bit

integration-tests/tests/integration/main.rs
clundin55: Nice, thank you! Updated

integration-tests/tests/integration/main.rs
clundin55: Done!

integration-tests/tests/integration/main.rs
clundin55: Added assert messages!

integration-tests/tests/integration/main.rs
clundin55: Whoops, didn't realize I did that :)

Updated!

cargo-nextest/src/dispatch.rs
sunshowers: Just this comment to address, otherwise this is good to go!

cargo-nextest/src/dispatch.rs
```diff
@@ -1797,6 +1706,123 @@ impl App {
...
+        let filter_bound = {
+            if self.build_filter.ignore_default_filter {
+                FilterBound::All
+            } else {
+                FilterBound::DefaultSet
+            }
+        };
+
+        let filter_exprs = self.build_filtering_expressions(&pcx, FiltersetKind::TestArchive)?;
+        let test_filter_builder = self.build_filter.make_test_filter_builder(filter_exprs)?;
```
sunshowers: While testing this out, I realized that the archive command will now accept:

* `--partition`, which we don't support (since it filters by test) -- there's a separate issue for supporting partitioning by binary which isn't done here
* `--ignore-default-filter` which I think we should always ignore in this case (because it can contain test predicates)

Neither is desirable, and I think we should define a separate struct from `TestFilterBuilder` called `ArchiveFilterBuilder` for this.

integration-tests/tests/integration/main.rs
```diff
@@ -842,6 +843,113 @@ archive.include = [
         .expect_err("archive should have failed");
 }
 
+#[test]
+fn test_archive_with_build_filter() {
```
sunshowers, line 847: This should also verify that we can run tests out of these stripped archives, and in particular that we don't try to run test binaries that aren't in the archive.

sunshowers (commented):
A couple more comments from testing this out locally.

cargo-nextest/src/dispatch.rs
clundin55: Sorry missed this previously, just pushed a commit that I believe addresses this.

cargo-nextest/src/dispatch.rs
clundin55: Ah that's a good point. I added a new type, `ArchiveBuildFilter` to only define flags that make sense for the archive use case.

I then convert it internally to a `TestBuildFilter` type to avoid duplicating any code.

Happy to tweak this if you had a different idea.

integration-tests/tests/integration/main.rs
clundin55: Made an attempt at this, it's a bit rough but a starting point.

cargo-nextest/src/dispatch.rs
```diff
@@ -762,6 +765,57 @@ impl TestBuildFilter {
...
+
+    /// Test name filters and emulated test binary arguments.
+    ///
+    /// Supported arguments:{n}
+    /// - --ignored:         Only run ignored tests{n}
+    /// - --include-ignored: Run both ignored and non-ignored tests{n}
+    /// - --skip PATTERN:    Skip tests that match the pattern{n}
+    /// - --exact:           Run tests that exactly match patterns after `--`
+    #[arg(help_heading = None, value_name = "FILTERS_AND_ARGS", last = true)]
+    filters: Vec<String>,
```
sunshowers: Hmm, we can't support these either I'm pretty sure.

cargo-nextest/src/dispatch.rs
```diff
@@ -762,6 +765,57 @@ impl TestBuildFilter {
...
+    filterset: Vec<String>,
+
+    /// Ignore the default filter configured in the profile.
+    ///
+    /// By default, all filtersets are intersected with the default filter configured in the
+    /// profile. This flag disables that behavior.
+    ///
+    /// This flag doesn't change the definition of the `default()` filterset.
+    #[arg(long)]
+    ignore_default_filter: bool,
```
sunshowers: I don't think we should support `ignore_default_filter` -- unless you see a reason to do so?

cargo-nextest/src/dispatch.rs
sunshowers: Ah -- personally I'd avoid converting `ArchiveBuildFilter` to `TestBuildFilter`, and instead put common code in a function that can be called with the appropriate arguments.

integration-tests/tests/integration/main.rs
```diff
@@ -939,10 +1074,20 @@ fn create_archive(
 }
 
 fn run_archive(archive_file: &Utf8Path) -> (TempProject, Utf8PathBuf) {
+    run_archive_with_args(archive_file, RunProperty::Relocated as u64, None)
+}
+
+fn run_archive_with_args(
+    archive_file: &Utf8Path,
+    run_property: u64,
+    expected_exit_code: Option<i32>,
```
sunshowers: These tests seem great.

Why use `Option<i32>` and not explicitly pass in `NextestExitCode::TEST_RUN_FAILED`?

cargo-nextest/src/dispatch.rs
clundin55: Seems reasonable, I was hoping to avoid refactoring but I will give it a try.

cargo-nextest/src/dispatch.rs
clundin55: Removed

cargo-nextest/src/dispatch.rs
clundin55: Not really, I wasn't sure so I left them in. Removed.

integration-tests/tests/integration/main.rs
clundin55: Out of habit I use `Option`'s in test helpers. I removed the option and now pass in the i32 directly.

clundin55:
> Ah -- personally I'd avoid converting ArchiveBuildFilter to TestBuildFilter, and instead put common code in a function that can be called with the appropriate arguments.

Will do!

sunshowers:
Looks great! Thanks for working on this. Have a couple of minor tweaks to make but I'll do them myself.

sunshowers:
See https://github.com/nextest-rs/nextest/pull/2697 for the followups. Thanks again!

clundin55:
Thanks for the kind and prompt review @sunshowers ! Your rock !

PR #2676, opened 2025-10-20
Add the 'running' progress mode

glehmann:
It shows the currently running tests in addition to the progress bar. Used with --status-level=fail --final-status-level=none, it produces a much more compact output while keeping some visual informations on the running tests.

https://github.com/user-attachments/assets/439d2014-7290-4890-9d9d-a0be79ca83b2

Related discussion: https://github.com/nextest-rs/nextest/discussions/2624

In draft while:
* ~~https://github.com/console-rs/indicatif/pull/738 is not merged~~
* ~~The displayed time is updated without another test starting or finishing~~
* Added some tests
* https://github.com/console-rs/indicatif/pull/740 is not merged

### DIFF
--- a/nextest-runner/src/reporter/displayer/progress.rs
+++ b/nextest-runner/src/reporter/displayer/progress.rs
@@ -7,11 +7,11 @@ use crate::{
     list::TestInstanceId,
     reporter::{displayer::formatters::DisplayBracketedHhMmSs, events::*, helpers::Styles},
 };
+use iddqd::{IdHashItem, IdHashMap, id_upcast};
 use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
 use nextest_metadata::RustBinaryId;
 use owo_colors::OwoColorize;
 use std::{
-    collections::HashMap,
     env, fmt,
     io::{self, IsTerminal, Write},
     time::Duration,
@@ -39,17 +39,32 @@ pub enum ShowProgress {
     Running,
 }
 
-#[derive(Debug, PartialEq, Eq, Hash)]
-pub(super) struct RunningTestInstance {
+#[derive(Debug)]
+pub(super) struct TestProgressBar {
     binary_id: RustBinaryId,
     test_name: String,
+    progress_bar: ProgressBar,
+}
+
+impl IdHashItem for TestProgressBar {
+    type Key<'a> = TestInstanceId<'a>;
+
+    fn key(&self) -> Self::Key<'_> {
+        TestInstanceId {
+            binary_id: &self.binary_id,
+            test_name: &self.test_name,
+        }
+    }
+
+    id_upcast!();
 }
 
 #[derive(Debug)]
 pub(super) struct ProgressBarState {
     multi_progress: MultiProgress,
     bar: ProgressBar,
-    test_bars: Option<HashMap<RunningTestInstance, ProgressBar>>,
+    spinner_chars: &'static str,
+    test_bars: Option<IdHashMap<TestProgressBar>>,
     // Reasons for hiding the progress bar. We show the progress bar if none of
     // these are set and hide it if any of them are set.
     //
@@ -66,7 +81,12 @@ pub(super) struct ProgressBarState {
 }
 
 impl ProgressBarState {
-    pub(super) fn new(test_count: usize, progress_chars: &str, show_running: bool) -> Self {
+    pub(super) fn new(
+        test_count: usize,
+        progress_chars: &str,
+        spinner_chars: &'static str,
+        show_running: bool,
+    ) -> Self {
         let multi_progress = MultiProgress::new();
         // NOTE: set_draw_target must be called before enable_steady_tick to avoid a
         // spurious extra line from being printed as the draw target changes.
@@ -91,11 +111,12 @@ impl ProgressBarState {
         // Enable a steady tick 10 times a second.
         bar.enable_steady_tick(Duration::from_millis(100));
 
-        let test_bars = show_running.then_some(HashMap::new());
+        let test_bars = show_running.then_some(IdHashMap::new());
 
         Self {
             multi_progress,
             bar,
+            spinner_chars,
             test_bars,
             hidden_no_capture: false,
             hidden_run_paused: false,
@@ -160,7 +181,7 @@ impl ProgressBarState {
                 test_instance,
                 ..
             } => {
-                self.remove_test_bar(&test_instance.suite_info.binary_id, test_instance.name);
+                self.remove_test_bar(&test_instance.id());
 
                 self.hidden_between_sub_runs = false;
 
@@ -177,7 +198,10 @@ impl ProgressBarState {
                 self.bar.set_position(current_stats.finished_count as u64);
             }
             TestEventKind::TestAttemptFailedWillRetry { test_instance, .. } => {
-                self.remove_test_bar(&test_instance.suite_info.binary_id, test_instance.name);
+                self.remove_test_bar(&test_instance.id());
+                // TODO: it would be nice to count the delay down rather than
+                // up. But it probably will require changes to indicatif to
+                // enable that (or for us to render time ourselves).
                 self.add_test_bar(
                     "DELAY".style(styles.retry).to_string(),
                     &test_instance.suite_info.binary_id,
@@ -186,7 +210,7 @@ impl ProgressBarState {
                 );
             }
             TestEventKind::TestRetryStarted { test_instance, .. } => {
-                self.remove_test_bar(&test_instance.suite_info.binary_id, test_instance.name);
+                self.remove_test_bar(&test_instance.id());
                 self.add_test_bar(
                     "RETRY".style(styles.retry).to_string(),
                     &test_instance.suite_info.binary_id,
@@ -196,11 +220,9 @@ impl ProgressBarState {
             }
             TestEventKind::TestSlow { test_instance, .. } => {
                 if let Some(test_bars) = &mut self.test_bars {
-                    if let Some(tb) = test_bars.get(&RunningTestInstance {
-                        binary_id: test_instance.suite_info.binary_id.clone(),
-                        test_name: test_instance.name.to_owned(),
-                    }) {
-                        tb.set_prefix("SLOW".style(styles.skip).to_string());
+                    if let Some(tb) = test_bars.get_mut(&test_instance.id()) {
+                        tb.progress_bar
+                            .set_prefix("SLOW".style(styles.skip).to_string());
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


sunshowers:
Looks quite good! I think we should have:

* `--show-progress=running`
* `--show-progress=running-only` which is equivalent to `--show-progress=running --status-level=slow --final-status-level=none` (though the latter two can be overridden if desired)

glehmann:
I've added `--show-progress=running-only` and implemented a version using `MultiProgress` to let indicatif update the elapsed time.
I think the `MultipProgress` is the way to go, but I'm not happy with the number of extra threads it creates—1 for each running test progress. I wonder if the `MultiProgress` could also have the `enable_steady_tick()` and forward the tick to all the sub progress bars, instead of calling that method on all the progress bars…

glehmann:
@sunshowers I was going to add some tests, but I can't find a test using `--show-progress`, and I'm not sure I can do anything of actual value for the new progress modes based on the tests in `progress.rs`.
Would you happen to have any recommendations on how I might approach testing these new modes?

sunshowers:
Good question -- honestly I'm not sure how to test this well. The only things I can think about are things like ensuring that `test_bars` is updated properly on `TestStarted` and `TestFinished`. I think the rest would probably just be a matter of dogfooding this a bunch.

One question I have is, what happens if a test attempt fails and there's [a delay](https://nexte.st/docs/features/retries/#delays-and-backoff) before the test is retried? It would be nice to show a message saying we're waiting on the delay.

Also, what happens if the number of lines overflows the terminal?

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -30,11 +34,22 @@ pub enum ShowProgress {
...
+pub(super) struct RunningTestInstance {
+    binary_id: RustBinaryId,
+    test_name: String,
 }
 
 #[derive(Debug)]
 pub(super) struct ProgressBarState {
+    multi_progress: MultiProgress,
     bar: ProgressBar,
+    test_bars: Option<HashMap<RunningTestInstance, ProgressBar>>,
```
sunshowers: What order are running tests displayed in? Ideally tests would not shift around in the progress bar as new ones come in, though in case of an overflow/truncated output we would want to promote overflow-hidden tests to the top especially as some of them take a long time.

glehmann: The tests are shown in started-first, displayed-first order, which means the slow tests appear at the top.
This seems like the desired behavior to me, as the faster tests run and vanish too quickly to be read anyway.

glehmann:
> It's a bit messier when the lines are too large for the terminal. I'll see if it's possible to crop them to the terminal width.

The `{wide_msg}` tag effectively truncates the message to terminal width only when colors are off (`NO_COLOR=1`). With colored output, the truncation is incorrect. We may be pushing indicatif into uncharted territory ;)

glehmann:
github somehow discarded my previous reply? I'm rewriting it…

> One question I have is, what happens if a test attempt fails and there's [a delay](https://nexte.st/docs/features/retries/#delays-and-backoff) before the test is retried? It would be nice to show a message saying we're waiting on the delay.

This is not a feature I've used. I'll have to check.

> Also, what happens if the number of lines overflows the terminal?

The global bar is displayed, as are the other test bars that fit in the terminal. The other test bars are hidden.

glehmann:
> One question I have is, what happens if a test attempt fails and there's [a delay](https://nexte.st/docs/features/retries/#delays-and-backoff) before the test is retried? It would be nice to show a message saying we're waiting on the delay.

So currently, the delayed tests still appear as a regular running test. A `Delayed` status, instead of `Running` would indeed be nice :+1:

nextest-runner/src/reporter/displayer/progress.rs
sunshowers: Yeah that seems reasonable.

sunshowers:
> > It's a bit messier when the lines are too large for the terminal. I'll see if it's possible to crop them to the terminal width.
> 
> The `{wide_msg}` tag effectively truncates the message to terminal width only when colors are off (`NO_COLOR=1`). With colored output, the truncation is incorrect. We may be pushing indicatif into uncharted territory ;)

Definitely would like this to work correctly with colors. I think it's reasonable for indicatif to strip ANSI escapes while doing this computation. Otherwise we can do the computation ourselves as well (possibly by copy/pasting indicatif code and making modifications to it).

glehmann:
> > One question I have is, what happens if a test attempt fails and there's [a delay](https://nexte.st/docs/features/retries/#delays-and-backoff) before the test is retried? It would be nice to show a message saying we're waiting on the delay.
> 
> So currently, the delayed tests still appear as a regular running test. A `Delayed` status, instead of `Running` would indeed be nice 👍

There is now a `Delayed` status shown in front of the test name, in place of `Running`.

The current behavior:
* resets the elapsed time in the test bar when switching from running to delayed, and when switching from delayed to running. I think that's what I would want. Let me know if you think we should keep the elapsed time from the start test event.
* moves the test bar to the end of the list when switching from running to delayed and again from delayed to running. That seemed more natural this way to me, but we can keep the position in the list unchanged if you prefer.

I'm not using this feature—my opinion is more a guess. Let me know if you think of something better :)

glehmann:
> Definitely would like this to work correctly with colors. I think it's reasonable for indicatif to strip ANSI escapes while doing this computation. Otherwise we can do the computation ourselves as well (possibly by copy/pasting indicatif code and making modifications to it).

I've already made a PR :)
https://github.com/console-rs/indicatif/pull/740

The indicatif team has been super quick to release my previous fix. Hopefully it will be the same with this one.

sunshowers:
@glehmann is this ready for review?

glehmann:
I've been wanting to add tests, but even checking that `test_bars` is populated doesn't seem very useful. So I guess I'm procrastinating on that :)
So let's say it's ready for review, and I'll follow your opinion on the tests.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -131,6 +176,34 @@ impl ProgressBarState {
                 self.bar.set_length(current_stats.initial_run_count as u64);
                 self.bar.set_position(current_stats.finished_count as u64);
             }
+            TestEventKind::TestAttemptFailedWillRetry { test_instance, .. } => {
+                self.delete_test_bar(&test_instance.suite_info.binary_id, test_instance.name);
```
sunshowers: maybe "remove_test_bar"?

sunshowers:
Some notes from testing this:

- If the binary + test name wraps around the terminal, the rendering is pretty bad. We should instead cut off the test name to the terminal width (ideally replacing the initial part of the test name with ...) -- will require some math. Tests for this would also be nice.
- Consider dropping the Running prefix for tests (I think it would help with visual hierarchy).
- On pausing the test run with ctrl-Z on Unix, timers should be stopped (see the logic around RunContinued in `progress.rs` for how to do this)

glehmann:
> Some notes from testing this:
> 
>     * If the binary + test name wraps around the terminal, the rendering is pretty bad. We should instead cut off the test name to the terminal width (ideally replacing the initial part of the test name with ...) -- will require some math. Tests for this would also be nice.

The bad rendering is fixed with https://github.com/console-rs/indicatif/pull/740, but the new version has not yet been released.

Truncating the left part is tricky because it's done by indicatif directly, and it allows truncating on the left only if the text is aligned on the right—I don't think that's what we want.

We could do it by ourselves, but we wouldn't be able to react to a terminal resize as indicatif does.

>     * Consider dropping the Running prefix for tests (I think it would help with visual hierarchy).
>     * On pausing the test run with ctrl-Z on Unix, timers should be stopped (see the logic around RunContinued in `progress.rs` for how to do this)

Done and done!

sunshowers:
@glehmann looks like 740 might be in indicatif 0.18.2, could you double-check?

glehmann:
Oh, indicatif 0.18.2 is not tagged, but it's already on crates.io, and nextest is already using it. I'll rebase this branch on `main`

glehmann:
@sunshowers truncation should work properly now

sunshowers:
Made a bunch of minor tweaks, but otherwise looks good, thanks!

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -131,6 +197,35 @@ impl ProgressBarState {
...
+                self.remove_test_bar(&test_instance.id());
+                // TODO: it would be nice to count the delay down rather than
+                // up. But it probably will require changes to indicatif to
+                // enable that (or for us to render time ourselves).
+                self.add_test_bar(
+                    "DELAY".style(styles.retry).to_string(),
+                    &test_instance.suite_info.binary_id,
+                    test_instance.name,
+                    styles,
+                );
```
sunshowers, line 210: Curious how hard this would be -- it would be really cool to count delays down, but I think the current scheme is okay

glehmann: That would be nice, but indeed I don't think it's currently possible without rendering the time manually

sunshowers:
No worries, we can iterate on this as we go along -- this isn't part of the stable interface so improvements can keep coming. Thanks for working on it!

glehmann:
> Made a bunch of minor tweaks

I'm not sure about the spinners. They look OK for long-running tests, but they don't look good for fast-running tests IMO.

> otherwise looks good, thanks!

Thank you so much for creating such a useful tool!

sunshowers:
> I'm not sure about the spinners. They look OK for long-running tests, but they don't look good for fast-running tests IMO.

Curious about this -- it looks pretty good to me:

https://github.com/user-attachments/assets/19c746c5-3432-4847-8c7a-b131809ca114

(I also tried on WezTerm on Linux where it looks quite nice)

Wondering how it looks for you. Is it a matter of turning down the refresh rate for spinners?

glehmann:
They seem to shake randomly instead of the nice rotation effect. Not the best visual effect IMO, but nothing unbearable either :)

https://github.com/user-attachments/assets/d4b747a9-bdee-497e-8a37-41e3502c059c

sunshowers:
Thanks, I think slowing down the refresh rate should help with that -- will try out a few options.

sunshowers:
In https://github.com/nextest-rs/nextest/pull/2708 I duplicated each character to effectively halve the refresh rate. I think this looks okay, though we'll see what kind of feedback we get. I think showing something spinning is quite useful overall, means that nextest is doing something.

https://github.com/user-attachments/assets/5eb70773-fb8e-4e38-9d12-1b9a4d05e299

glehmann:
The spinners look much better, unfortunately the list of running tests doesn't always show the 8 (on my laptop) running tests:

https://github.com/user-attachments/assets/58a71d20-be1c-461f-8919-cdc482a04255

I'll look at your changes later and see if I can do something :)

sunshowers:
On main I made it do that tests are only displayed after 100ms have passed, which significantly reduced flickering. Might need to tweak the values a bit.

I think making this better would require switching to a different model like ratatui's inline viewport. A challenge with inline viewports is, right now they can't be resized (https://github.com/ratatui/ratatui/issues/984). https://github.com/ratatui/ratatui/pull/1964 looks like it should be able to do this.

PR #2697, opened 2025-10-25
[cargo-nextest] followups to #2670

sunshowers:
Just a few minor tweaks.


sunshowers:
Another change that needs to be made is to not include non-test binaries corresponding to test ones that get filtered out.

PR #2721, opened 2025-10-31
switch progress to ratatui inline viewport

sunshowers:
Still lots to do here, but putting this up for now.


sunshowers:
Unfortunately not feasible until https://github.com/ratatui/ratatui/issues/1426 is addressed. We've found a sufficient workaround in indicatif for now in https://github.com/nextest-rs/nextest/pull/2722.

PR #2723, opened 2025-10-31
[nextest-runner] with --show-progress=running, avoid output jumps up and down

sunshowers:
Use a free list of empty bars to provide padding corresponding to the number of formerly-used rows. This means that the overall progress bar slides up rapidly, but never slides down.

### DIFF
--- a/nextest-runner/src/reporter/displayer/progress.rs
+++ b/nextest-runner/src/reporter/displayer/progress.rs
@@ -12,6 +12,7 @@ use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
 use nextest_metadata::RustBinaryId;
 use owo_colors::OwoColorize;
 use std::{
+    collections::VecDeque,
     env, fmt,
     io::{self, IsTerminal, Write},
     time::Duration,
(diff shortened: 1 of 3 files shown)


sunshowers:
cc @glehmann would love a review from you -- this seems to fix the output sliding up and down quite nicely.

This isn't perfect but terminals were really not built for the output rapidly changing like this, unfortunately.

nextest-runner/src/runner/executor.rs
```diff
@@ -768,7 +768,7 @@ impl<'a> ExecutorContext<'a> {
         // Use an event that happens shortly after the test starts to avoid
         // overwhelming progress bar display code.
         let mut show_progress_sleep =
-            std::pin::pin!(crate::time::pausable_sleep(Duration::from_millis(100)));
+            std::pin::pin!(crate::time::pausable_sleep(Duration::from_millis(10)));
```
sunshowers, line 771: Also shortened this timeout -- 10ms is long enough that rapid-fire tests won't spam output, but is also less than 1 frame on 60fps.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -39,11 +40,151 @@ pub enum ShowProgress {
...
+        &mut self,
+        multi: &MultiProgress,
+        id: &TestInstanceId<'_>,
+        prefix: String,
+        styles: &Styles,
+    ) {
+        if let Some(tb) = self.used_bars.get(id) {
+            // Reset the bar to 0.
+            tb.bar.set_prefix(prefix);
+            tb.bar.clone().with_elapsed(Duration::ZERO);
```
glehmann, line 78: I don't think this is doing something useful, unless the new bar is reassigned to `tb.bar`.
Am I missing something?

glehmann (commented):
It completely removes the moving global bar for me

nextest-runner/src/reporter/displayer/progress.rs
sunshowers: This is a bit confusing for sure -- but `ProgressBar` is a lightweight `Arc`'d handle, and calling `with_elapsed` on the clone also affects the original. I'll add a comment explaining this, thanks.

sunshowers: https://github.com/console-rs/indicatif/pull/742 submitted upstream to add a `set_elapsed` method

glehmann: I completely missed that—thanks for the explanation!

Issue #2720, opened 2025-10-31
--show-progress=running polish


sunshowers:
While trying out `--show-progress=running` and `--show-progress=only`, I noticed a few places where we could be more polished:

* As progress bars are added or removed, the main progress bar shifts up and down. This is more pronounced with `--show-progress=running`.
* There's a bunch of flickering which we should try to avoid.

I think for a polished experience indicatif's model is probably not complete enough, and we'd need to switch to using ratatui's inline viewport. There are a few things to handle:

* the viewport height should be set to the number of test threads
* we do some careful poking around terminal handling (see https://nexte.st/docs/design/architecture/input-handling/) -- we'll want to probably move terminal handling into its own module that both input and progress handling report to
* we can't use any of the ratatui `init` functions directly, since they'll enable terminal raw mode which we don't want. Instead, we'll have to figure out the correct termios options ourselves depending on which combination of input handling and the progress bar is enabled.

sunshowers:
cc @glehmann if you'd like to tackle any of the above -- I think the current running implementation is good for a first pass but not quite to the polish level of the rest of the UI

sunshowers:
Started working on this in https://github.com/nextest-rs/nextest/pull/2721. Don't have the time to continue on this at the moment but definitely would like to get this fixed soon.

sunshowers:
Ah -- ratatui is probably not feasible until https://github.com/ratatui/ratatui/issues/1426 happens -- would definitely not want to lose overlong lines as currently output by nextest. Looking at other options.

glehmann:
I agree, this is not as clean as I'd like. I'd like to add to the list of problems:

* the missing test lines with short-lived tests, which make the user think that there are not as many tests running as there should be (https://github.com/nextest-rs/nextest/pull/2676#issuecomment-3469047669).

Note that the movement of the main progress bar is probably more pronounced because of it.

I haven't yet looked at ratatui, and it may well be a better option. On the indicatif side, I think we can do better by not deleting the test bar when a test ends, but instead replacing it with an empty bar.

sunshowers:
Yeah, here's my PR replacing it with an empty bar. Would love to get your eyes on it: https://github.com/nextest-rs/nextest/pull/2723

sunshowers:
#2722 was also crucial -- having the Tokio executor generate ticks instead of every progress bar doing its own thing -- means that all progress bar updates happen at the same time.

glehmann:
[...]

PR #2729, opened 2025-11-01
--show-progress=running enhancements

glehmann:
ref #2720

### DIFF
--- a/nextest-runner/src/reporter/displayer/progress.rs
+++ b/nextest-runner/src/reporter/displayer/progress.rs
@@ -13,10 +13,11 @@ use crate::{
 use iddqd::{IdHashItem, IdHashMap, id_upcast};
 use indexmap::IndexSet;
 use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
+use itertools::Itertools as _;
 use nextest_metadata::RustBinaryId;
 use owo_colors::OwoColorize;
 use std::{
-    collections::VecDeque,
+    cmp::max,
     env, fmt,
     io::{self, IsTerminal, Write},
     num::NonZero,
@@ -152,11 +153,6 @@ impl SummaryBar {
 
 #[derive(Debug)]
 pub(super) struct TestProgressBars {
-    // A list of currently unused progress bars.
-    //
-    // Once a test bar is removed, we add an empty bar, then use it to avoid the
-    // overall progress bar shifting rapidly up and down.
-    free_bars: VecDeque<ProgressBar>,
     used_bars: IdHashMap<TestProgressBar>,
 
     // Overflow tests in FIFO order (the IndexSet preserves insertion order).
@@ -164,7 +160,9 @@ pub(super) struct TestProgressBars {
     // Tests are displayed if they're in used_bars but not in overflow_order.
     overflow_queue: IndexSet<(RustBinaryId, String)>,
     // Maximum number of tests to display.
-    max_displayed: MaxProgressRunning,
+    max_progress_running: MaxProgressRunning,
+    // Actual maximum number of test displayed
+    max_displayed: usize,
 
     // Summary bar showing the overflow count. This is Some if the summary bar
     // is currently displayed.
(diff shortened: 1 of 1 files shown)


nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -502,7 +502,7 @@ impl ProgressBarState {
     ) -> Self {
         let multi_progress = MultiProgress::new();
         multi_progress.set_draw_target(Self::stderr_target());
-        multi_progress.set_move_cursor(true);
+        // multi_progress.set_move_cursor(true);
```
sunshowers, line 477: In my testing I found that `set_move_cursor(true)` made things a bit better -- did it make things worse for you?

nextest-runner/src/reporter/displayer/imp.rs
```diff
@@ -46,10 +46,10 @@ use std::{
 };
 
 /// The duration after which ticks are generated.
-pub const TICK_INTERVAL: Duration = Duration::from_millis(200);
+pub const TICK_INTERVAL: Duration = Duration::from_millis(50);
 
 /// The refresh rate for the progress bar, set to half the tick interval in hz.
-pub const PROGRESS_REFRESH_RATE_HZ: u8 = 10;
+pub const PROGRESS_REFRESH_RATE_HZ: u8 = 20;
```
sunshowers, line 52: Hmm, I feel like this might be too fast especially over a slower connection like sshing into a remote.

I think setting both to be the same might cause weird issues where both are slightly out of sync with each other. In any case the comment here needs to be updated.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -593,12 +594,6 @@ impl ProgressBarState {
                 if let Some(test_bars) = &mut self.test_bars {
                     test_bars.update_summary_bar(&self.multi_progress, *running, styles);
                 }
-            }
-            TestEventKind::TestShowProgress {
-                test_instance,
-                running,
-                ..
-            } => {
```
sunshowers, line 601: Hmm, again worried about this overwhelming the display with fast test suites like clap over a slower connection.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -743,8 +738,14 @@ impl ProgressBarState {
...
-        self.multi_progress
-            .suspend(|| std::io::stderr().write_all(buf))
+        // suspend forces a full redraw, so we call it only if there is
+        // something in the buffer
+        if !buf.is_empty() {
+            self.multi_progress
+                .suspend(|| std::io::stderr().write_all(buf))
+        } else {
+            Ok(())
+        }
```
sunshowers, line 729: This seems like a good change.

nextest-runner/src/reporter/displayer/progress.rs
glehmann: Enabling `set_move_cursor(true)` causes a visual bug where the end of a longer, previously drawn line isn't cleared when a shorter line overwrites it.
It was not visible before because of the total redraw when calling `suspend()`

sunshowers:
Something I noticed while testing this out is that we'd show "n-1 tests running" (e.g. 7 tests running or 31 tests running) a lot more than we do on main.

glehmann:
> Something I noticed while testing this out is that we'd show "n-1 tests running" (e.g. 7 tests running or 31 tests running) a lot more than we do on main.

Yes, I've noticed that too. I wonder if this is because of the blank line we're adding that may prevent the test from appearing. I'm preparing a change that adds the blank lines only before using `suspend()` to prevent the bar movement.

sunshowers:
I think we may need to improve our event model a bit — currently, when a test completes and the next test is started we send two separate events. I think we may need to coalesce them down into one event so that the number of running tests stays the same. (I learned this lesson in another event protocol I built later.)

nextest-runner/src/reporter/displayer/imp.rs
glehmann: I've reduced the refresh rate to 10 Hz. I'm still experimenting with the tick interval.

glehmann:
> I think we may need to improve our event model a bit — currently, when a test completes and the next test is started we send two separate events. I think we may need to coalesce them down into one event so that the number of running tests stays the same. (I learned this lesson in another event protocol I built later.)

Yes, it should help!

glehmann:
Creating a bar with spaces only when writing a message to the user with `write_buf` significantly reduced the flickering of the last line with `--show-progress=only`, but didn't remove it totally.

sunshowers:
> Creating a bar with spaces only when writing a message to the user with `write_buf` significantly reduced the flickering of the last line with `--show-progress=only`, but didn't remove it totally.

Fascinating, is that because the added bar is the very last one and there's some weird logic in indicatif around the last bar?

glehmann:
superseded by #2732

PR #2732, opened 2025-11-02
Implement test bar as a simple message

glehmann:
ref https://github.com/nextest-rs/nextest/issues/2720

I dropped the `MultiProgress` and reimplemented the test bar display as a simple progress bar message.
Almost no flickering now as a consequence :)
And I've implemented the cropping on the left instead of the right when the terminal is not large enough.

It should be quite easy to implement the countdown for the delayed tests. Edit: done.

I'm not pleased (at all) by how I passed the styles to the `tick()` function.
@sunshowers  Could you advise a better way to do that?

### DIFF
--- a/nextest-runner/src/reporter/displayer/imp.rs
+++ b/nextest-runner/src/reporter/displayer/imp.rs
@@ -46,9 +46,9 @@ use std::{
 };
 
 /// The duration after which ticks are generated.
-pub const TICK_INTERVAL: Duration = Duration::from_millis(200);
+pub const TICK_INTERVAL: Duration = Duration::from_millis(100);
 
-/// The refresh rate for the progress bar, set to half the tick interval in hz.
+/// The refresh rate for the progress bar, set to match the tick interval in hz.
 pub const PROGRESS_REFRESH_RATE_HZ: u8 = 10;
 
 pub(crate) struct DisplayReporterBuilder {


glehmann:
I haven't re-implemented the spinners:
* It should reduce the amount of data sent to the terminal
* There is the elapsed time in front of the test to show that the test is still running
* I still think it looks better without them
* I've already spent too much time on this feature :)

But of course I can re-add them, if necessary

sunshowers:
Oh this is really exciting! Let me give it a go.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -496,23 +217,19 @@ impl ProgressBarState {
     pub(super) fn new(
         test_count: usize,
         progress_chars: &str,
-        spinner_chars: &'static str,
+        _spinner_chars: &'static str,
```
sunshowers: Yeah go ahead and remove this.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -98,385 +97,107 @@ pub enum ShowProgress {
...
 pub(super) struct ProgressBarState {
-    multi_progress: MultiProgress,
     bar: ProgressBar,
-    test_bars: Option<TestProgressBars>,
+    stats: RunStats,
+    styles: Styles,
+    running: usize,
+    max_progress_running: MaxProgressRunning,
+    max_running_displayed: usize,
+    running_tests: Option<Vec<RunningTest>>,
```
sunshowers, line 204: I'd put in a comment here explaining that `None` means that running tests aren't being displayed.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -98,385 +97,107 @@ pub enum ShowProgress {
...
 #[derive(Debug)]
 pub(super) struct ProgressBarState {
-    multi_progress: MultiProgress,
     bar: ProgressBar,
-    test_bars: Option<TestProgressBars>,
+    stats: RunStats,
+    styles: Styles,
+    running: usize,
+    max_progress_running: MaxProgressRunning,
+    max_running_displayed: usize,
```
sunshowers, line 202: Is `max_running_displayed` just `max(length of running_tests vector, max_progress_running)`? If so we can probably avoid storing it here and just compute it each time.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -521,13 +238,17 @@ impl ProgressBarState {
...
-        let test_bars =
-            show_running.then(|| TestProgressBars::new(spinner_chars, max_progress_running));
+        let running_tests = show_running.then(Vec::new);
 
         Self {
-            multi_progress,
             bar,
-            test_bars,
+            stats: RunStats::default(),
+            styles: Styles::default(),
```
sunshowers, line 250: This isn't a huge deal but can you just pass the styles down everywhere as used to be done? Not quite following why that doesn't work any more.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -536,23 +257,67 @@ impl ProgressBarState {
...
+            self.max_running_displayed = max(self.max_running_displayed, count);
+            for running_test in &running_tests[..count] {
+                msg.push('\n');
+                msg.push_str(&running_test.message(&now, width, &self.styles));
+            }
+            msg.push_str(&"\n".to_string().repeat(self.max_running_displayed - count));
+            if count < running_tests.len() {
+                msg.push_str(&format!(
+                    "\n             ... and {} more tests running",
+                    running_tests.len() - count
```
sunshowers: use plural for `tests` here, plus `styles.count` for the count

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -574,39 +339,30 @@ impl ProgressBarState {
...
 
-                if let Some(test_bars) = &mut self.test_bars {
-                    test_bars.update_summary_bar(&self.multi_progress, *running, styles);
-                }
-            }
-            TestEventKind::TestShowProgress {
-                test_instance,
-                running,
-                ..
-            } => {
```
sunshowers, line 601: You can go ahead and remove this event I think, it's not relevant any more.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -536,23 +257,67 @@ impl ProgressBarState {
...
 
         match &event.kind {
             TestEventKind::StressSubRunFinished { .. } => {
                 // Clear all test bars to remove empty lines of output between
                 // sub-runs.
-                if let Some(test_bars) = &self.test_bars {
-                    test_bars.finish_and_clear_all();
-                }
+                // TODO: do we need to do something here?
+
```
sunshowers:
```suggestion
        match &event.kind {
            TestEventKind::StressSubRunStarted { .. } => {
                self.bar.reset();
            }
            TestEventKind::StressSubRunFinished { .. } => {
                // Clear all test bars to remove empty lines of output between
                // sub-runs.
                self.bar.finish_and_clear();

```

sunshowers:
(re the hakari failure, you can install or binstall cargo-hakari then run cargo hakari generate)

nextest-runner/src/reporter/displayer/progress.rs
glehmann: I think the current implementation is better when selecting only a few tests: it uses the number of lines required for these tests and not the number of possible parallel jobs.

nextest-runner/src/reporter/displayer/progress.rs
glehmann: The tick event is not going down the same code path as the other events, so I'm not sure at which point it's better to pass these.
Can I push the switch event up to `update_progress_bar` and remove the `tick()` methods?

sunshowers: Oh interesting, I see the issue. Honestly feel free to do anything reasonable here, and I'll poke at it later to see if I like something else more. The current implementation is fine.

nextest-runner/src/reporter/displayer/progress.rs
sunshowers: Ah, could you add a comment about that then?

nextest-runner/src/reporter/displayer/progress.rs
glehmann: Ok, I'll resist the temptation to try to make it more to my liking then :)

glehmann:
I've tried it on a few projects, and I must say I like how it looks this time :)

sunshowers (approved):
Looks fantastic!

PR #2740, opened 2025-11-04
[nextest-runner] use indicatif's println to print statuses

sunshowers:
The `println` function causes notably less flicker than `suspend` -- we arrange things such that message updates and newlines are part of the same terminal flush. (Spent some time in gdb walking over this code.)

Also bump up the default tick rate to 20/sec, which feels much nicer than 10/sec. Make this configurable via a semi-hidden environment variable.

### DIFF
--- a/nextest-runner/src/reporter/displayer/imp.rs
+++ b/nextest-runner/src/reporter/displayer/imp.rs
@@ -46,12 +46,6 @@ use std::{
     time::Duration,
 };
 
-/// The duration after which ticks are generated.
-pub const TICK_INTERVAL: Duration = Duration::from_millis(100);
-
-/// The refresh rate for the progress bar, set to match the tick interval in hz.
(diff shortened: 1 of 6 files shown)


sunshowers:
@glehmann could you give this PR a go, and have a look at the code changes? I spent some time in gdb observing the behaviors here.

This PR seems to basically eliminate flickering for me locally, and almost entirely over ssh as well. We can't get rid of it fully (especially over ssh) but this feels really good in practice.

glehmann:
It looks awesome! I can't see any flickering :)

glehmann:
I've tested a bit more, I see no flickering up to 20 running tests in `running` mode, and up to `24` in `only` mode.
So apparently the flickering appears because we are dumping a lot of text for the running tests.

I doubt displaying that much test lines is useful anyway, so this is all good IMO :)

PR #2742, opened 2025-11-05
feat: treat timeouts as failure or success based on config

### LINKED ISSUE #1291: Option to treat timeouts as non-failure
I am currently using `nextest` via `miri` in Wasmi to run the entire Wasm spec testsuite (which is huge) nightly. Since `miri` execution with all of its checks is rather slow and since we use default GitHub action runners most tests timeout. Locally I ran the entire testsuite via `miri` with a timeout of 4h per test case and even then I had some timeouts, so that's how bad it really is. However, this is not as bad as it sounds since even timed out tests produced a ton of checks and validated their input until their timeout.

So what would be nice is to make it possible via config to treat timed out tests as non-failures since I only really want to be notified by this CRON `miri` run if it actually found an actual bug.

Is this maybe even already possible? In case you do not want to implement this feature what would be the best way to handling this on my side?

Link to the cron job if that is of interest: https://github.com/paritytech/wasmi/actions/runs/7771389011/job/21192633974

Thanks!

eduardorittner:
Adds a new configuration option inside "slow-timeout": "on-timeout" which when set to "pass" makes nextest treat timed out tests as passing.

This PR adds a new `SlowTimeoutResult` struct similar to `LeakTimeoutResult` which characterizes a timeout as passing or failing.

There are some things which need to be addressed, and most of them are marked with `TODO`s in the code, these should all be fixed before merging. Most of them have to do with how to display this information, since we now have a new category of tests to display: tests that passed with a timeout. Any opinions here would be great!
There are some failing tests due to the formatting changes made, I decided not to accept the snapshots just yet since I'm not sure this is the final formatting we're going to end up with.

P.S.: I noticed nextest follows atomic commits, I figured only one commit would be enough for this change but I can split it up into smaller commits if it's deemed better.

Closes #1291

### DIFF
--- a/nextest-runner/tests/integration/basic.rs
+++ b/nextest-runner/tests/integration/basic.rs
@@ -139,7 +139,8 @@ fn test_timeout_with_retries() -> Result<()> {
                 assert_eq!(
                     run_statuses.len(),
                     1,
-                    "{test_name} should have been run 3 times (1 initial + 2 retries)",
+                    "{test_name} should have been run exactly once \
+                     (timed out tests that pass are not retried)",
                 );
 
                 let status = run_statuses.last_status();
@@ -972,7 +973,7 @@ fn test_override_timeout_result() -> Result<()> {
     let (instance_statuses, run_stats) = execute_collect(runner);
 
     println!("{instance_statuses:?}");
-    assert_eq!(run_stats.finished_count, 3, "4 tests should have finished");
+    assert_eq!(run_stats.finished_count, 3, "3 tests should have finished");
     assert_eq!(
         run_stats.passed_timed_out, 1,
         "1 test should pass with timeout"


eduardorittner:
Just confirmed that the failing tests in CI are these 3, which are snapshot tests for the formatting output. So all other tests are passing
```
nextest-runner reporter::displayer::imp::tests::test_info_response
nextest-runner reporter::displayer::imp::tests::test_summary_line
nextest-runner reporter::displayer::progress::tests::progress_str_snapshots
```

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -15,6 +15,12 @@ pub struct SlowTimeout {
     pub(crate) terminate_after: Option<NonZeroUsize>,
     #[serde(with = "humantime_serde", default = "default_grace_period")]
     pub(crate) grace_period: Duration,
+    #[serde(default, rename = "on-timeout")]
+    pub(crate) timeout_result: SlowTimeoutResult,
+}
+
+fn default_grace_period() -> Duration {
+    Duration::from_secs(10)
 }
```
sunshowers: any reason this was moved?

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -15,6 +15,12 @@ pub struct SlowTimeout {
     pub(crate) terminate_after: Option<NonZeroUsize>,
     #[serde(with = "humantime_serde", default = "default_grace_period")]
     pub(crate) grace_period: Duration,
+    #[serde(default, rename = "on-timeout")]
+    pub(crate) timeout_result: SlowTimeoutResult,
```
sunshowers: can we just call this field `on_timeout`?

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -24,13 +30,10 @@ impl SlowTimeout {
         period: far_future_duration(),
         terminate_after: None,
         grace_period: Duration::from_secs(10),
+        timeout_result: SlowTimeoutResult::Fail,
```
sunshowers: An inconsistency here -- we use `SlowTimeoutResult::Fail` here but `default` below. Let's pick one or the other -- I'm partial to using `Fail` and removing the `Default` impl, unless there's a good reason to retain it.

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -76,6 +80,18 @@ where
...
 
+/// The result of controlling slow timeout behavior.
+#[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum SlowTimeoutResult {
+    #[default]
+    /// The test is marked as failed.
+    Fail,
+
+    /// The test is marked as passed.
```
sunshowers, line 98: I'd put in an explanation here because "the test is marked as passed" is pretty strange on a timeout being hit. Just talk about the fact that there are some use cases that benefit from being run up to a time limit.

nextest-runner/src/reporter/displayer/imp.rs
```diff
@@ -2256,30 +2273,36 @@ fn short_status_str(result: ExecutionResult) -> Cow<'static, str> {
         ExecutionResult::Leak {
             result: LeakTimeoutResult::Fail,
         } => "LKFAIL".into(),
-        ExecutionResult::Timeout => "TMT".into(),
+        ExecutionResult::Timeout {
+            result: SlowTimeoutResult::Pass,
+            // TODO: what string to put here?
+        } => "TMTPAS".into(),
```
sunshowers: Ah this is definitely difficult. What about `TMPASS`?

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -753,15 +1010,22 @@ pub(super) fn write_summary_str(run_stats: &RunStats, styles: &Styles, out: &mut
...
                 "{} {}",
                 passed_slow.style(styles.count),
                 "slow".style(styles.skip),
             ));
         }
+        if passed_timed_out > 0 {
+            text.push(format!(
+                "{} {}",
+                passed_timed_out.style(styles.count),
+                "timed out".style(styles.skip),
```
sunshowers: honestly I'd skip showing something special for `passed_timed_out`. These are relatively niche tests.

nextest-runner/src/reporter/events.rs
```diff
@@ -519,23 +541,26 @@ pub struct RunStats {
...
     pub passed_slow: usize,
 
+    /// The number of timed out tests that passed.
+    pub passed_timed_out: usize,
+
     /// The number of tests that passed on retry.
     pub flaky: usize,
 
-    /// The number of tests that failed.
+    /// The number of tests that failed. TODO: what fields are included here?
```
sunshowers: You can see the logic here: https://github.com/nextest-rs/nextest/blob/342aa178837750acc63c1e341c0633bbac729e09/nextest-runner/src/reporter/events.rs#L682

nextest-runner/src/reporter/events.rs
```diff
@@ -640,7 +665,8 @@ impl RunStats {
             ExecutionResult::ExecFail => {
                 self.setup_scripts_exec_failed += 1;
             }
-            ExecutionResult::Timeout => {
+            ExecutionResult::Timeout { .. } => {
+                // TODO: can setup scripts timeout and still pass?
```
sunshowers: Let's say no to this. I don't think there are any legitimate use cases for it.

nextest-runner/src/reporter/events.rs
```diff
@@ -694,7 +720,21 @@ impl RunStats {
...
+                self.passed += 1;
+                self.passed_timed_out += 1;
+                if run_statuses.len() > 1 {
+                    self.flaky += 1;
+                }
+            }
+            ExecutionResult::Timeout {
+                result: SlowTimeoutResult::Fail,
+            } => {
+                self.failed += 1;
```
sunshowers: Note this is a logic change and makes the `failed_count` function incorrect. Let's address this in a followup rather than changing the semantics here.

site/src/docs/configuration/reference.md
```diff
@@ -145,8 +145,16 @@ Profiles are configured under `[profile.<name>]`. The default profile is called
...
   slow-timeout = { period = "120s", terminate-after = 2, grace-period = "10s" }
+  # or
+  slow-timeout = { period = "30s", terminate-after = 4, on-timeout = "pass" }
   ```
 
+The `slow-timeout` object supports the following parameters:
+- `period`: Time period after which a test is considered slow
+- `terminate-after`: Number of periods after which to terminate the test
+- `grace-period`: Time to wait for graceful shutdown before force termination
+- `on-timeout`: <!-- md:version 0.9.88 --> What to do when a test times out (`"fail"` or `"pass"`)
```
sunshowers: The version number should be 0.9.112.

site/src/docs/features/slow-tests.md
```diff
@@ -42,6 +42,24 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
 slow-timeout = { period = "30s", terminate-after = 4 }
 ```
 
+### Configuring timeout behavior
+
+<!-- md:version 0.9.88 -->
```
sunshowers: Same here.

site/src/docs/features/slow-tests.md
```diff
@@ -42,6 +42,24 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
...
+
+By default, tests that time out are treated as failures. You can configure this behavior using the `on-timeout` parameter. For example, to treat timeouts as successes:
```
sunshowers: Definitely mention some use cases here because this is pretty  nonstandard at first glance. In general, it's important to justify especially nonobvious features: nextest is very feature-rich, but all of the features have real reasons behind them.

sunshowers (commented):
Thanks. One commit is fine for this change (though there's one particular thing in this PR that should be in a followup).

nextest-runner/src/config/elements/slow_timeout.rs
eduardorittner: Nope, was a mistake

nextest-runner/src/config/elements/slow_timeout.rs
eduardorittner: Sure!

nextest-runner/src/config/elements/slow_timeout.rs
eduardorittner: The only issue with not having a `SlowTimeoutResult::default` is that AFAICT (I'm not that familiar with serde) the `#[serde(default)]` line in `SlowTimeout`'s definition will have to be something like `#[serde(default = "default_result")]` where `default_result` is a custom function that returns the default value for `SlowTimeoutResult`. At that point I think it's just better to just keep the derived `default` and change this usage from `SlowTimeoutResult::Fail` to `default`.

From
```
pub struct SlowTimeout {
    // Omitted fields
    #[serde(default)]
    pub(crate) on_timeout: SlowTimeoutResult,
}
```

To
```
pub struct SlowTimeout {
    // Omitted fields
    #[serde(default = "default_on_timeout")]
    pub(crate) on_timeout: SlowTimeoutResult,
}
```

nextest-runner/src/config/elements/slow_timeout.rs
eduardorittner: What do you think of something like this?

```
/// The result of controlling slow timeout behavior.
///
/// For most situations, a timed out test should be marked failing. However, there are certain
/// classes of tests which are expected to run indefinetely long (or just too long), like fuzzing
/// and property tests, which both explore a huge state space. For these tests it's nice to be able
/// to treat a timeout as a success since, they generally check for invariants and other properties
/// of the code under test. A timeout in this context doesn't necessarily mean that there are no
/// failing inputs, just that they weren't found up until that moment, which is still valuable
/// information.
#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
#[serde(rename_all = "kebab-case")]
pub enum SlowTimeoutResult {
    /// The test is marked as failed.
    Fail,

    /// The test is marked as passed.
    Pass,
}
```

nextest-runner/src/reporter/displayer/progress.rs
eduardorittner: Makes sense

nextest-runner/src/reporter/events.rs
eduardorittner: Thanks!

nextest-runner/src/reporter/events.rs
eduardorittner: I see, should I also move the changes to `RunStats` (adding `passed_timed_out` and renaming `timed_out` to `failed_timed_out`) into that followup? Or keep the struct changes in the first commit and address only the logic changes in a followup?

site/src/docs/features/slow-tests.md
eduardorittner: Added! See what you think

nextest-runner/src/config/elements/slow_timeout.rs
eduardorittner: I just noticed something, since `VERY_LARGE` is `const`, I can't actually call `SlowTimeoutResult::default()` since it's not `const`, so now I'm unsure what to do:
1. Keep `#[derive(Default)]` in `SlowTimeoutResult`, use `SlowTimeoutResult::default` in non-`const` contexts and `SlowTimeoutResult::Fail` in `const` contexts.
2. Keep `#[derive(Default)]` just for `#[serde(default)]`, use `SlowTimeoutResult::Fail` everywhere else
3. Remove `#[derive(Default)]`, add a `pub const fn default() -> Self` which can be used everywhere. This does mean that instead of `#[serde(default)]` we'll have `#[serde(default = "SlowTimeoutResult::default")]` in `SlowTimeout`'s definition.

Option 1 doesn't seem that great to me, I think I prefer option 3, but option 2 is also fine.

eduardorittner:
So I've split off the changes into two commits, where the second one only deals with the logic in `reporter/events.rs` for counting the number of failed tests, timed out tests, etc. Some tests are now also failing because I made `self.failed` include `self.failed_timed_out`, instead of them being disjoint, which broke a bunch of the tests that relied on this logic. I can revert this and make them disjoint again if that's better

nextest-runner/src/config/elements/slow_timeout.rs
sunshowers: Thanks. This is pretty good. Some notes:

* It's "property-based tests" -- "property testing" has a different meaning.
* Generally PBT tends to run for a limited number of iterations -- the confidence in them comes from the fact that they get run in CI all the time. So I'd drop the note about them, and instead just say fuzzing.

nextest-runner/src/reporter/events.rs
sunshowers: The struct changes have to go in the first commit, I think. The semantic changes to the meaning of things like `failed` should go in the followup.

site/src/docs/features/slow-tests.md
```diff
@@ -42,6 +42,21 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
 slow-timeout = { period = "30s", terminate-after = 4 }
 ```
 
+### Configuring timeout behavior
+
+<!-- md:version 0.9.112 -->
+
+By default, tests that time out are treated as failures. For fuzz and property tests with very large state spaces (or on a constrained environment like CI), it may be useful to treat timeouts as successes, since they're not expected to test every possible input. A timeout in this context is not the same as a failure, it just means they weren't able to find any input that caused the test to fail. You can configure this behavior using the `on-timeout` parameter. For example, to treat timeouts as successes:
```
sunshowers: Same comment here, I'd drop the mention of PBT.

nextest-runner/src/reporter/events.rs
sunshowers: Ah the failing tests are unfortunate -- can we split them up into another PR?

nextest-runner/src/runner/executor.rs
```diff
@@ -821,7 +830,7 @@ impl<'a> ExecutorContext<'a> {
                                 job.as_ref(),
                                 slow_timeout.grace_period,
                             ).await;
-                            status = Some(ExecutionResult::Timeout);
+                            status = Some(ExecutionResult::Timeout {result: slow_timeout.on_timeout});
```
sunshowers, line 879: rustfmt should have fixed this -- did that not work? (Oh maybe not because this is inside a `tokio::select!` arm, ugh.)

eduardorittner: Would the "correct" formatting be this?

```rust
status = Some(ExecutionResult::Timeout { result: slow_timeout.on_timeout });
```

eduardorittner:
Tests are now passing, I changed the logic in `events.rs` so that `self.failed` and `self.failed_timed_out` are disjoint and `self.failed_count()` is now

```rust
pub fn failed_count(&self) -> usize {
    self.failed + self.exec_failed + self.failed_timed_out
}
```

I haven't added any new tests, but I'll get started on that now that all the old existing tests are passing.

nextest-runner/src/runner/executor.rs
sunshowers: Yeah, either that or across multiple lines:

```rust
status = Some(ExecutionResult::Timeout {
    result: slow_timeout.on_timeout,
});

**Comment by sunshowers** [comment]
Looks great -- yeah, just missing new tests otherwise this is good to go.

**Comment by eduardorittner** [comment]
I have two questions regarding junit:
1. Should passing timed out tests be reported as failures? I ask because in `reporter/aggregator/junit.rs` passing tests that leaked are reported as `NonSuccessKind::Error`: https://github.com/eduardorittner/nextest/blob/42f92f1ed6c33eba1965f93bdb862548764892b3/nextest-runner/src/reporter/aggregator/junit.rs#L353-L358

2. How/where is junit support tested?

**Comment by sunshowers** [comment]
1. Good question. I think it should be marked as passed.
2. Uhmmm, there aren't good tests for deserializing JUnit reports. But there probably should be. I wouldn't worry about that, at some point I'll have Claude Code or something write those tests.

**Comment by sunshowers** [comment]
> Uhmmm, there aren't good tests for deserializing JUnit reports. But there probably should be. I wouldn't worry about that, at some point I'll have Claude Code or something write those tests.

Did this in #2761.

**Comment by eduardorittner** [comment]
Hi! Something I just thought of, how should retries be handled when `on-timeout="pass"` is configured? Right now what happens is that timed out tests are not retried, since they were marked as passing the first time around. This behavior makes sense to me (the alternative doesn't seem that great, to retry "passing" tests until the maximum number of retries just to mark them as passing again), but is maybe not that intuitive? Maybe issue a warning when both `on-timeout="pass"` and `retries` are configured?

**Comment by sunshowers** [comment]
@eduardorittner I think not retrying is correct -- I also don't think a warning is necessary here, because the rules are very simple (pass = no retry, fail for any reason = retry).

**Comment by eduardorittner** [comment]
Added some new tests but they're now failing on windows, will try to fix that once I have some more time

**Comment by sunshowers** [comment]
I'm probably missing something but it looks like some of the comments you marked resolved aren't actually resolved yet -- are you planning to work on them?

**Comment by eduardorittner** [comment]
yeah I'm sorry, hadn't even noticed that. I was working from two different computers and without realising it lost some of the changes I made whoch addressed them. I'll go back through every one to fix them again. My bad!

**Comment by eduardorittner** [comment]
I've now gone over the resolved comments and I *think* I've re-resolved them all!

File: fixtures/nextest-tests/tests/basic.rs [hunk 15]
```diff
@@ -319,6 +319,16 @@ fn test_slow_timeout_subprocess() {
...
 }
 
+#[test]
+#[ignore]
+fn test_flaky_slow_timeout_mod_3() {
+    let nextest_attempt = nextest_attempt();
+    if nextest_attempt % 3 != 0 {
+        panic!("Failed because attempt {} % 3 != 0", nextest_attempt)
+    }
+    std::thread::sleep(std::time::Duration::from_secs(4));
```
sunshowers, line 330: Where does this number come from? See the comments in the lines above, could you add a similar explanation?

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -76,6 +80,25 @@ where
     deserializer.deserialize_any(V)
 }
 
+/// The result of controlling slow timeout behavior.
+///
+/// In most situations a timed out test should be marked failing. However, there are certain
+/// classes of tests which are expected to run indefinetely long, like fuzzing, which explores a
```
sunshowers:
```suggestion
/// classes of tests which are expected to run indefinitely long, like fuzzing, which explores a
```

nextest-runner/src/config/elements/slow_timeout.rs
```diff
@@ -76,6 +80,25 @@ where
...
+/// huge state space. For these tests it's nice to be able to treat a timeout as a success since
+/// they generally check for invariants and other properties of the code under test during their
+/// execution. A timeout in this context doesn't mean that there are no failing inputs, it just
+/// means that they weren't foudn up until that moment, which is still valuable information.
```
sunshowers:
```suggestion
/// means that they weren't found up until that moment, which is still valuable information.
```

nextest-runner/src/reporter/displayer/imp.rs
```diff
@@ -2273,7 +2296,12 @@ fn short_status_str(result: ExecutionResult) -> Cow<'static, str> {
         ExecutionResult::Leak {
             result: LeakTimeoutResult::Fail,
         } => "LKFAIL".into(),
-        ExecutionResult::Timeout => "TMT".into(),
+        ExecutionResult::Timeout {
+            result: SlowTimeoutResult::Pass,
+        } => "TMTPAS".into(),
```
sunshowers:
```suggestion
        } => "TMPASS".into(),
```

site/src/docs/configuration/reference.md
```diff
@@ -145,8 +145,16 @@ Profiles are configured under `[profile.<name>]`. The default profile is called
...
   slow-timeout = { period = "120s", terminate-after = 2, grace-period = "10s" }
+  # or
+  slow-timeout = { period = "30s", terminate-after = 4, on-timeout = "pass" }
   ```
 
+The `slow-timeout` object supports the following parameters:
+- `period`: Time period after which a test is considered slow
+- `terminate-after`: Number of periods after which to terminate the test
+- `grace-period`: Time to wait for graceful shutdown before force termination
+- `on-timeout`: <!-- md:version 0.9.112 --> What to do when a test times out (`"fail"` or `"pass"`)
```
sunshowers:
```suggestion
- `on-timeout`: <!-- md:version 0.9.114 --> What to do when a test times out (`"fail"` or `"pass"`)
```

Also, use a definition list here: https://squidfunk.github.io/mkdocs-material/reference/lists/#using-definition-lists

site/src/docs/features/slow-tests.md
```diff
@@ -43,6 +43,21 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
 slow-timeout = { period = "30s", terminate-after = 4 }
 ```
 
+### Configuring timeout behavior
+
+<!-- md:version 0.9.112 -->
```
sunshowers:
```suggestion
<!-- md:version 0.9.114 -->
```

site/src/docs/configuration/reference.md
```diff
@@ -145,8 +145,16 @@ Profiles are configured under `[profile.<name>]`. The default profile is called
   slow-timeout = "60s"
   # or
   slow-timeout = { period = "120s", terminate-after = 2, grace-period = "10s" }
+  # or
+  slow-timeout = { period = "30s", terminate-after = 4, on-timeout = "pass" }
   ```
 
+The `slow-timeout` object supports the following parameters:
```
sunshowers:
```suggestion
The `slow-timeout` object accepts the following parameters:
```

site/src/docs/features/slow-tests.md
```diff
@@ -43,6 +43,21 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
...
+
+By default, tests that time out are treated as failures. For fuzz tests with very large state spaces (or on a constrained environment like CI), it may be useful to treat timeouts as successes, since they're usualy not expected to run until completion, which in this case would mean testing every possible input. A timeout in this context is not the same as a failure, it just means that no failing input was found up until this point. You can configure this behavior using the `on-timeout` parameter. To treat timeouts as successes:
```
sunshowers:
```suggestion
By default, tests that time out are treated as failures. However, for fuzz tests with very large state spaces (or on a constrained environment like CI), it may be useful to treat timeouts as successes, since they're usually not expected to run until completion. A timeout in this context means that no failing input was found up until this point.

For such tests, you can configure timeouts to be marked as successes. For example, to run tests in the `fuzz-targets` crate for 30 seconds, then mark them as successes:
 
```

site/src/docs/features/slow-tests.md
```diff
@@ -43,6 +43,21 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
...
+
+```toml title="Timeouts as successes"
+[profile.default]
+slow-timeout = { period = "30s", terminate-after = 4, on-timeout = "pass" }
+```
```
sunshowers, line 78:
```suggestion
```toml title="Timeouts as successes"
[[profile.default.overrides]]
filter = 'package(fuzz-targets)'
slow-timeout = { period = "30s", terminate-after = 1, on-timeout = "pass" }
```
```

File: site/src/docs/features/slow-tests.md [hunk 24]
```diff
@@ -43,6 +43,21 @@ Nextest lets you optionally specify a number of `slow-timeout` periods after whi
...
+
+The possible values for `on-timeout` are:
+- `"fail"` (default): Tests that time out are treated as failures
+- `"pass"`: Tests that time out are treated as successes
```
sunshowers: Use a definition list here: https://squidfunk.github.io/mkdocs-material/reference/lists/#using-definition-lists

sunshowers:
Catching up with things after a week off -- have a big backlog, planning to look at this later this week.

site/src/docs/configuration/reference.md
```diff
@@ -145,8 +145,16 @@ Profiles are configured under `[profile.<name>]`. The default profile is called
...
   slow-timeout = { period = "120s", terminate-after = 2, grace-period = "10s" }
+  # or
+  slow-timeout = { period = "30s", terminate-after = 4, on-timeout = "pass" }
   ```
 
+The `slow-timeout` object accepts the following parameters:
+- `period`: Time period after which a test is considered slow
+- `terminate-after`: Number of periods after which to terminate the test
+- `grace-period`: Time to wait for graceful shutdown before force termination
+- `on-timeout`: <!-- md:version 0.9.114 --> What to do when a test times out (`"fail"` or `"pass"`)
```
sunshowers: Please update this and the below case to 0.9.115.

eduardorittner:
Updated docs references to 115, and illumos test are now green. I'll try to take a look at the old failing logs and see if I can find anything

sunshowers:
With apologies, I need to tackle something urgently at work this week. Hoping to look at this by the end of the week.

eduardorittner:
No worries! I'm not in a rush, take as much time as you need 😁

sunshowers:
Thank you!

PR #2754, opened 2025-11-11
Fix broken link to Stability policy

JustusFluegel:
Fixes a broken markdown link to the stability policy from the "Installing from source" section on the website

### DIFF
--- a/site/src/docs/installation/from-source.md
+++ b/site/src/docs/installation/from-source.md
@@ -18,10 +18,10 @@ cargo install cargo-nextest --locked
 
     A plain `cargo install cargo-nextest` without `--locked` is **not supported**. If you run into build issues, please try with `--locked` before reporting an issue.
 
-`cargo nextest` must be compiled and installed with **Rust {{ msrv }}** or later (see [_Stability policy_](../stability/index.md) for more), but it can build and run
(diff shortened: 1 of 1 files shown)


site/src/docs/installation/from-source.md
```diff
@@ -18,7 +18,7 @@ cargo install cargo-nextest --locked
 
     A plain `cargo install cargo-nextest` without `--locked` is **not supported**. If you run into build issues, please try with `--locked` before reporting an issue.
 
-`cargo nextest` must be compiled and installed with **Rust {{ msrv }}** or later (see [_Stability policy_ for more), but it can build and run
+`cargo nextest` must be compiled and installed with **Rust {{ msrv }}** or later (see [_Stability policy_](../stability/index.md) for more), but it can build and run
```
sunshowers: I think you can just have a closing bracket here, since it's defined at line 24 below.

JustusFluegel: Does that work even with the italic formatting?

sunshowers: Ah I think you'll need to add the same italics to the reference below.

JustusFluegel: :+1:

sunshowers:
Thanks!

Issue #2372, opened 2025-05-30
Bug: WebAssembly Target Runner Support


insipx:
### Description of the issue

**Description:**
Running WebAssembly tests with `wasm32-unknown-unknown` seem to compile and run, but webassembly tests in fact are not running

**Steps to reproduce:**
1. Create webassembly test with wasm_bindgen, `#[wasm_bindgen_test]` 
```rust
#[wasm_bindgen_test(unsupported = test)]
fn wasm_test() {
    if cfg!(target_arch = "wasm32") {
        panic!("should panic if running in webassembly");
    }
    panic!("panic not triggered at all");
    println!("did not run webassembly code");
}
```

2. set `wasm-bindgen-test-runner` to the test runner in .cargo
```toml
[target.wasm32-unknown-unknown]
runner = 'wasm-bindgen-test-runner'

```
3. run tests with nextest
`cargo nextest run --target wasm32-unknown-unknown` according to https://nexte.st/docs/features/target-runners/

following these steps, the tests fail b/c they're not being run in webassembly

It seems like according to these issues in wasm-bindgen, this should be supported:
https://github.com/rustwasm/wasm-bindgen/pull/4356
https://github.com/rustwasm/wasm-bindgen/issues/3891#issuecomment-2048580009

### Expected outcome

test should run in webassembly, and fail with a panic: "should panic if running in webassembly"
<details>

<summary>`cargo test --target wasm32-unknown-unknown` output</summary>

```bash
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 filtered out; finished in 0.01s
console.log div contained:
    panicked at src/lib.rs:14:13:
    should panic if running in webassembly

    Stack:

    Error
        at http://127.0.0.1:43923/wasm-bindgen-test:518:21
        at logError (http://127.0.0.1:43923/wasm-bindgen-test:15:18)
        at imports.wbg.__wbg_new_78093c5bd701d017 (http://127.0.0.1:43923/wasm-bindgen-test:517:66)
        at test_nextest-373bc291c5c6fd67.wasm.__wbg_new_78093c5bd701d017 externref shim (http://127.0.0.1:43923/wasm-bindgen-test_bg.wasm:wasm-function[1642]:0x71b15)
[...]

PR #2763, opened 2025-11-12
Fix parsing wasm-bindgen-test-runner test list output

weiznich:
Newer versions of wasm-bindgen-test-runner (>=0.2.105?) will emit an empty line in the end of the test list output given by `--list --format terse`. This commit changes the parsing logic in cargo-nextest to just ignore empty (only whitespace) lines in this output to make cargo nextest compatible again with wasm-bindgen-test-runner.

### DIFF
--- a/nextest-runner/src/list/test_list.rs
+++ b/nextest-runner/src/list/test_list.rs
@@ -675,8 +675,14 @@ impl<'g> TestList<'g> {
         // <test name>: test
         // ...
 
-        list_output.lines().map(move |line| {
-            line.strip_suffix(": test")
+        list_output.lines().filter_map(move |line| {
+            // we filter out empty lines here
+            // as certain wasm-bindgen-test-runner versions
+            // emit an empty line in the end of the output
+            if line.trim().is_empty() {
+                None
+            } else {
+                Some(line.strip_suffix(": test")
                 .or_else(|| line.strip_suffix(": benchmark"))
                 .ok_or_else(|| {
                     CreateTestListError::parse_line(
@@ -686,7 +692,8 @@ impl<'g> TestList<'g> {
                         ),
                         list_output,
                     )
-                })
+                }))
+            }
         })
     }
 


sunshowers:
Thanks. Could wasm-bindgen-test-runner be fixed instead?

weiznich:
Technically it is certainly possible to fix wasm-bindgen-test-runner instead. I opted for fixing this in nextest as the output format of cargo test and therefor any tool emulating it is unspecified. From my point of view then the consuming tool bears the load to handle such incompatibilities. Also this specific fix feels like it would make parsing that list a bit more robust to similar changes as well. 

If you are strictly opposed to accept this fix we will need to find a different solution.

sunshowers:
The required output format for custom test harnesses is documented here: https://nexte.st/docs/design/custom-test-harnesses/#manually-implementing-a-test-harness. Nextest does not guarantee that it works with every custom test harness, just the ones that implement this (relatively straightforward) protocol.

I believe the wasm-bindgen test runner has been updated to work with nextest in the past (#2372). I'm not strictly opposed to taking this change if there's a good reason to, but I'd like to understand why the test runner output changed in this fashion. I have generally negative feelings about Postel's law and would rather nextest not become a dumping ground for every test runner's weirdnesses.

weiznich:
For the record the current output of `wasm-bindgen-test-runner` is 

```
my-test-1: test
my-test-2: test

```

so it merely adds an additional empty line in the end, likely to simplify their implementation. 

I would go even as far as stating that this still satisfies the documented output format for nextest as that specification doesn't talk about line endings at all. Now you can certainly argue that your specification also doesn't show the new line, but it also doesn't show other things (like other test names, colons in test names, etc) that are accepted.

> I believe the wasm-bindgen test runner has been updated to work with nextest in the past (https://github.com/nextest-rs/nextest/issues/2372).

I can confirm that it did work up to wasm-bindgen-test-runner version 0.2.104

sunshowers:
Why would that simplify their implementation?

Look, I'm not interested in nitpicking here -- I'm happy to accept this change if there is a compelling reason to, and so far one has not been presented. Have you filed an issue against wasm-bindgen?

weiznich:
It would simplify their implementation because they now don't need to keep track of whether a specific line is the last one or not. They can then just use `println!()` or similar to emit the output. I expected that to be rather obvious, so I did not wrote this before.

I did not fill this as issue against wasm-bindgen because I believe there behavior is correct, as demonstrated by the fact that wasm-bindgen-test-runner still works without issues on it's own. If you believe that's not the case, feel free to raise the issue there.

sunshowers:
A series of printlns would result in what the default test runner does and not result in an extra empty line at the end.

weiznich:
I'm happy to accept you explanation on what else might case this. For reference the relevant code is here: https://github.com/wasm-bindgen/wasm-bindgen/blob/main/crates/cli/src/wasm_bindgen_test_runner.rs#L202-L206 Notably it did not change since in the last month, but only broke in the latest version.

sunshowers:
Looks like the change was clearly an unintentional regression, then, in which case I'd strongly recommend that this is fixed upstream. Please follow up there, maybe even by sending a PR to them. They might also choose to have CI validate that nextest works with the test runner and doesn't have such regressions in the future.

weiznich:
I still don't see what broke there beside the change in the output. It's always easy to say: "That's not my bug", but you did not provide any future details here.

Also as written before: The current output satisfies the documented format, so this is still a bug in nextest. Do you want me to fill another bug for this?

sunshowers:
> I still don't see what broke there beside the change in the output. It's always easy to say: "That's not my bug", but you did not provide any future details here.

The bug is that the custom test runner produces a line of output that does not end in `: test`.

No, the documented format is just fine as it is (though it is missing the support for `: benchmark` which I'll add)

weiznich:
The documentation states:

>The test harness MUST support being run with --list --format terse. This command MUST print to stdout all tests in exactly the format
>
>    my-test-1: test
>    my-test-2: test
>
>  Other output MUST NOT be written to stdout.
>
>  Custom test harnesses that are meant to be run as a single unit MUST produce just one line in the output.

Let's start from the end. The last requirement is satisfied by the output provided by wasm-bindgen-test-runner

It depends on how your read the requirement before if it is satisfied or not. If you take it literally as written there no the wasm-bindgen-test-runner output doesn't satisfy it, but at the same time no test runner (including libtest) satisfies that as noone literally puts out this text, but always replaces test names, etc.
If you don't require literally the test from the requirement the current output satisfies this requirement. So either the requirement is not specific enough or there is a bug in the implementation.

sunshowers:
Thanks, have resolved this.

Issue #387, opened 2022-07-22
Add a way for some profiles to inherit others


sunshowers:
Cargo profiles have the [`inherits`](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles) setting which lets some profiles inherit from others.

Nextest profiles should have the `inherits` setting as well. Steps to do:
* Add "inherits" to the `[profile]` data model.
* Check for cycles in the inherits graph (can consider using petgraph for this)
* Implement support for inherits -- in `NextestProfile`, store a chain of profiles rather than just one profile + default.

This is a really good first issue and will probably involve ~1 day of work -- please comment here if you'd like to work on it!

asder8215:
Hi, has this issue been taken or completed already? If not, is it possible if I could work on it? I might need a few days to understand the repo/codebase, but I did try out nextest a bit within uutils.

sunshowers:
Done in #2786. Thanks!

PR #2786, opened 2025-11-19
feat: nextest profile inheritance

asder8215:
This PR implements a way for nextest profiles to inherit from other profiles as mentioned in #387.

### DIFF
--- a/nextest-runner/src/config/elements/inherits.rs
+++ b/nextest-runner/src/config/elements/inherits.rs
@@ -310,11 +310,11 @@ mod tests {
         let tool_configs = [
             ToolConfigFile {
                 tool: "tool1".to_string(),
-                config_file: tool1_config.try_into().unwrap(),
+                config_file: tool1_config,
             },
             ToolConfigFile {
                 tool: "tool2".to_string(),
-                config_file: tool2_config.try_into().unwrap(),
+                config_file: tool2_config,
             },
         ];
 
@@ -380,11 +380,11 @@ mod tests {
         let tool_configs = [
             ToolConfigFile {
                 tool: "tool1".to_string(),
-                config_file: tool1_config.try_into().unwrap(),
+                config_file: tool1_config,
             },
             ToolConfigFile {
                 tool: "tool2".to_string(),
-                config_file: tool2_config.try_into().unwrap(),
+                config_file: tool2_config,
             },
         ];
 


asder8215:
Hi @sunshowers, whenever you are able to, could you take a look at this PR in introducing profile inheritance?

The inheritance cycle detection is a little bit verbose due to how isolated nodes without edges are considered as SCCs by `korasaju_scc()`. I'm up for any feedback on how to reduce or improve that code. 

I'll also add tests for profile inheritance (testing for cycles, nonexistent profile to inherit from, and whether config fields are appropriately filled in from an inherited profile) later today as well.

sunshowers:
Can you just check for any SCCs with 2 or more nodes? That's how I've done this in the past.

asder8215:
Isn't it possible for there to be a self referential SCC if the user sets the inherits to the current profile?

sunshowers:
Ah I think we should just reject that case separately, no good reason to even try to model it or support it.

asder8215:
I think I'm having a hard time finding out on how to do this in a neat and efficient manner. 

Earlier, I wanted to use `remove_node()` to get rid of all isolated nodes in the first graph, but I realized that removing a node shifts the index of all the other nodes in the graph (which repeatedly checking for isolated nodes would have been an O(N^2) issue). 

Instead of doing that, in the current implementation, I'm grabbing all nodes that are not isolated and inserting them into a new graph (which I'm not too sure if the way I'm doing this right now is the neatest way to do this). From reading through the docs in `petgraph` crate and searching up for examples, I wasn't sure if there was a better way to just keep the current graph and filter out isolated nodes since `Graph<N, E, Directed, u32>` doesn't have an `.iter()` method; it does have a `.into_nodes_edges()` method (which returns a  `(Vec<Node<N, Ix>>, Vec<Edge<E, Ix>>)`, but that itself means I'd be reconstructing the new graph again and in a clunky manner since `Node<N, Ix>` contains an `[EdgeIndex<Ix>; 2]` which I believe stores indices to the incoming and outgoing edges within `Vec<Edge<E, Ix>>`.

The one thing that comes up to my head is that I could do an earlier check here:
```rust
        // For each custom profile, we add a directed edge from the inherited node
        // to the current custom profile node
        for (profile_name, profile) in &self.other_profiles {
            if let Some(inherit_name) = &profile.inherits
                && let (Some(&from), Some(&to)) =
                    (profile_map.get(inherit_name), profile_map.get(profile_name))
            {
                profile_graph.add_edge(from, to, ());
            }
        }
```
to detect self referential SCCs and append it in a vec of cyclic inherited profile nodes and not insert it into the graph. This way, all non-isolated nodes are guaranteed to be in a chain of 2 or more nodes and I could create a new graph around this. 

Just so I'm understanding correctly, when you say reject the self referential SCC case separately, do you want it's own error message to be thrown or should it be bundled up with all other detected SCCs? My understanding from zohkcat's PR is that you would find it nice for all SCCs to be printed out. The current implementation here does that through using `kosaraju_scc()`, but do you still want this behavior? Not sure because you asked earlier if I could check for <ins>any</ins> SCCs with 2<= nodes.

I also had a couple of questions regarding error handling for `inherits`:
* One difference that I made from zokhcat's PR is to raise the inheritance cycle error within `deserialize_individual_config()` instead of unwrapping it within `make_profile()`. I was wondering if this is the appropriate place to check for inheritance cycle?
* Currently, `inherits` does not have a `ConfigParseErrorKind` for a nonexistent profile that it wants to inherit from. This is because there's already a `ProfileNotFound` error within `nextest-runner/src/errors.rs`, which is used and raised in `resolve_profile_chain(...)` (called in `make_profile()`).  I was wondering whether inheriting from a nonexistent profile should be detected as its own `ConfigParseErrorKind` and raised within `deserialize_individual_config()`?

How did you approach checking for SCCs with 2 or more nodes personally?

Edit: My apologies if I'm bombarding you with a lot of questions on this!

sunshowers:
I think this can be *way* simpler than what you're imagining. There are 4 properties which need to be verified:

1. The default profile cannot inherit from any other profile. (And in fact, any profiles under default- cannot inherit from any other profile.)
2. A profile cannot inherit from itself.
3. A profile cannot inherit from a profile that isn't defined.

1, 2, and 3 don't need the construction of a graph at all, so check for them at the beginning. Then, construct a _reduced_ graph with each node being a profile and each edge being an inherits relationship not rejected by any of the 1/2/3 checks above. (Include the implicit inherits relationship from each non-default profile to the default profile if `inherits` is not defined.) You can probably combine all 3 properties + graph construction into the same loop.

Then, the only remaining check is:

4. In this reduced graph, there are no SCCs with 2 or more nodes.

These 4 properties together should establish the overall invariant that the inheritance graph is well-formed (acyclic, and nothing points to unknown nodes.) You don't need to remove nodes or anything at any point, just break it up into graph and non-graph checks.

To report errors here I'd use an "error collector" or "diagnostic sink" pattern. Pass in a `&mut Vec<InheritError>` or something where `InheritError` is an enum with variants for all the properties, then push to that vector on seeing an error.

Hope this helps.

asder8215:
I appreciate the advice and I agree with your properties and error handling here (didn't think about the error collector pattern!). I have a couple of follow up questions to ask on this.
* What if we had an SCC where the end node in the inheritance chain is self referential (i.e. A -> B -> C -> C)? 
  * ~Formally, the SCC cycle is {A, B, C}, but the primary node that causes this cycle is C.~ 
  * Should we prioritize property 2 and report this as a self referential inheritance error? Or should we use property 4, look at it as an entire SCC, and report this as an inheritance chain cycle error? 
  * My opinion is doing the former, as constructing the reduced graph with self referential profiles removed will prevent cycles from being detected from this inheritance chain case.
* When reporting errors for property 4, should we report the whole SCC or the first profile that contributes to this cycle?
  * With using the first profile, the benefit is that the error message will be much shorter, where if you have a cycle like A -> B -> C -> A, all you're reporting is an inheritance chain cycle at prof A
  * With using the entire SCC, the error message will be bigger, but it gives the user a better idea on where the cycle could be occurring. For example, A -> B -> C -> D -> E -> C, the SCC that will be detected is ~{A, B, C, D, E}~ {C, D, E}, but if I were to return prof ~A~ C, it may not be super clear to the user where the cycle is occurring
    * Edit: Okay, another idea that came to my head is to report the beginning element and last element in the SCC to let user know about the cycle range.
 
Edit: Correction on SCC, I think I got it wrong. It should be {C} instead of {A, B, C} for the first and {C, D, E} instead of {A, B, C, D, E} for the second one, but still curious about how reporting errors for inheritance chain cycles should work.

sunshowers:
Can you just report the entire SCC? And definitely prioritize property 2.

asder8215:
@sunshowers, I modified the code to follow the properties we talked about earlier and added a couple of test cases. Let me know if there's anything I should change or if there are any other test cases I should add here!

nextest-runner/src/config/core/imp.rs
```diff
@@ -575,6 +577,9 @@ impl NextestConfig {
             );
         }
 
+        // Checks that the profiles correctly use the inherits setting.
```
sunshowers:
```suggestion
        // Check that the profiles correctly use the inherits setting.
```

We use simple present tense for comments (similar to the comments below).

nextest-runner/src/config/core/imp.rs
```diff
@@ -794,6 +799,13 @@ impl NextestConfig {
     fn make_profile(&self, name: &str) -> Result<EarlyProfile<'_>, ProfileNotFound> {
         let custom_profile = self.inner.get_profile(name)?;
 
+        // Resolves the inherited profile into a profile chain
```
sunshowers:
```suggestion
        // Resolve the inherited profile into a profile chain.
```

nextest-runner/src/config/core/imp.rs
```diff
@@ -794,6 +799,13 @@ impl NextestConfig {
...
+        let inheritance_chain = if custom_profile.is_some() {
+            self.inner.resolve_profile_chain(name)?
```
sunshowers: Can this ever legitimately fail?

nextest-runner/src/config/core/imp.rs
```diff
@@ -951,6 +967,30 @@ pub struct EvaluatableProfile<'cfg> {
...
+    };
+    ($eval_prof:ident.$field:ident.$ref_func:ident()) => {
+        $eval_prof
+            .custom_profile
+            .iter()
+            .chain($eval_prof.inheritance_chain.iter())
+            .find_map(|inherit_profile| inherit_profile.$field.$ref_func())
+            .unwrap_or(&$eval_prof.default_profile.$field)
+    };
+}
```
sunshowers, line 1031: Let's split these into two separate macros to be somewhat less magical.

nextest-runner/src/config/core/imp.rs
```diff
@@ -1154,6 +1176,151 @@ impl NextestConfigImpl {
             .iter()
             .map(|(key, value)| (key.as_str(), value))
     }
+
+    /// Resolves a profile with an inheritance chain recursively
```
sunshowers:
```suggestion
    /// Resolves a profile with an inheritance chain recursively.
```

nextest-runner/src/config/core/imp.rs
```diff
@@ -1154,6 +1176,151 @@ impl NextestConfigImpl {
...
+            // covers for other and any future default profiles reserved by Nextest
+            // (i.e. "default-miri")
+            else if let Ok(ok_default) = self.get_profile(default_profile)
+                && let Some(other_default) = ok_default
+                && other_default.inherits().is_some()
+            {
+                inherit_err_collector.push(InheritsError::DefaultProfileInheritance(
+                    default_profile.to_string(),
+                ));
+            }
```
sunshowers: Don't quite follow what's happening here -- we should prevent any profiles whose name starts with `default-` from specifying `inherits`.

nextest-runner/src/config/core/imp.rs
```diff
@@ -1154,6 +1176,151 @@ impl NextestConfigImpl {
...
+
+        // Iterates through all custom profiles within the config file and constructs
+        // a reduced graph of the inheritance chain(s) after handling nonexistent
+        // inheritance and self referential inheritance
+        for (name, custom_profile) in self.other_profiles() {
+            // Certain reserved default profile are in other_profiles (i.e. "default-miri")
+            // which should be ignored. Note that DEFAULT_PROFILES contains 2 strings, so this
+            // check is pretty much constant time, but if more reserved default profile were to
+            // be added, we should consider having a Hashmap impl to check if current profile name
+            // is a default profile
```
sunshowers: See above -- we should apply this check to any profiles that start with `default-`.

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+        Self(inherits)
+    }
+
+    /// Returns the profile that the custom profile inherits from
+    pub fn inherits_from(&self) -> Option<&str> {
+        self.0.as_deref()
+    }
+}
+
+// TODO: Need to write test cases  for this
```
sunshowers: Is this TODO still relevant?

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+pub struct Inherits(Option<String>);
+
+impl Inherits {
+    /// Creates a new `Inherits`.
+    pub fn new(inherits: Option<String>) -> Self {
+        Self(inherits)
+    }
+
+    /// Returns the profile that the custom profile inherits from
```
sunshowers:
```suggestion
    /// Returns the profile that the custom profile inherits from.
```

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+        max_fail: Option<MaxFail>,
+        slow_timeout: Option<SlowTimeout>,
+        global_timeout: Option<GlobalTimeout>,
+        leak_timeout: Option<LeakTimeout>,
+        overrides: Vec<DeserializedOverride>,
+        scripts: Vec<DeserializedProfileScriptConfig>,
+        junit: JunitImpl,
+        archive: Option<ArchiveConfig>,
+        inherits: Option<String>,
+    }
```
sunshowers, line 47: is this dead?

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    #[test_case(
```
sunshowers, line 49: good tests!

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+                let default_profile = default_profile.apply_build_platforms(&build_platforms());
+                let profile = config
+                    .profile(&custom_profile.name)
+                    .unwrap_or_else(|_| panic!("{} profile is known", &custom_profile.name));
+                let profile = profile.apply_build_platforms(&build_platforms());
+                assert_eq!(default_profile.inherits(), None);
+                assert_eq!(profile.inherits(), custom_profile.inherits.as_deref());
+                // This is a confirmation on a custom profile inheriting another custom
+                // profile's configs properly; however, this should cross check the all expected
+                // field with the inherited custom profile fields fully
```
sunshowers: I'd simplify this down to something like "Spot check that inheritance works correctly." I don't think we need to write exhaustive tests for every field.

nextest-runner/src/config/elements/inherits.rs
```diff
@@ -0,0 +1,238 @@
...
+                        let expected_err: HashSet<&InheritsError> =
+                            expected_inherits_err.iter().collect();
+                        for actual_err in inherits_err.iter() {
+                            match actual_err {
+                                InheritanceCycle(sccs) => {
+                                    // to check if the sccs exists in our expected errors,
+                                    // we must sort the SCC (since these SCC chain are also
+                                    // in a non-deterministic order). this runs under the
+                                    // assumption that our expected_err contains the SCC cycle
+                                    // in alphabetical sorting order as well
```
sunshowers: Hmm, I think we should sort the SCCs while constructing the error, not here, since the order is defined to be arbitrary. (We could provide a better error message here, outlining the cycle more clearly, but that's tricky to do without huge benefit.)

nextest-runner/src/config/elements/inherits.rs
asder8215: Forgot to remove this from a previous commit

nextest-runner/src/config/elements/inherits.rs
asder8215: Certain fields that are not used here are dead (i.e. default_filter, test_threads, etc.). I actually didn't know if I should create full thorough tests on all the config fields that could be inherited within the custom profile, so I was waiting for input on this.

nextest-runner/src/config/elements/inherits.rs
asder8215: Sounds good to me!

nextest-runner/src/config/elements/inherits.rs
asder8215: When you say sort the SCCs while constructing the error, are you referring to sorting the SCC vector in the `check_inheritance_cycles()` function within `nextest-runner/src/core/imp.rs`? 

The function currently displays the cycle for the error, but it isn't deterministic in the sense that if we had a cycle like this A -> B -> C -> A, the SCC cycle could be written as {A, B, C}, {C, A, B}, or {B, C, A}; it shows the cycle and connection of the nodes in the sense that A's prev is C and next is B, B's prev is A and next is C, and C's prev is B and next is A. (my bad if this wasn't explained properly in the comment)

nextest-runner/src/config/core/imp.rs
asder8215: I think I saw that `.get_profile()` return an Option<&CustomProfileImpl> and thought that if custom profile is none we shouldn't have an inheritance chain; otherwise if custom profile is some, then we should have an inheritance chain. 

On second look, I see that `.get_profile()` returns None if we're dealing with the "default" profile and Some otherwise (and error if profile is nonexistent). In that case, I think the `.is_some()` check is unnecessary, but just to be sure, is the `make_profile()` function used on the "default" profile?

nextest-runner/src/config/core/imp.rs
asder8215: Sounds good. I misunderstood the no inheritance on default profiles as regarding only the default profile names reserved in `DEFAULT_PROFILES` array.

I'll just do a sanity check on the `DefaultProfileImpl` here (which I believe works on the `default` profile) in this function, and then handle all the other default profiles  that starts with `default-` within `check_inheritance_cycles()`.

nextest-runner/src/config/elements/inherits.rs
asder8215: Let me know if you want me to remove this struct!

asder8215:
Took your suggestions and modularized the `check_inheritance_cycles()` (now called `custom_profile_inheritances()`; there's a separate  `check_inheritance_cycles()` function that fits the purpose of the name) into a few smaller functions for neatness.

I haven't sorted the SCC cycle vec in the new `check_inheritance_cycles()`-- not too sure how useful that would be since the SCCs are reported in cyclical order just can't deterministically show one cyclical order and I think it would be confusing to the user to not be able to visibly follow the cycle in the error. Let me know your thoughts on this!

sunshowers:
With apologies, I need to tackle something urgently at work this week. Hoping to look at this by the end of the week.

asder8215:
It's all good and no pressure or rush in checking this out.

sunshowers:
Looks good overall -- I've fixed a few issues (particularly inheritance order being incorrect) and pushed them. Going to land once this passes CI and once I've had a final look.

nextest-runner/src/config/elements/inherits.rs
sunshowers: Ended up shortening this to just the fields being tested in the test below.

asder8215:
Okay sweet, it's nice to see that this got merged in. Thank you for the opportunity on letting me work on this issue!

PR #2927, opened 2026-01-13
iTerm supports OSC 9;4

case:
Similar to https://github.com/nextest-rs/nextest/pull/2724

> it would be really nice if there was an universally established protocol to query supported emulator features, too late for that

Yup. :/


sunshowers:
Thanks. You may want to send the same fix over to Cargo: https://github.com/rust-lang/cargo/blob/6451eb8ea679514fd45c28a0dff1a53142f6e0a9/src/cargo/core/shell.rs#L638-L645

Issue #978, opened 2023-09-26
Tracking feature for setup-scripts


sunshowers:
# Open questions

- [ ] How do we support build archive/reuse with Cargo commands?
- [ ] Is the current format for environment variables (KEY=value) okay? Do we need to check that KEY/value are unique identifiers?
- [ ] Should we have a special syntax for running Cargo binaries?

tisonkun:
This setup script seems to be executed before test. How can I do if I'd set up a database and teardown after the test suite finished?

For non-blocking start, I may use `&` to keep the database running in the background. But it seems missing a teardown script hook.

sunshowers:
@tisonkun correct, there is no support for teardown scripts yet (at work, we use setup scripts to pre-seed and write out a database as a tarball to disk, which tests can then pick up). But it's something that would definitely be nice to have.

tlater-famedly:
It would be nice if there was a way to either access `$NEXTEST_PROFILE` in the setup scripts, *or* override the setup scripts from the default profile.

Currently, it's difficult to set up a different set of things in my CI profile, which is annoying as I'm trying to set up docker containers which have to use different volume paths on CI hosts. I can rely on the fact that the scripts seem to be executed in order, at least, but it feels hacky to rely on that and is hard to express in code.

It would be much nicer if I could change script behavior slightly depending on profile, or just execute different scripts in different profiles altogether.

sunshowers:
@tlater-famedly Yeah that's a good idea, I think we should uniformly set NEXTEST_PROFILE in test and script environments.

sunshowers:
@tlater-famedly this is now set in nextest 0.9.89, which was just released.

sunshowers:
As of nextest 0.9.98, [setup scripts](https://nexte.st/docs/configuration/setup-scripts) now use the `[scripts.setup]` configuration. The location of the configuration has been changed to support the new [wrapper scripts](https://nexte.st/docs/configuration/wrapper-scripts/).

The old `[script.*]` location will work for a short while, but will be removed in the near future. (This is not a breaking change because setup scripts aren't stable yet... though hopefully soon!)

metatoaster:
Would the ability to provide environment variables specific to a given script be a welcomed addition?  As an example, given some `cargo` tool (or any other script/binary), it might accept specific environment variables to alter how it behaves or what it does, and naturally it would be nice to be able to define it in the setup script definition. As an example, the `.config/nextest.toml` could be defined like:

```toml
[...]

PR #3001, opened 2026-01-26
Allow setup scripts to define specific environment variables

metatoaster:
As per [the comment](https://github.com/nextest-rs/nextest/issues/978#issuecomment-3738460322) I made on #978, I've gone ahead and implement exactly that, a way to set specific environment variables for setup scripts.

During the construction of the test for this, I've discovered a possibly undocumented silent failure (which got me down some rabbit holes, but I digress). if the setup script writes the `NEXTEST` prefixed variable to `NEXTEST_ENV`, it seems to cause all environment variables assigned to that file be ignored and thus not passed to the intended tests. I'd expect that it would rather only ignore values keyed under `NEXTEST*` and thus the test will see all other values under valid keys.  I haven't dived too deep into the code for that part but maybe this could also be documented with the setup scripts if this is intended.


nextest-runner/src/config/scripts/imp.rs
```diff
@@ -1072,6 +1085,12 @@ mod tests {
             leak-timeout = "1s"
             capture-stdout = true
             capture-stderr = true
+
+            [scripts.setup.qux]
+            command = "qux"
+            env = {
+                MODE = "qux_mode",
+            }
```
sunshowers: `env` should be part of `command` (always requiring passing in as a map)

fixtures/nextest-tests/.config/nextest.toml
```diff
@@ -26,6 +26,16 @@ filter = "test(=test_cargo_env_vars)"
...
+[[profile.default.scripts]]
+platform = { host = "cfg(unix)" }
+filter = "test(=test_setup_script_env_vars)"
+setup = "my-script-unix-with-env"
+
+[[profile.default.scripts]]
+platform = { host = "cfg(windows)" }
+filter = "test(=test_setup_script_env_vars)"
+setup = "my-script-windows-with-env"
+
```
sunshowers, line 38: Do we need an extra test? Seems like we can combine this check into test_cargo_env_vars.

fixtures/nextest-tests/.config/nextest.toml
```diff
@@ -185,6 +195,22 @@ command = { command-line = "my-script.sh", relative-to = "workspace-root" }
 # "scripts\\my-script.bat" ensures this.
 command = 'cmd /c "scripts\\my-script.bat"'
 
+[scripts.setup.my-script-unix-with-env]
+command = { command-line = "my-script.sh", relative-to = "workspace-root" }
+env = {
+    NEXTEST = "0",
+    NEXTEST_PROFILE = "0",
```
sunshowers: I think these should error out similar to the environment that setup scripts set.

sunshowers (commented):
Thanks for doing this!

> if the setup script writes the NEXTEST prefixed variable to NEXTEST_ENV, it seems to cause all environment variables assigned to that file be ignored and thus not passed to the intended tests. I'd expect that it would rather only ignore values keyed under NEXTEST* and thus the test will see all other values under valid keys.

Interesting -- I think this should just error out loudly instead (this is an experimental feature so it's fine to change it). I'd collect all env vars that are bad and explicitly error out on them.

metatoaster:
Thanks for the feedback on the pull request. I will be working on the suggested changes in about a weeks time once my partner and our newborn is settled back at home hopefully soon. The delay in response is due to this evolving personal circumstance, but please know that I do want to get this change in so I may get rid of the workaround using platform specific scripts for the other project where I'm a contributor of.

sunshowers:
Congratulations! Hope everyone's well 🩷

nextest-runner/src/config/scripts/imp.rs
metatoaster: Yep, and this makes a lot of sense given how `command` is effectively shared with wrapper scripts.  Naturally having wrapper scripts support environment variables will take some additional work - should that be included with this pull request?

In any case I've pushed a commit (without flattening onto the original) to reflect this - please let me know if flattening this new commit onto the existing one is required.

fixtures/nextest-tests/.config/nextest.toml
metatoaster: I had this extra definition and separate test so that I don't touch the existing definition that is working without environment variables defined, as the idea of this change is to be able to specify different environment variables for the same underlying script (e.g. `my-script.sh`), hence why I ultimately split this out.

That said It should be possible to combine it into `test_cargo_env_vars` if this is what's desired. Please let me know if this is the case.

fixtures/nextest-tests/.config/nextest.toml
metatoaster: By error out, should this be done at the point of parsing?

metatoaster:
> I think this should just error out loudly instead (this is an experimental feature so it's fine to change it). I'd collect all env vars that are bad and explicitly error out on them.

I had a bit more time to chase this down (between newborn needs) and the following are just notes and observations that I've made. The current behavior of `parse_env_file` simply [returns an error](https://github.com/nextest-rs/nextest/blob/nextest-runner-0.107.0/nextest-runner/src/runner/script_helpers.rs#L44-L49) upon encountering environment variables that start with `NEXTEST`, and [in the executor](https://github.com/nextest-rs/nextest/blob/nextest-runner-0.107.0/nextest-runner/src/runner/executor.rs#L632-L638) this error is stored, where eventually it will [make its way to an `InfoResponse<'a>`](https://github.com/nextest-rs/nextest/blob/nextest-runner-0.107.0/nextest-runner/src/runner/executor.rs#L1285-L1299), though I am unsure how that bit gets to where it gets reported.

I also noted that when the setup script is finished it gets reported when the display reporter handles a [`TestEventKind::SetupScriptFinished`](https://github.com/nextest-rs/nextest/blob/nextest-runner-0.107.0/nextest-runner/src/reporter/displayer/imp.rs#L720-L741). The `run_status` value is a `SetupScriptExecuteStatus` which contains the attribute `error_summary`, but it is unused at this part.  I lazily used `dbg!` to reveal the following:

```
           SETUP [      1/2] my-script-unix: /tmp/nextest-fixture-2HiJEa/src/my-script.sh
    [nextest-runner/src/reporter/displayer/imp.rs:728:17] &run_status.error_summary = None
      SETUP PASS [   0.022s] my-script-unix: /tmp/nextest-fixture-2HiJEa/src/my-script.sh
           SETUP [      2/2] my-script-unix-with-env: /tmp/nextest-fixture-2HiJEa/src/my-script.sh
    [nextest-runner/src/reporter/displayer/imp.rs:728:17] &run_status.error_summary = Some(
        ErrorSummary {
            short_message: "error reading setup script output",
            description: "error reading setup script output\n  caused by:\n  - key `NEXTEST` begins with `NEXTEST`, which is reserved for internal use",
        },
    )
      SETUP PASS [   0.012s] my-script-unix-with-env: /tmp/nextest-fixture-2HiJEa/src/my-script.sh
```

In both options for  `error_summary` they simply generate `SETUP PASS` with no indication that one of them had an error. Perhaps if either the short message or the description is written out it would serve to indicate a swallowed error.

So by "explicitly error out on them" (when a reserved key is set), should the behavior of this setup script be of an failure and trigger early abort the test run, much like how a non-zero exit code does? Or take a more lenient approach by collecting the env keys that are reserved and then ensure that they all show up under `error_summary` and then update the report for `TestEventKind::SetupScriptFinished` to include that?  I can take a stab at doing either implementation if this is desired.

sunshowers:
Sorry for the delay, work has kept me very busy.

> So by "explicitly error out on them" (when a reserved key is set), should the behavior of this setup script be of an failure and trigger early abort the test run, much like how a non-zero exit code does

Yeah, that is definitely what we want here -- fail loudly. Working on fixing this right now -- thanks for catching!

edit: fixed in https://github.com/nextest-rs/nextest/pull/3094.

metatoaster:
Thanks for the fix! Also the pull request you just did is also quite instructive with regards to project organization. I will tidy up this pull request in due time, and let me know if this plan works for you (to address the two remaining issues raised during the code review):

- Have the config parser error out on all `NEXTEST` reserved values that are present inside `command.env`
- Combine the new test I added into `test_cargo_env_vars` (and move the standalone tests I intend to add from `fixtures/nextest-tests/tests/basic.rs` into `integration-tests/tests/integration/main.rs`).
- Update the docs to reflect the reserved keys.

sunshowers:
Sounds great!

metatoaster:
I ended up squashing all the code commits into one, given the intrusive nature of the snapshot modifications from that one test that got added then removed (conflicting with the current main branch, the merge target) and that some of the later commits ended up being rather trivial (e.g. adding tests). Should the original commits are of interests, I [pushed a separate branch](https://github.com/metatoaster/nextest/commits/setup-scripts-env-old/) with most of the individual commits based on the earlier commit that I was working on top of. Sorry for the wait, the needs of a newborn are many after all.

fixtures/nextest-tests/tests/basic.rs
```diff
@@ -326,6 +326,14 @@ fn test_cargo_env_vars() {
...
 
     assert_eq!(std::env::var("MY_ENV_VAR").as_deref(), Ok("my-env-var"));
+    // The command for the script has received the value
+    assert_eq!(std::env::var("SCRIPT_CMD_ENV_VAR").as_deref(), Ok("set-in-conf"));
+    // While the command may be started with or without a dummy value, which is also checked
+    // here for completeness.
+    assert!(matches!(
+        std::env::var("CMD_ENV_VAR").as_deref(),
+        Err(&std::env::VarError::NotPresent) | Ok("not-set-in-conf"),
+    ));
```
sunshowers: Is this ever not present? If so, could this be written as an exhaustive match? If not, then just assert on the specific value.

nextest-runner/src/config/scripts/imp.rs
```diff
@@ -881,6 +890,7 @@ impl<'de> Deserialize<'de> for ScriptCommand {
                 Ok(ScriptCommand {
                     program,
                     args,
+                    env: BTreeMap::new(),
                     relative_to: ScriptCommandRelativeTo::None,
```
sunshowers, line 892: Ah so I realized that `ScriptCommand` is used for both setup and wrapper scripts, but we only apply `env` to setup scripts. Could you also extend `env` to apply to wrapper scripts (and add a test)?

nextest-runner/src/config/scripts/imp.rs
```diff
@@ -314,6 +314,11 @@ impl SetupScriptCommand {
         // `CARGO_*` and `NEXTEST_*` variables set directly on `cmd` below.
         test_list.cargo_env().apply_env(&mut cmd);
 
+        // Set the additional user-provided environment variables assigned to the setup
+        // script configuration after the global values assigned above but before the
+        // test runner controlled ones which are assigned below, as per above note.
+        cmd.envs(config.command.env.iter());
```
sunshowers: To ensure that current and future keys are covered, can we use an exhaustive destructure here (and in wrapper scripts)?

nextest-runner/src/config/scripts/imp.rs
```diff
@@ -984,6 +1003,46 @@ impl<'de> serde::de::DeserializeSeed<'de> for CommandInnerSeed {
...
+            where
+                A: serde::de::MapAccess<'de>,
+            {
+                let mut env = BTreeMap::new();
+                while let Some((key, value)) = map.next_entry::<String, String>()? {
+                    if key.starts_with("NEXTEST") {
+                        return Err(A::Error::invalid_value(
+                            serde::de::Unexpected::Str(&key),
+                            &"a key that does not begin with `NEXTEST`, which is reserved for internal use",
+                        ));
```
sunshowers: I feel like we should apply some more checks to the environment key here (e.g. not supporting cases like `' '`) though we have this issue elsewhere so that can be a followup.

fixtures/nextest-tests/tests/basic.rs
```diff
@@ -326,6 +326,14 @@ fn test_cargo_env_vars() {
     );
 
     assert_eq!(std::env::var("MY_ENV_VAR").as_deref(), Ok("my-env-var"));
+    // The command for the script has received the value
```
sunshowers: Please add a period to the end of this comment.

nextest-runner/src/config/scripts/imp.rs
```diff
@@ -984,6 +1003,46 @@ impl<'de> serde::de::DeserializeSeed<'de> for CommandInnerSeed {
...
+                    }
+                    env.insert(key, value);
+                }
+                Ok(env)
+            }
+        }
+
+        deserializer.deserialize_any(EnvMapVisitor)
```
sunshowers: Can this be `deserialize_map`?

nextest-runner/src/config/scripts/imp.rs
```diff
@@ -314,6 +314,11 @@ impl SetupScriptCommand {
         // `CARGO_*` and `NEXTEST_*` variables set directly on `cmd` below.
```
sunshowers: I believe this comment may not be true any more -- I think `test_list.cargo_env().apply_env` should move below `cmd.envs`.

fixtures/nextest-tests/tests/basic.rs
metatoaster: I ended up resorting to an additional environment variable specific to `test_setup_script_defined_env` for the disambiguation of this particular test, which is also part of the fix that addresses the location of `test_list.cargo_env()`, which is done in 0d00ce9f.

nextest-runner/src/config/scripts/imp.rs
metatoaster: Yeah I was thinking of doing this in a separate pull request, but we can do it here in one go also. It is now done in 68919b38.

nextest-runner/src/config/scripts/imp.rs
metatoaster: Sorry I am not following this particular comment, as the intention is to assign all arbitrary keys and values in this map to the command, the keys and values here should have been validated by this point given the deserializer provided that. I thought assignment of all values with an iterator is suitable, so I am not sure what is meant by an exhaustive destructure on an arbitrary map here.

Edit: Unless of course you mean I should use `cmd.env` on each of the key/value pairs obtained by destructuring every single entry inside `config.command.env` rather than simply using `cmd.envs`, though if we are going down this far, there should probably be a newtype for this `BTreeMap` and implement `fn apply_env(cmd: &mut Command)` for that with the additional filtering/validation logic if this is what you are truly after, though at this point that particular map, if deserialized through serde, should have the unexpected values filtered out.

That said, should we have this newtype, the needs from [this comment](https://github.com/nextest-rs/nextest/pull/3001#discussion_r2907594583) can simply be addressed.

fixtures/nextest-tests/tests/basic.rs
metatoaster: Not sure how I managed to delete the rest of that sentence without noticing it myself, but this is now addressed with the fix to the ordering of `test_list.cargo_env().apply_env` in 0d00ce9f.

nextest-runner/src/config/scripts/imp.rs
metatoaster: Done in 55ac2d62.

nextest-runner/src/config/scripts/imp.rs
metatoaster: Yeah, I should have investigated what `test_list.cargo_env()` does in greater detail, indeed it does not inhibit how I want to apply `command.env`. Swapping it around as suggested does not change the intended behavior where `command.env` overrides standard environment variables, and on top of that, doing so will allow `force = true` to function as intended to override values specified in `command.env`. I've updated the code and extended the test case to account for this which is done in 0d00ce9f.

nextest-runner/src/config/scripts/imp.rs
metatoaster: I ended up started working on this too as part of the above comment about environment keys, since I found relevant info about environment variables in POSIX.1-2024.  Will push new commit hopefully in the next couple days.

metatoaster: I added 6fba33ba and c0f52734 to address the need to do this check. Refer to the lengthy comment before `ENV_KEY_VALIDATION` which was added in 6fba33ba for the background I've included with why the restriction and checks are in place. In c0f52734 I added the minimal viable implementation to propagate the new error returned from the validation back up to where they may be consumed, so the level of detail captured might be lower than expected. Though given how the validation should have been done during deserialization, the code paths that can result in invalid environment variables keys triggering the error might not actually be hit under usage through the CLI/configuration files.  Anyway I am looking forward to what you think, and please let me know if and what else I should improve.

sunshowers:
Looks great, thanks. I ran into the same issue with hyphens in env vars a while ago: see https://nexte.st/docs/configuration/env-vars/#environment-variables-nextest-sets (`NEXTEST_BIN_EXE_<name>`)

Will do a final review and then get this in for release next week.

metatoaster:
Thank you for your time for the code review and merging this in!

PR #3074, opened 2026-02-15
Updated zip dependency to version 8.0.0.

koutheir:
This fixes the following warning when `cargo install`:
```plain
warning: package `zip v7.3.0` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
```

Closes #3073

PR #3083, opened 2026-02-19
fix  version 7.3.0 is yanked    version 7.4.0 is yanked

Utishnik:


sunshowers:
Addressed in https://github.com/nextest-rs/nextest/pull/3074. Thanks for putting this up!

Issue #3128, opened 2026-03-09
optimize `cargo nextest list --list-type binaries-only` to avoid unnecessary setup work


cornmander:
`cargo nextest list --list-type binaries-only` appears to do a substantial amount of work that is not needed for its current behavior.

Today, the `binaries-only` path still eagerly:
- loads config and profile,
- parses filtersets and builds a `TestFilter`,
- initializes shared base state that includes Cargo metadata / `PackageGraph` setup, and
- in some cases resolves build platform metadata that is only needed for later phases.

However, the `binaries-only` branch ultimately just writes the `BinaryList` and does not use the profile or test filter.

Relevant code paths:
- `cargo-nextest/src/dispatch/core/list.rs`
- `cargo-nextest/src/dispatch/core/base.rs`

Measured on the nextest workspace itself, warm cache, macOS:
- `cargo nextest list --list-type binaries-only`: ~473 ms
- `cargo nextest list --list-type binaries-only --cargo-metadata ...`: ~330 ms
- `cargo nextest list --list-type binaries-only --binaries-metadata ...`: ~194 ms
- `cargo nextest list --list-type binaries-only --cargo-metadata ... --binaries-metadata ...`: ~52 ms

This suggests there is a meaningful win available from making the `binaries-only` path lazier:
- branch earlier in `exec_list`,
- skip config/profile/filterset construction for `binaries-only` while preserving current semantics, and
- defer `cargo metadata` / `PackageGraph` creation when reused `binaries-metadata` is already sufficient.

One open question is whether `binaries-only` should continue to ignore filtersets and profile-driven config. This issue is about preserving current behavior and avoiding dead work, not changing semantics.

Related:
- #229
- #154

PR #3130, opened 2026-03-09
optimize reused binaries-only list path

cornmander:
Refs #3128.

## summary

- add a dispatch-level fast path for `cargo nextest list --list-type binaries-only` when `--binaries-metadata` is reused
- skip `BaseApp::new`, `cargo metadata`, and `PackageGraph` setup for that narrow reused-binaries case
- keep the fast path conservative by falling back to the existing implementation if filtersets, explicit profiles, or explicit config files are in play
- factor the version-only config and workspace-root loading needed by both paths into shared dispatch helpers
- add integration coverage for `list --list-type binaries-only --binaries-metadata ...`

## testing

- `cargo xfmt`
- `cargo check -p cargo-nextest -p integration-tests`
- `cargo nextest run -p cargo-nextest`
- `cargo nextest run -p integration-tests -E 'test(test_list_binaries_only) or test(test_list_binaries_only_after_build) or test(test_list_full_after_build) or test(test_list_host_after_build) or test(test_list_target_after_build)'

## performance

- I did not find an existing benchmark-style test harness for nextest itself in this repo, so these numbers are from local `hyperfine` runs.
- On the locally built binary, the optimized path measured: `./target/debug/cargo-nextest nextest list --list-type binaries-only --binaries-metadata /tmp/nextest-binaries-metadata.json --message-format json >/dev/null` -> `56.8 ms ± 13.1 ms`.
- To isolate the savings on the same branch, I forced the fallback path by passing an explicit profile: `... --profile default ...` -> `357.2 ms ± 63.6 ms`.
- I also forced the fallback path with an explicit config file: `... --config-file /Users/gshikhman/nextest/.config/nextest.toml ...` -> `293.5 ms ± 40.5 ms`.
- Before this change, my earlier workspace measurement for `cargo nextest list --list-type binaries-only --binaries-metadata /tmp/nextest-binaries-metadata.json` was about `194 ms`.

### DIFF
--- a/cargo-nextest/src/dispatch/app.rs
+++ b/cargo-nextest/src/dispatch/app.rs
@@ -114,6 +114,16 @@ impl AppOpts {
     ) -> Result<i32> {
         match self.command {
             Command::List(list_opts) => {
+                if list_opts.try_exec_binaries_only_fast(
+                    &early_args,
+                    self.common.manifest_path.as_deref(),
+                    &self.common.config_opts,
+                    output,
+                    output_writer,
+                )? {
+                    return Ok(0);
+                }
+
                 let base = BaseApp::new(
                     output,
                     early_args,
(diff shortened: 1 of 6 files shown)


cornmander:
Closing this draft. The optimization is too narrow for the added control-flow complexity; I plan to revisit this by making BaseApp lazier instead.

PR #3132, opened 2026-03-09
make graph data lazy for reused binary lists

### LINKED ISSUE #3128: optimize `cargo nextest list --list-type binaries-only` to avoid unnecessary setup work

cornmander:
fixes #3128.

## summary

This change delays loading the Cargo package graph until it is actually needed, instead of constructing it in `BaseApp::new` unconditionally.

For `cargo nextest list --list-type binaries-only` with reused `--binaries-metadata`, that means nextest can skip `cargo metadata` and the `PackageGraph` in the common fast path, resolve the workspace root with `cargo locate-project`, and still preserve the existing diagnostics for CLI filter parsing.

The normal non-reused list path is not materially changed by this patch.

## performance

Benchmarks were run with warm caches on macOS (`aarch64-apple-darwin`) using `hyperfine`, comparing `origin/main` with this branch.

- `tokio` (`e67cd87`, 264 binaries): `121.6 ms ± 2.3 ms` -> `52.8 ms ± 0.8 ms` for `nextest list --list-type binaries-only --binaries-metadata ... --message-format json`.
- `tokio`: `2.30x` faster, or about `56.5%` lower wall time.
- `rust-analyzer` (`51966da`, 68 binaries): `174.7 ms ± 9.6 ms` -> `57.3 ms ± 1.5 ms` for the same command.
- `rust-analyzer`: `3.05x` faster, or about `67.2%` lower wall time.

The avoided Cargo startup cost lines up with the improvement:

- `tokio`: `cargo metadata --format-version=1 --all-features --filter-platform aarch64-apple-darwin` was `87.2 ms ± 3.5 ms`, while `cargo locate-project --workspace --message-format plain` was `39.6 ms ± 0.7 ms`.
- `rust-analyzer`: the same commands were `122.5 ms ± 5.7 ms` and `45.5 ms ± 2.4 ms`.

## testing

- `cargo xfmt --check`
- `cargo check -p cargo-nextest -p integration-tests`
- `cargo nextest run -p cargo-nextest`
- `cargo nextest run -p integration-tests -E 'test(test_list_binaries_only) or test(test_list_binaries_only_after_build) or test(test_list_full_after_build) or test(test_list_host_after_build) or test(test_list_target_after_build)'`

### DIFF
--- a/cargo-nextest/src/dispatch/core/base.rs
+++ b/cargo-nextest/src/dispatch/core/base.rs
@@ -533,16 +533,16 @@ impl BaseApp {
     }
 
     fn manifest_path_for_build(&self) -> Result<Option<Utf8PathBuf>> {
-        if let Some(graph_data) = self.graph_data.get() {
-            if self.reuse_build.cargo_metadata().is_some() {
-                return Ok(Some(
-                    graph_data
-                        .package_graph
-                        .workspace()
-                        .root()
-                        .join("Cargo.toml"),
-                ));
-            }
+        if let Some(graph_data) = self.graph_data.get()
+            && self.reuse_build.cargo_metadata().is_some()
+        {
+            return Ok(Some(
+                graph_data
+                    .package_graph
+                    .workspace()
+                    .root()
+                    .join("Cargo.toml"),
+            ));
         }
 
         if let Some(cargo_metadata) = self.reuse_build.cargo_metadata() {


sunshowers:
Is this something you'd still like to pursue?

PR #3190, opened 2026-03-19
CI: Add self hosted riscv64 runners

yuzibo:
This is one attempt to add self hosted riscv64 runner for CI job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,9 @@ jobs:
   build:
     name: Build and test
     runs-on: ${{ matrix.os }}
+    # NOTE: self-hosted riscv64 runners are experimental and may be flaky.
+    # Do not block CI on failures from this platform for now at inital stages of support.
+    continue-on-error: ${{ contains(matrix.os, 'self-hosted') }}
     strategy:
       matrix:
         os:
@@ -102,6 +105,7 @@ jobs:
           - ubuntu-24.04-arm
           - macos-latest
           - windows-latest
+          - [self-hosted, linux, riscv64]
         # 1.91 is the MSRV
         rust-version: ["1.91", stable]
       fail-fast: false


yuzibo:
Okay. I have to admit we have one test case that failed due to timeout I think. see [here](https://github.com/yuzibo/nextest/actions/runs/23281127313/job/67713305808).

Or could you help me to confirm this also?

sunshowers:
@yuzibo good catch -- fixed in https://github.com/nextest-rs/nextest/pull/3197.

sunshowers:
@yuzibo do I have to do anything on my end to configure self-hosted runners? (Probably, right?)

yuzibo:
> @yuzibo do I have to do anything on my end to configure self-hosted runners? (Probably, right?)

I think no? If zizmor job has been added to nextest's CI workflow as Trevor Gross suggests.

And I am waiting for the result from the job :) I forgot to push the latest code to my branch.

Once done, I need the token from your side to register the riscv64 runner. Please to read this [link](https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/add-runners) then Then we can agree on a time to pass the token which will be expired in one hour.

yuzibo:
hmm, test [failed](https://github.com/yuzibo/nextest/actions/runs/23328067098/job/67853458609) again.

let me to check it again

.github/workflows/ci.yml
```diff
@@ -95,13 +95,17 @@ jobs:
   build:
     name: Build and test
     runs-on: ${{ matrix.os }}
+    # NOTE: self-hosted riscv64 runners are experimental and may be flaky.
+    # Do not block CI on failures from this platform for now at inital stages of support.
+    continue-on-error: ${{ contains(matrix.os, 'self-hosted') }}
```
yuzibo, line 100: I think we can keep it running for a while then to decide if enable it nor not

sunshowers:
> Once done, I need the token from your side to register the riscv64 runner. Please to read this [link](https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/add-runners) then Then we can agree on a time to pass the token which will be expired in one hour.

Thanks. I've read the documentation. What time of day works best for you? I can do 17:00-21:00 Pacific Sunday March 22 or Monday March 23 (00:00-04:00 Monday March 23/Tuesday March 24 in UTC). After that my next availability is the same time, Sunday March 29.

yuzibo:
> Thanks. I've read the documentation. What time of day works best for you? I can do 17:00-21:00 Pacific Sunday March 22 or Monday March 23 (00:00-04:00 Monday March 23/Tuesday March 24 in UTC). After that my next availability is the same time, Sunday March 29.

All times sound good to me. So we can do it at 00:00-04:00 Monday March 23 in UTC. Is it okay for you? Or let's sync this on zulipchat IM again.

yuzibo:
This is one job already in [there](https://github.com/nextest-rs/nextest/commit/8c474726eacd5badeb2df66204dfa063219aea90), so I think we can close the PR now

yuzibo:
@sunshowers 

How about we enable riscv64 CI for [pull requests](https://github.com/nextest-rs/nextest/blob/main/.github/workflows/ci-riscv64.yml#L6) also? Now we use two runners for this CI and the freq of pushing to main is very limited, so I think we could give it a try.:)

sunshowers:
> How about we enable riscv64 CI for [pull requests](https://github.com/nextest-rs/nextest/blob/main/.github/workflows/ci-riscv64.yml?rgh-link-date=2026-03-26T02%3A22%3A19Z#L6) also?

Thanks for the suggestion. At the moment I'd like to avoid running riscv64 CI on PRs just to avoid backing up on queues, since there are just two runners available.

PR #3314, opened 2026-04-29
feat: add JSON Schema

ya7010:
Support: https://github.com/nextest-rs/nextest/discussions/3315

I have created a JSON Schema using [schemars](https://graham.cool/schemars/) so that `nextest.toml` can be validated with [tombi](https://github.com/tombi-toml/tombi) or [taplo](https://taplo.tamasfe.dev/lib/rust/taplo.html).


sunshowers:
Thanks for doing this!

Have you tested this end-to-end? I tried doing this a while ago and ran into https://github.com/tamasfe/taplo/pull/779 which never got accepted upstream.

ya7010:
I was testing with `Tombi`, but when I tried it with `Taplo`, an error occurred.

Taplo is currently unmaintained, so it is unlikely that it will be fixed.


`nextest.toml` in Tombi:

image

Test Code:

```toml
#:schema https://raw.githubusercontent.com/ya7010/nextest/3118e9a40e73428151c7a6f97a0b90fee464aba3/jsonschemas/nextest.json

# Define serial test group for running tests sequentially.
[test-groups]
serial = { max-threads = 1 }

# Run ty file watching tests sequentially to avoid race conditions.
[[profile.default.overrides]]
filter = "binary(file_watching)"
test-group = "serial"

[profile.ci]
# Print out output for failing tests as soon as they fail, and also at the end
# of the run (for easy scrollability).
failure-output = "immediate-final"
# Do not cancel the test run on the first failure.
fail-fast = false

status-level = "skip"

# Mark tests that take longer than 1s as slow.
# Terminate after 60s as a stop-gap measure to terminate on deadlock.
slow-timeout = { period = "1s", terminate-after = 60 }

# Show slow jobs in the final summary
final-status-level = "slow"
```

sunshowers:
Ahh, thanks, didn't realize you were the maintainer of Tombi! I'll have a look at this in more detail soon.

ya7010:
If there are any improvements needed on the Tombi side, I’ll take care of them.

jsonschemas/nextest.json
```diff
@@ -0,0 +1,1296 @@
...
+      "additionalProperties": {
+        "$ref": "#/$defs/TestGroupConfig"
+      }
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "store",
+    "profile"
+  ],
```
sunshowers: I started reviewing this and trying it out with tombi, but immediately noticed that we have these properties as required. In almost no cases are properties required -- an entirely empty config file is valid, for example, and is equivalent to not having a config file at all.

I think there's a difference here between the schema for the default config (where most things are required) and for custom repo-specific configs (where most things are optional). Could you look into this deeper?

I'll let you address this before continuing the review.

ya7010: I was made aware that there are exceptions that cannot be automatically generated from Rust structs.

I have modified `NextestConfigDeserialize` and `StoreConfigImpl` to use exceptions and implemented a special generation method for them.

https://github.com/nextest-rs/nextest/pull/3314/changes/858ea7dca9a48949352487cf978a70c278fe22d1

ya7010: @sunshowers 

I have finished fixing the lint errors in nextest.toml for this repository. Could you please review it?

.config/tombi.toml
```diff
@@ -0,0 +1,6 @@
+[lsp]
+formatting.enabled = false
```
ya7010, line 2: The formatting has been disabled as it is not the focus of this PR.

With this configuration, you can verify how the nextet.json schema functions in tombi.

jsonschemas/nextest.json
sunshowers: Thanks. It's going to take me a little while because I have two very high priority things to work on this week.

ya7010: It is not urgent. Thank you!

nextest-runner/src/config/elements/archive.rs
```diff
@@ -21,14 +27,24 @@ pub struct ArchiveConfig {
 /// This is `deny_unknown_fields` because if we take additional arguments in the future, they're
 /// likely to change semantics in an incompatible way.
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "config-schema", schemars(deny_unknown_fields))]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
```
sunshowers: You can rely on schemars's serde attribute inheritance here, I think.

nextest-runner/src/config/elements/archive.rs
```diff
@@ -233,6 +252,22 @@ impl<'de> Deserialize<'de> for RecursionDepth {
...
+        "RecursionDepth".into()
+    }
+
+    fn json_schema(_generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
+        schemars::json_schema!({
+            "oneOf": [
+                { "type": "integer", "minimum": 0 },
+                { "type": "string", "enum": ["infinite"] }
+            ]
+        })
```
sunshowers, line 271: This is great.

Cargo.toml
```diff
@@ -126,6 +126,7 @@ serde = { version = "1.0.228", features = ["derive"] }
 serde_ignored = "0.1.14"
 serde_json = "1.0.149"
 serde_path_to_error = "0.1.20"
+schemars = { version = "1.2.1", features = ["indexmap2"] }
```
sunshowers, line 130: I think this is fine, but it's funny -- at my workplace (Oxide Computer) we're still stuck on schemars 0.8 due to a few reasons. I was hoping at some point to use some tooling we have at work to also analyze schema evolution over time for nextest (both this config schema and things like the schema for JSON output). I guess if I want to do that, then I'll have to either downgrade this to schemars 0.8 or find a way to move the work tooling to schemars 1.

Justfile
```diff
@@ -6,6 +6,10 @@ set positional-arguments
 help:
     just --list
 
+# Generate the JSON schema for .config/nextest.toml.
+generate-config-schema:
+    cargo run --package nextest-runner --features config-schema --bin generate-config-schema
```
sunshowers, line 11: We should add this generate check as a CI step -- I'd probably add this as part of the lint job.

nextest-runner/src/config/overrides/imp.rs
sunshowers: Could you move the parenthetical to a regular comment?

nextest-runner/src/config/core/imp.rs
```diff
@@ -1487,27 +1487,38 @@ impl NextestConfigImpl {
...
     old_setup_scripts: IndexMap<ScriptId, SetupScriptConfig>,
+
     #[serde(default)]
     scripts: ScriptConfig,
+
     #[serde(rename = "profile")]
+    #[cfg_attr(
+        feature = "config-schema",
+        schemars(with = "Option<HashMap<String, CustomProfileImpl>>")
+    )]
```
sunshowers, line 1540: This is pretty non-obvious so I think it needs a comment.

I would either inline a comment here, or add a short comment here and refer to the longer writeup in the helper binary.

.config/tombi.toml
sunshowers: Thanks! Tested it out a bit -- this is wonderful.

nextest-runner/src/bin/generate-config-schema.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) The nextest Contributors
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use camino::{Utf8Path, Utf8PathBuf};
+use miette::{Context, IntoDiagnostic, Result};
+use nextest_runner::config::core::nextest_config_schema;
+use std::fs;
+
+fn main() -> Result<()> {
```
sunshowers, line 9: I feel like there needs to be a writeup somewhere of some of the internal design decisions taken while generating the schema:

* how `deny_unknown_fields` is meant to play the role of `serde_ignored` (which is why you can't rely on schemars's serde inheritance
* related, why some fields of type `T` say that their schema is `Option<T>`
* how types with custom deserializers are handled

All of these make sense, but I'm likely going to forget about them when I look a few months from now :)

Thoughts about putting this writeup in this file? This would be a natural place to look. Maybe if it gets complex enough we can even have a small design doc on the nextest site.

jsonschemas/repo-config.json
```diff
@@ -0,0 +1,1321 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "NextestConfigDeserialize",
+  "type": "object",
+  "properties": {
```
sunshowers, line 5: As a followup PR, would you be willing to write something that (1) embeds this schema within nextest, and (2) adds a `cargo nextest self config-schema` command that dumps out a file? I would prefer we not have to include schemars as part of the release (so your optional feature is fantastic, thanks), but it would still be nice to have access to the schemal.

nextest-runner/src/config/core/imp.rs
```diff
@@ -1487,27 +1487,38 @@ impl NextestConfigImpl {
 
 // This is the form of `NextestConfig` that gets deserialized.
 #[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "config-schema", schemars(deny_unknown_fields))]
 #[serde(rename_all = "kebab-case")]
-struct NextestConfigDeserialize {
+pub(crate) struct NextestConfigDeserialize {
```
sunshowers, line 1497: Curious, how hard would it be to add a test which ensures the default config passes schema validation?

sunshowers (commented):
This is really, really great work -- very thorough, thank you. Just a few questions.

nextest-runner/src/config/core/imp.rs
ya7010: Added a test for `.config/nextest.toml`.

77d356af559e8fc61213f63848b97709c8ee17ea

nextest-runner/src/config/overrides/imp.rs
ya7010: Done!

95771d4a32a622b11f6f02517a912661c3216001

nextest-runner/src/config/elements/archive.rs
ya7010: Done: https://github.com/nextest-rs/nextest/pull/3314/changes/0bc28ae55978b2d34dce7398b23a8aff6eab2c7f

Cargo.toml
ya7010: Support for schemars v0.8 has been added.

Since the new syntax cannot be used, it is necessary to use the more verbose syntax.

https://github.com/nextest-rs/nextest/pull/3314/changes/f0c222265cbf627e9c14af48db99e8db991db295

nextest-runner/src/bin/generate-config-schema.rs
ya7010: I moved the comments to NextestConfigDeserialize because having them in `generate-config-schema.rs` felt too far away from the definitions.

b30accb692bb2edde4b9cd38f311b4af3f1638f9

jsonschemas/repo-config.json
ya7010: Done: https://github.com/nextest-rs/nextest/pull/3314/changes/3b4323897b47a78bcb954bd3a5c9772907036174

nextest-runner/src/config/core/imp.rs
ya7010: Done: b30accb692bb2edde4b9cd38f311b4af3f1638f9

ya7010:
@sunshowers 

I’ve gone ahead and fixed all the review comments I could address.

Cargo.toml
ya7010: If you're okay with the latest `schemars`, I can revert it. How would you like to proceed?

sunshowers: Yeah I think I'd prefer the latest schemars here.

ya7010: Reverted: https://github.com/nextest-rs/nextest/pull/3314/changes/676910689b1f271a66473a9e6fd6c38e2bb27d5f

sunshowers:
Looks great. I'll address the remaining issues I identified myself before landing this. Thanks again for doing all this work and for maintaining Tombi (I'm excited to finally have JSON Schema support for the nextest config -- Taplo was a bit of a brutal experience)

PR #3331, opened 2026-05-08
Add support to use rustc --print=cfg as fallback to resolve target information

### LINKED ISSUE #3319: Feature request: Allow custom targets that do not adhere to target-lexicon
### Discussed in https://github.com/nextest-rs/nextest/discussions/3224

<div type='discussions-op-text'>

<sup>Originally posted by **mhatzl** March 27, 2026</sup>
### Motivation

At Ferrous Systems we have some custom Rust targets that we use for testing purposes, which do not pass the target naming scheme detected by the `target-lexicon` crate used indirectly in `cargo-nextest` via `target-spec`.

Therefore it is not possible to use `cargo-nextest` for those targets, because I get
e.g. `error: unknown triple string: aarch64-a53-ferrocene.facade`

I could not find anything in the docs that would allow to specify custom targets.

### Proposal

Ideally, I'd like to define custom targets in the nextest configuration file,
because it then does not require to add support to `target-spec` or `target-lexicon` for every new target that does not strictly follow the Rust target naming scheme.

If this is something that would get merged, I'd try to get a PR up myself.

### Alternatives

I locally went down the chain and *patched* `target-lexicon` to explicitly detect one of our test targets. This kind of worked, but `target-spec` currently uses `target-lexicon` directly with v0.13.5 and indirectly as re-export from `cfg-expr` at v0.13.3.
This makes it quite hard to add support for new targets via `target-lexicon`, because it would require a bump in `target-lexicon`, `cfg-expr`, and `target-spec`.

### Additional context

_No response_</div>

mhatzl:
closes #3319 

Uses `rustc --print=cfg --target=<given target>` in nextest internally to resolve the platform in case target is not built-in or custom JSON target.
If this fails, heuristic check is done as before.

bumps `target-spec` to v3.6.0 to get support for `rustc --print=cfg`.
(Cargo.lock was already at v3.6.0, but Cargo.toml wasn't)

To test this feature with de upstream rustc, I created a shell script and set it for RUSTC only for this one test case.
The shell script forwards all rustc calls except when `--print=cfg` is the first argument.
In this case it returns a fixed cfg that is then compared in the test case.

Wouldn't know how else to test this, because `rustc --print=cfg` requires to pass the target as well, but upstream rustc only knows targets that would get resolved by the built-in case, so would never reach the `--print=cfg` fallback.


sunshowers:
Thanks for doing this! Have a look at the rustc-shim binary, you should be able to fake a rustc invocation using that: https://github.com/nextest-rs/nextest/blob/a2c4c13e16946c0c5ad06a9fedeabf10f83e4fe3/integration-tests/test-helpers/rustc-shim.rs

mhatzl:
thanks @sunshowers 
will try to use rustc-shim

mhatzl:
switched to `rustc-shim` and removed the script.
CI is now failing, because I expect `NEXTEST_BIN_EXE_rustc_shim` to be set and pointing to the `rustc-shim` binary.

I don't see how this is set for the integration-test, because inside `set_env_vars_for_test`, the env is also only checked but not set.
Since this is the only place where I could find a mention of this env, I guess it is set in GitHub environments?

Should I hardcode the path using `NEXTEST_WORKSPACE_ROOT` and then `target/release/rustc-shim`?

sunshowers:
Hmm, nextest is responsible for setting it. I'll have a look.

mhatzl:
> Hmm, nextest is responsible for setting it. I'll have a look.

my added test is still under `nextest-runner/tests/integration/target_triple.rs`.
didn't want to move it to `integration-tests/tests`, because the test is about the target triple. Maybe `NEXTEST_BIN_EXE_rustc_shim` is only available for `integration-tests/tests`?

sunshowers:
Ah yeah it has to live under integration-tests.

I would have a look at `cargo nextest debug build-platforms` -- you may need to add a mode which just prints out all the platform settings. The debug commands aren't stable so you can make whatever changes you need to them.

mhatzl:
added `cargo debug print-cargo-target-arg` to compare the `CargoTargetArg` output of a set target and that seems to work quite good now.

not sure about the MSRV fails, because edition=2024 was already set.
the macos-latest fails are also strange

sunshowers:
The nextest-metadata MSRV thing will be addressed by #3342.

This generally looks great, thanks! I'll take care of the rest (want to make a few tweaks to the output format before landing).

mhatzl:
perfect. thank you. 

and thanks for the pointers

integration-tests/tests/integration/main.rs
```diff
@@ -2374,6 +2374,56 @@ fn test_debug_build_platforms_json() {
...
+    let mut command = CargoNextestCli::for_test(&env_info);
+    let output = command
+        .args([
+            "debug",
+            "build-platforms",
+            "--target",
+            triple_str,
+            "--output-format",
+            "json-pretty",
+        ])
```
sunshowers, line 2399: I added JSON output to `debug build-platforms` in #3343 and used it here.

integration-tests/test-helpers/rustc-shim-print-cfg.txt
```diff
@@ -0,0 +1,13 @@
...
+target_arch="arm"
+target_endian="little"
+target_env=""
+target_has_atomic="16"
+target_has_atomic="32"
+target_has_atomic="8"
+target_has_atomic="ptr"
+target_os="none"
+target_pointer_width="32"
+target_vendor="unknown"
```
sunshowers, line 13: Moved this to a file to avoid duplicating it across the shim and the test.

integration-tests/tests/integration/snapshots/integration__custom_target_rustc_cfg_json.snap
```diff
@@ -0,0 +1,29 @@
+---
+source: integration-tests/tests/integration/main.rs
+expression: "&stdout"
+---
+{
```
sunshowers, line 5: This is the new snapshot.

sunshowers:
The macOS failures are definitely unrelated to the PR and I think are an infrastructure issue with GitHub.

PR #3424, opened 2026-06-26
[cargo-nextest] add a help topic for filtersets

sunshowers:
Add `cargo nextest help filterset` (and the `filtersets` alias), which renders the filterset DSL reference to the terminal. (As part of this, the reference markdown has been moved to being into `nextest-filtering`).


sunshowers:
I think this is in pretty good shape now. Let's get this in and then continue to iterate on it.

PR #3443, opened 2026-07-03
[WIP] Implement experimental test result caching

camelid:


cargo-nextest/src/dispatch/core/run.rs
```diff
@@ -986,6 +1002,32 @@ impl App {
...
         };
 
+        // Resolve the cache backend, if enabled (gated behind an experimental
+        // env var). An unresolvable cache dir disables caching rather than
+        // erroring, since the cache must never fail a run. Also disabled by
+        // `--no-cache`/`NEXTEST_NO_CACHE`, and for reruns: a test a rerun wants to
+        // run is one that did not pass last time, so a stale cached pass could
+        // only wrongly skip it — and reruns already skip prior passes themselves.
+        let cache_enabled =
+            std::env::var("NEXTEST_EXPERIMENTAL_RESULT_CACHE").as_deref() == Ok("1");
```
sunshowers: * Let's add a user config `experimental.result-cache = true` for this as well.
* There's a few other things that shouldn't be cached, such as stress and debugger/tracer runs.
* Reruns should add to the cache but not consult it.

cargo-nextest/src/dispatch/core/run.rs
```diff
@@ -1270,22 +1333,15 @@ impl App {
...
-            version_env_vars: &version_env_vars,
             profile_name: profile.name(),
             double_spawn,
             target_runner,
         };
 
-        let test_list = self.build_test_list(&ctx, binary_list, &test_filter, &profile)?;
+        // Benchmark results are not cached: their value is the measured time,
+        // not a pass/fail outcome.
+        let test_list = self.build_test_list(&ctx, binary_list, &test_filter, &profile, None)?;
```
sunshowers: Could you put the "not cached" comment on a line above the `None`? That would make it obvious which line the comment applies to.

cargo-nextest/src/dispatch/core/list.rs
```diff
@@ -128,21 +126,16 @@ impl App {
...
                     profile_name: profile.name(),
                     double_spawn,
                     target_runner,
                 };
 
-                let test_list = self.build_test_list(&ctx, binary_list, &test_filter, &profile)?;
+                // Listing never consults the result cache: it enumerates tests
+                // rather than running them.
+                let test_list =
+                    self.build_test_list(&ctx, binary_list, &test_filter, &profile, None)?;
```
sunshowers: I think `None` is the right default, but I think we should accept a `--use-cache` that filters out cached results.

nextest-metadata/src/test_list.rs
```diff
@@ -1013,6 +1018,7 @@ impl fmt::Display for MismatchReason {
             }
             MismatchReason::Partition => write!(f, "is in a different partition"),
             MismatchReason::RerunAlreadyPassed => write!(f, "already passed"),
+            MismatchReason::UnchangedSinceCached => write!(f, "unchanged"),
```
sunshowers: let's say something like "binary unchanged since last cached result"

nextest-runner/src/test_filter.rs
```diff
@@ -543,6 +563,12 @@ impl TestFilter {
             };
         }
 
+        if self.is_unchanged_since_cached(binary_query.binary_id, test_name) {
+            return FilterMatch::Mismatch {
+                reason: MismatchReason::UnchangedSinceCached,
+            };
+        }
```
sunshowers: This is definitely not correct in a bunch of different ways, such as that a typoed filter would result in an exit code of 0. A previously-cached ignored test would be classified as cached rather than ignored as well.

I think this should apply at the very end of all filters rather than the beginning.

But thinking about it more, I think there's a larger detail here, which is that it isn't correct to treat a cache hit as a skipped test at all. Rather, we should treat it as matching a filter, and also change the `FilterMatch::Matches` variant to record whether a cache hit occurred. Then, at runtime, we should produce a new `TestEventKind::TestCached` event. 

I think this has some nice downstream effects:

* a lot of the special-case handling gets cleaned up
* We get to handle cached tests cleanly in JUnit reports and such, which should report cached runs as successes (similar to Bazel, I believe).
* The final summary should say something like "120 passed (100 cached)" rather than marking cache hits as skipped.

nextest-metadata/src/test_list.rs
```diff
@@ -980,6 +980,10 @@ pub enum MismatchReason {
     /// This is a rerun and the test already passed.
     RerunAlreadyPassed,
 
+    /// This test's result is cached and the test binary is unchanged since it
+    /// was cached, so the cached passing result is reused.
+    UnchangedSinceCached,
```
sunshowers: This needs a store format version bump (a minor bump to version 2.1 should suffice).

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,402 @@
...
+    None => panic!("7 days is a valid TimeDelta"),
+};
+
+/// Minimum time between automatic prunes. Mirrors the record store's cadence so
+/// pruning stays off the hot path — a prune walks every cache directory, which
+/// is cheap but pointless to repeat on every run.
+const PRUNE_INTERVAL: TimeDelta = match TimeDelta::try_days(1) {
+    Some(d) => d,
+    None => panic!("1 day is a valid TimeDelta"),
+};
```
sunshowers: We should make these configurable via user config.

cargo-nextest/src/dispatch/core/run.rs
```diff
@@ -1834,5 +1911,39 @@ mod tests {
...
+            stats,
+            Some(NoTestsBehaviorOpt::Fail),
+            None,
+            false,
+            3,
+        );
+        assert!(
+            result.is_ok(),
+            "cached skips are treated as a successful no-op regardless of --no-tests"
+        );
```
sunshowers: This doesn't seem consistent.

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,166 @@
...
+        let entry = CacheEntry {
+            created_at: now,
+            last_hit_at: now,
+        };
+        if let Err(error) = self.backend.store(&key, &entry) {
+            warn!(
+                "cache: failed to store result for {} in {}: {error}",
+                test_instance.test_name, test_instance.binary_id,
+            );
+        }
```
sunshowers: This will cause a ton of write amplification. We should only write out a manifest once all tests scheduled to be run for a binary have completed. (The JUnit stuff has an example of how to do this, I believe.)

cargo-nextest/src/dispatch/core/run.rs
```diff
@@ -138,6 +140,18 @@ pub struct TestRunnerOpts {
...
+    ///
+    /// Has no effect unless the experimental result cache is enabled. When the
+    /// cache is enabled, every test runs as normal and no results are stored.
+    #[arg(
+        long,
+        name = "no-cache",
+        env = "NEXTEST_NO_CACHE",
+        value_parser = BoolishValueParser::new(),
+    )]
+    pub(crate) no_cache: bool,
```
sunshowers, line 156: We need a set of integration tests for this that cover the full workflow, and interaction with all the other features (such as stress runs, reruns, debugger/tracer, etc).

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,402 @@
...
+        if refreshed {
+            self.write_manifest(&binary_hash_hex, &manifest)?;
+        }
+
+        Ok(())
+    }
+
+    fn store(&self, key: &CacheKey, entry: &CacheEntry) -> Result<(), CacheError> {
+        let binary_hash_hex = key.binary_hash_hex();
+        let mut manifest = self.read_manifest(&binary_hash_hex)?;
```
sunshowers: Hmm, does this mean that if the manifest is corrupt it'll just be like that forever? That seems wrong. Unlike recorded runs which are real state, this is a cache, so I think if it is useless we should just discard it.

nextest-runner/src/cache/imp.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// [`TestFilter`]: crate::test_filter::TestFilter
+/// [`CacheWriter`]: crate::cache::CacheWriter
+#[derive(Clone, Debug, Default)]
+pub struct ComputedCacheInfo {
+    /// Cached-passing test names, keyed by binary ID.
+    pub passing: HashMap<RustBinaryId, BTreeSet<TestCaseName>>,
+
+    /// Content hash of each binary that could be hashed, keyed by binary ID.
+    pub binary_hashes: HashMap<RustBinaryId, ContentHash>,
```
sunshowers: Let's use a single `iddqd::IdHashMap` for this.

nextest-runner/src/cache/key.rs
```diff
@@ -0,0 +1,117 @@
...
+/// negligible.
+pub fn hash_reader<R: Read>(mut reader: R) -> io::Result<ContentHash> {
+    let mut hasher = Xxh3::new();
+    let mut buf = [0u8; HASH_CHUNK_SIZE];
+    loop {
+        let n = reader.read(&mut buf)?;
+        if n == 0 {
+            break;
+        }
+        hasher.update(&buf[..n]);
```
sunshowers, line 120: Needs a retry on `ErrorKind::Interrupted`.

nextest-runner/src/cache/imp.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Returns `None` if the base cache dir or canonical workspace root can't be
+/// resolved; caching is then disabled rather than guessed, since it must never
+/// fail a run.
+pub fn default_cache_dir(workspace_root: &Utf8Path) -> Option<Utf8PathBuf> {
+    let strategy = choose_base_strategy().ok()?;
+    let base = Utf8PathBuf::from_path_buf(strategy.cache_dir()).ok()?;
+    let canonical_workspace = workspace_root.canonicalize_utf8().ok()?;
+    let encoded_workspace = encode_workspace_path(&canonical_workspace);
+    Some(cache_dir_from_base(base, &encoded_workspace))
+}
```
sunshowers: I think we should return a richer error here and print that (in the caller) if it couldn't be determined.

cargo-nextest/src/dispatch/core/run.rs
```diff
@@ -1175,20 +1219,39 @@ impl App {
...
+            }
+            reporter.report_event(event)
+        })?;
         let reporter_stats = reporter.finish();
 
+        // Self-prune the result cache so it never grows without bound. Binaries
+        // consulted this run had their access times refreshed, so pruning by age
+        // leaves them alone. Best-effort and rate-limited internally.
+        if let Some(backend) = &cache_backend {
+            backend.prune_if_needed(Utc::now());
```
sunshowers: We should log the prune stats at debug level.

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,166 @@
...
+/// since they are not a deterministic function of the binary — a cached failure
+/// would also defeat retries.
+///
+/// A pass that *leaked handles* or *timed out but was tolerated* is likewise not
+/// cached: those depend on subprocess and timing behavior, so caching them would
+/// suppress leak and timeout detection on later runs.
+pub struct CacheWriter<'a> {
+    backend: &'a dyn CacheBackend,
+    binary_hashes: &'a HashMap<RustBinaryId, ContentHash>,
+}
```
sunshowers, line 42: If you do what I said and buffer up results, then the `CacheWriter` becomes stateful. I'd then switch the write path over to using an actor, similar to how rerun events get sent to an actor thread. Then the actor can have exclusive ownership over the buffered state and be responsible for storing results.

nextest-runner/src/cache/result.rs
```diff
@@ -0,0 +1,48 @@
...
+/// A cached test result entry.
+///
+/// Only passing test results are stored in the cache. Failed and flaky tests are never cached —
+/// they are always re-executed to detect intermittent issues.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CacheEntry {
+    /// When this result was first stored in the cache.
+    pub created_at: DateTime<Utc>,
+
+    /// When this cache entry was last accessed, either stored or looked up.
```
sunshowers: Is "either stored or looked up" correct here? Only stores refresh, right?

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,166 @@
...
+use nextest_metadata::RustBinaryId;
+use std::collections::HashMap;
+use tracing::warn;
+
+/// Observes test events and stores passing results in the cache.
+///
+/// A result is cached only when the test cleanly passed on its first attempt.
+/// Flaky (failed then passed on retry) and stress results are never cached,
+/// since they are not a deterministic function of the binary — a cached failure
+/// would also defeat retries.
```
sunshowers: Hmm, I don't think this comment is quite right? Can you make a table of all the different possibilities here and talk about which ones result in cache lookups and which ones result in cache storage? It's worth thinking about this systematically.

nextest-runner/src/cache/parallel.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) The nextest Contributors
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! A small work-stealing helper for hashing test binaries in parallel.
```
sunshowers: Where's the work-stealing?

Also can we use rayon here?

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,402 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct CacheMeta {
+    version: u32,
+    last_pruned_at: Option<DateTime<Utc>>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Manifest {
+    version: u32,
+    entries: BTreeMap<String, ManifestEntry>,
```
sunshowers: This should use an `iddqd::IdOrdMap` where the key is a `&'a TestCaseName`.

sunshowers (commented):
Thanks for doing this! My understanding is that you're using an LLM to assist with development, so I'm going to hold this to a higher standard than otherwise, as explained in `AGENTS.md`. In particular, I'm focusing here on the general structure being clear and maintainable, and would like this change to not introduce any new tech debt since LLMs make it easy to address tech debt.

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        manifest.entries.insert(
+            key.test_name().to_owned(),
+            ManifestEntry {
+                created_at: entry.created_at,
+                last_hit_at: entry.last_hit_at,
+            },
+        );
+
+        self.write_manifest(&binary_hash_hex, &manifest)
```
sunshowers: One of the things I'm concerned about here is what happens if a failure _is_ seen in a rerun/stress test/some other situation where the cache is disabled.

I think it is worth building a formal model for this, with verification in two ways:

* proof via exhaustive enumeration
* stateful property-based tests (model-based tests) to show that the implementation matches the model (refinement)

We follow this pattern in reruns as well, and I think (similar to that) this state machine is complex enough that it needs this treatment.

I had Claude write up a sketch for what such a model and verification strategy might look like, and it came up with https://gist.github.com/sunshowers/ac913e7bde761bcfe085813852fe313d. (If I were writing the design document here, I would take this as inspiration but describe it in my own words.)

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+/// Top-level cache metadata, stored once at the cache root in `meta.json`.
+///
+/// Currently just tracks the last prune so automatic pruning can be rate-limited
+/// (see [`FsBackend::prune_if_needed`]).
+#[derive(Debug, Serialize, Deserialize)]
+struct CacheMeta {
+    version: u32,
+    last_pruned_at: Option<DateTime<Utc>>,
```
sunshowers: Does pruning also prune caches for which the manifest couldn't be read?

cargo-nextest/src/dispatch/core/run.rs
camelid: Regarding (2), stress test results are already excluded, though tests that have already been cached are skipped. Would you prefer that stress tests automatically disables the cache?

sunshowers: I don't think enable/disable cache is a single boolean -- rather, it is a pair of (write to cache, read from cache) booleans. I'd encourage you to make a decision table with every possibility (`--no-cache`, stress runs, etc) and whether that results in reading from or writing to the cache. I'd expect the decision table to have somewhere on the order of 10-20 rows.

nextest-runner/src/cache/fs_backend.rs
camelid: Yeah, I wasn't sure the best way to handle these types of errors which are quite unlikely in practice. I don't want bugs in the cache to annoy people just trying to run the tests, while also not swallowing the errors. If the manifest is corrupt (or if any other issue is encountered with the cache), it will emit a warning to the user but otherwise ignore the cache.

nextest-runner/src/cache/result.rs
camelid: Lookups also refresh the last access time. The last access time is so that we prune cached results that are no longer used. If we only refreshed on store, then we'd also prune cached tests that are being hit, which would defeat the purpose of the cache.

nextest-runner/src/cache/fs_backend.rs
sunshowers: I think printing a warning and then discarding the cache is the right thing to do here. It is a cache, so discarding it must always be correct.

nextest-runner/src/cache/parallel.rs
camelid: Yeah, work stealing is not the right way to describe it. It's just a queue that's shared between the threads.

I actually wanted to use rayon originally but wasn't sure if you were OK taking on another dependency. Happy to switch to it if that's fine.

nextest-runner/src/cache/fs_backend.rs
camelid: Agreed, just didn't do that yet since I wanted feedback on the overall design 👍

cargo-nextest/src/dispatch/core/list.rs
camelid: Hmm, I wonder if a better way to do it would be to always use the cache if available, but only use it to add a note on cached tests rather than do filtering? Up to you though. Adding a `--use-cache` flag just for test listing feels a bit weird, especially since for running tests it's the opposite semantics (cache by default, `--no-cache` opt-out).

nextest-runner/src/cache/fs_backend.rs
camelid: It does now. (Actually the project's whole cache is just discarded at the moment the corrupt manifest is found.)

nextest-runner/src/cache/backend.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn write(&self, writes: &[CacheWrite]) -> Result<(), CacheError>;
+}
+
+/// Errors that can occur during cache operations.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum CacheError {
+    /// An I/O error occurred while accessing the cache storage.
+    #[error("cache I/O error")]
+    Io(#[from] std::io::Error),
```
sunshowers, line 86: This loses information about the event that led to the I/O error. Prefer having a bunch of different variants here each of which corresponds to a specific operation. (In general, `#[from] std::io::Error` is not a good idea for this reason.)

nextest-runner/src/cache/backend.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The cached data could not be deserialized.
+    #[error("cache data is corrupt or incompatible: {0}")]
+    InvalidData(String),
```
sunshowers, line 90: Use an enum here, e.g., `InvalidDataReason`, with explicit variants.

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,406 @@
...
+        AtomicFile::new(&path, OverwriteBehavior::AllowOverwrite)
+            .write(|f| {
+                let mut writer = BufWriter::new(f);
+                serde_json::to_writer(&mut writer, manifest).map_err(io::Error::from)?;
+                writer.flush()
+            })
+            .map_err(|e| match e {
+                atomicwrites::Error::Internal(io) | atomicwrites::Error::User(io) => {
+                    CacheError::Io(io)
+                }
```
sunshowers, line 118: These should be separate variants, for example. Or you can expose the atomicwrites error here.

nextest-runner/src/list/test_list.rs
```diff
@@ -331,6 +338,28 @@ impl<'g> TestList<'g> {
...
+            let mut cached_passing = HashMap::new();
+            for binary in cache_info.binaries {
+                binary_hashes.insert(binary.binary_id.clone(), binary.hash);
+                cached_passing.insert(binary.binary_id, binary.passing);
+            }
+            let mut filter = filter.clone();
+            filter.set_cached_passing(cached_passing);
+            filter
+        });
+        let filter = cache_filter.as_ref().unwrap_or(filter);
```
sunshowers, line 361: 1. Please de-slop these comments. It's hard to follow what this means.
2. Does only gating on `consult` mean that reruns never record anything? That seems wrong.
3. Cloning the filter like this feels like the wrong shape. I think I'd prefer something like:

```rust
let cache_info: Option<ComputedCacheInfo> = cache
    .consult()
    .map(|backend| Self::collect_cache_info(&*backend, &parsed_binaries));

// ... then, introducing a type that's returned by the filter:
pub(crate) enum FilterResult {
    Matches,
    Mismatch(MismatchReason),
}

// ... then, build_suites maps a FilterResult to FilterMatch by
// consulting the cache to determine the `cached` value.
```

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,248 @@
...
+            } => {
+                let result = execute_result(run_statuses);
+                if TestCacheDecision::compute(self.policy, result) == TestCacheDecision::RecordPass
+                    && let Some(key) = self.key_from_test_instance(test_instance)
+                {
+                    // Ignore send errors.
+                    _ = self.sender.send(CacheWrite::Store { key });
+                }
+            }
+            TestEventKind::TestCached { test_instance, .. } => {
```
sunshowers, line 91: Not very confident about the interaction between reruns and this. What happens in the case where:

* A run is mostly served from cache, but some tests fail.
* A rerun occurs?

A cached success should map to `TestOutcome::Passed`.

Also, I think libtest-json output needs to be updated to handle `TestCached` correctly as well.

nextest-runner/src/cache/backend.rs
```diff
@@ -0,0 +1,91 @@
...
+/// Trait abstracting cache storage for test results.
+///
+/// This holds only storage operations (lookup, batch query, batch write, stats)
+/// and assumes nothing about the underlying medium. Filesystem-specific cleanup,
+/// such as directory eviction, is an inherent method on
+/// [`FsBackend`](super::FsBackend) instead.
+///
+/// # Contract
+///
+/// Reads ([`lookup`](Self::lookup), [`passing`](Self::passing)) and writes
```
sunshowers, line 51: Doesn't look like lookup is a thing.

nextest-runner/src/reporter/aggregator/junit.rs
```diff
@@ -253,6 +253,9 @@ impl<'cfg> MetadataJunit<'cfg> {
                 //
                 // testsuite.add_testcase(testcase);
             }
+            TestEventKind::TestCached { .. } => {
+                // TODO: report cached tests as successes
+            }
```
sunshowers, line 258: Definitely need to address this. People depend on the JUnit support and we can't just drop cached tests like this.

nextest-runner/src/cache/fs_backend.rs
```diff
@@ -0,0 +1,406 @@
...
+                ),
+                Err(error) => format!("corrupt manifest: {error}"),
+            },
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(Manifest::empty()),
+            Err(error) => format!("read error: {error}"),
+        };
+
+        warn!("cache: discarding cache at {}: {reason}", self.cache_dir);
+        self.discard()?;
+        Ok(Manifest::empty())
```
sunshowers, line 89: This is too aggressive. For example, EPERM shouldn't result in us trying to discard the cache (which will likely fail with the same EPERM).

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,248 @@
...
+            binary_hashes: test_list.binary_hashes(),
+        })
+    }
+
+    /// Inspects an event and writes the corresponding cache update, if any.
+    ///
+    /// Storage errors only warn — a failing cache never fails a passing run — but
+    /// are surfaced rather than dropped, since one likely indicates a bug while
+    /// this feature is experimental.
+    pub fn observe(&self, event: &ReporterEvent<'_>) {
```
sunshowers, line 72: Can observe be `&mut self`?

nextest-runner/src/cache/writer.rs
```diff
@@ -0,0 +1,248 @@
...
+
+fn actor(backend: Arc<dyn CacheBackend>, receiver: mpsc::Receiver<CacheWrite>) {
+    let mut writes = vec![];
+    while let Ok(write) = receiver.recv() {
+        writes.push(write);
+    }
+
+    if let Err(error) = backend.write(&writes) {
+        warn!("cache: failed to write to backend: {error}");
+    }
```
sunshowers, line 131: 1. We should be doing incremental flushes here (otherwise there's no real point to using an actor once `observe` is `&mut self`).
2. What happens if multiple nextest processes are running on the same repo at the same time? Would this just be that the last writer wins? Can you spend some time characterizing this?

nextest-runner/src/list/test_list.rs
camelid: We gate on `consult` here because this is where we consult the cache to find cached passes that be reused. Recording happens later and gates on `record`. Consult is for reading (and writing touches for cache entries that were used) and record is for writing new cached passes.

sunshowers: How would we record anything to the cache if we don't know the hash of the binary?

camelid: You're right, my bad. I think I got distracted by trying to add safety with the `consult()`/`record()` helper methods, and didn't notice the issue it causes here. Another reason I need to write the integration tests. :)

camelid:
As discussed offline, I'm pivoting to implementing the cache outside of nextest as a wrapper, with only minimal changes to nextest (like having wrappers be able to report labels for their test results).

PR #3459, opened 2026-07-08
[nextest-runner] cleanup RecordReporter a bit

camelid:
Removes an unneeded lifetime parameter and unnecessary Option wrapper.


sunshowers:
Thanks!

PR #3470, opened 2026-07-13
[site] document snap rustup caveat in installation instructions

### LINKED ISSUE #3398: Installation instruction in tutorial fails when using `snap` to install `rustup`.
I installed my rust toolchain via rustup (`sudo snap install rustup --classic`, then install the default stable toolchain.

When executing the installation command recommend in nextest's tutorial, I get the following error:

```
curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
tar: /home/ambrus/.cargo/bin: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
curl: (23) Failure writing output to destination

```
My rustup installed cargo here: `/snap/bin/cargo`.

It would be useful to create either a less intrusive setup process or the setup script detecting common cargo paths or using `$(which cargo)` in the tutorial instead of a hardcoded path.

hiro-nikaitou:
When rustup is installed via `snap install rustup`, `~/.cargo/bin` may not exist. This causes the `tar ... -C ${CARGO_HOME:-~/.cargo}/bin` command to fail with a confusing error.

This change adds a note to the Linux installation instructions warning about this and suggesting an alternative extraction directory.

Closes #3398

### DIFF
--- a/site/src/docs/installation/pre-built-binaries.md
+++ b/site/src/docs/installation/pre-built-binaries.md
@@ -27,6 +27,8 @@ The instructions below are suitable for both end users and CI. These links will 
 
         If you'd like to stay on the 0.9 series to avoid breaking changes (see the [stability policy](../stability/index.md) for more), replace `latest` in the URL with `0.9`.
 
+        If you installed Rust via `snap install rustup`, the `~/.cargo/bin` directory may not exist by default. Extract the binary to a different directory in your PATH, for example `~/.local/bin`.
+
     Run in a terminal:
 
     ```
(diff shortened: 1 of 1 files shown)


sunshowers (changes requested):
Thanks for doing this. This is a good start but needs a bit more research:

* What does snap do that's special?
* Are there other environments that are similarly special? Some that come to mind are homebrew, nix, and other places where rustup is installed via the system package manager.
* Instead of instructing people to do something else, should we update the command itself?

Basically I'd like to characterize this issue properly.

hiro-nikaitou:
Thanks for the review! I investigated the three questions:

**1. What does snap do that's special?**
Snap installs rustup at `/snap/bin/rustup` without creating `~/.cargo/`. The directory only gets created when the user runs `rustup install` commands and sources `~/.cargo/env`. The `curl | tar -C ~/.cargo/bin` command fails because the target directory simply doesn't exist yet.

**2. Other environments?**
Same pattern for homebrew (`/opt/homebrew/bin/cargo` on ARM, `/usr/local/bin/cargo` on Intel) and nix (store paths) — `~/.cargo/bin` is a rustup convention, not a cargo convention. Any non-rustup installation of Rust/cargo won't have it.

**3. Fix the command itself?**
I think so. Prepending `mkdir -p${CARGO_HOME:-~/.cargo}/bin` makes the command self-healing — if the directory already exists it's a no-op, otherwise it creates it. It's still one compound command (no script), backwards compatible, and covers all the above cases.

Updated the PR: replaced the snap-specific note with `mkdir -p` on all Unix `curl|tar` commands, and updated the info note to mention snap/homebrew/nix.

PR #3514, opened 2026-07-31
[nextest-runner] support cargo config `include` key

### LINKED ISSUE #3511: Bug: `[env]` variables from included config files (via `include` directive) are not passed to test processes
### Description of the issue

**Description:** When `.cargo/config.toml` uses the `include` directive to pull in another TOML file, any `[env]` variables defined in that included file are not set when nextest runs tests. The same variables work correctly with `cargo test`.

**Steps to reproduce:**
1. Create `.cargo/config.toml`:
   ```toml
   include = ["include.toml"]

   [env]
   DIRECT = "direct"
   ```

2. Create `.cargo/include.toml`:
   ```toml
   [env]
   INCLUDE = "include"
   ```

3. Add tests that read both variables at runtime:
   ```rust
   #[test]
   fn env_from_config_toml_direct() {
       // DIRECT is set in [env] inside .cargo/config.toml itself
       assert_eq!(&std::env::var("DIRECT").unwrap(), "direct");
   }

   #[test]
   fn env_from_included_toml() {
       // INCLUDE is set in [env] inside .cargo/include.toml, pulled in via
       // `include = ["include.toml"]` at the top of .cargo/config.toml
       assert_eq!(&std::env::var("INCLUDE").unwrap(), "include");
   }
   ```

4. Run `cargo test` → all tests pass
5. Run `cargo nextest run` → `env_from_included_toml` fails, `env_from_config_toml_direct` passes

### Expected outcome

Both `DIRECT` and `AFF` should be set when nextest runs tests, since cargo itself resolves the `include` directive and both variables are visible to `cargo test`.

### Actual result

`cargo nextest run` passes `DIRECT` (from the `[env]` section in `config.toml` directly) but does **not** pass `INCLUDE` (from the `[env]` section in the included file):

```
    Starting 3 tests across 1 binary
        PASS [   0.006s] (1/3) aff2 tests::it_works
        PASS [   0.006s] (2/3) aff2 tests::env_from_config_toml_direct
        FAIL [   0.007s] (3/3) aff2 tests::env_from_included_toml
  stderr ───
    thread 'tests::env_from_included_toml' panicked at src/lib.rs:19:42:
    called `Result::unwrap()` on an `Err` value: NotPresent
```

`cargo test` passes all three tests without issue.

### Nextest version

```text
cargo-nextest 0.9.140 (a9fef2964 2026-07-05)
release: 0.9.140
commit-hash: a9fef2964e34f64ed4fceeee7c0c3559ce560920
commit-date: 2026-07-05
host: x86_64-unknown-linux-gnu
```

### Additional context

The motivation for using `include` here is to separate the env variables that are system-dependent and auto-generated (produced by a environment setup step). Keeping them in a separate file allows the generated file to be included without modifying the checked-in `config.toml`.

celinval:
Nextest reimplements Cargo's config discovery but it didn't handle the top-level `include` key stabilized in 1.94. Deserialization just ignored the `include` in a discovered `.cargo/config.toml`.

This change now loads each config file's includes recursively, matching Cargo's semantics:

- `include` is a list of bare path strings or `{ path, optional }` tables. Paths are relative to the directory of the config file that specifies them and must end with `.toml`.
- A file's own values take precedence over the files it includes, and later includes take precedence over earlier ones.
- Includes are resolved recursively, with per-root-file cycle detection.
- `optional = true` silently skips a missing include, matching Cargo.

I also added unit tests and extended the fixture project's `.cargo/config` to pull in an included file and assert on its variables in the existing cargo env integration test.

Resolves #3511

### DIFF
--- a/nextest-runner/src/cargo_config/discovery.rs
+++ b/nextest-runner/src/cargo_config/discovery.rs
@@ -3,8 +3,15 @@
 
 use crate::errors::{CargoConfigError, CargoConfigParseError, InvalidCargoCliConfigReason};
 use camino::{Utf8Path, Utf8PathBuf};
-use serde::Deserialize;
-use std::collections::BTreeMap;
+use itertools::Itertools;
+use serde::{
+    Deserialize,
+    de::{self, MapAccess, Visitor, value::MapAccessDeserializer},
+};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt, io,
+};
 use toml_edit::Item;
 use tracing::debug;
 
@@ -167,20 +174,20 @@ fn parse_cli_configs(
     cli_configs: impl Iterator<Item = impl AsRef<str>>,
 ) -> Result<Vec<(CargoConfigSource, CargoConfig)>, CargoConfigError> {
     cli_configs
-        .into_iter()
         .map(|config_str| {
             // Each cargo config is expected to be a valid TOML file.
             let config_str = config_str.as_ref();
 
             let as_path = cwd.join(config_str);
             if as_path.exists() {
-                // Read this config as a file.
+                // Read this config as a file, expanding any files it includes.
                 load_file(as_path)
             } else {
                 let config = parse_cli_config(config_str)?;
-                Ok((CargoConfigSource::CliOption, config))
+                Ok(vec![(CargoConfigSource::CliOption, config)])
             }
         })
+        .flatten_ok()
         .collect()
 }
 
@@ -354,21 +361,52 @@ fn discover_impl(
         }
     }
 
-    let configs = config_paths
+    config_paths
         .into_iter()
         .map(load_file)
-        .collect::<Result<Vec<_>, CargoConfigError>>()?;
-
-    Ok(configs)
+        .flatten_ok()
+        .collect()
 }
 
+/// Loads a config file and, recursively, any files it includes via the `include` key.
+///
+/// The returned configs are in precedence order, highest first, matching the order expected by
+/// [`CargoConfigs::discovered_configs`]: a config file's own values take precedence over the files
+/// it includes, and among included files, those listed later take precedence over those listed
+/// earlier. Includes are resolved depth-first, so a file's entire include subtree outranks the next
+/// sibling include.
+///
+/// Errors if any file is reached more than once while resolving this file's includes, matching
+/// Cargo. For example, if `a` includes `b` and `c`, and `b` also includes `c`, then `c` is reached
+/// twice and is rejected. Each call tracks reached files independently, so the same file included
+/// from two separate top-level config files (e.g. at different hierarchy levels) is fine.
 fn load_file(
     path: impl Into<Utf8PathBuf>,
-) -> Result<(CargoConfigSource, CargoConfig), CargoConfigError> {
+) -> Result<Vec<(CargoConfigSource, CargoConfig)>, CargoConfigError> {
     let path = path.into();
     let path = path
         .canonicalize_utf8()
         .map_err(|error| CargoConfigError::FailedPathCanonicalization { path, error })?;
+
+    // `seen` is traversal bookkeeping shared across the whole include tree, so it is threaded
+    // through by mutable reference. The loaded configs, in contrast, are returned by each call and
+    // concatenated by the caller.
+    let mut seen = BTreeSet::new();
+    load_file_impl(path, &mut seen)
+}
+
+/// Loads a single already-canonicalized config file and its includes, returning them in precedence
+/// order (highest first).
+///
+/// `seen` tracks the paths reached so far across the entire include tree, so that a file reached
+/// more than once (via a cycle or multiple include paths) is rejected.
+fn load_file_impl(
+    path: Utf8PathBuf,
+    seen: &mut BTreeSet<Utf8PathBuf>,
+) -> Result<Vec<(CargoConfigSource, CargoConfig)>, CargoConfigError> {
+    if !seen.insert(path.clone()) {
+        return Err(CargoConfigError::IncludeReachedTwice { path });
+    }
 
     let config_contents =
         std::fs::read_to_string(&path).map_err(|error| CargoConfigError::ConfigReadError {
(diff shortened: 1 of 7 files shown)


nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -354,22 +361,53 @@ fn discover_impl(
...
+/// The returned configs are in precedence order, highest first, matching the order expected by
+/// [`CargoConfigs::discovered_configs`]: a config file's own values take precedence over the files
+/// it includes, and among included files, those listed later take precedence over those listed
+/// earlier. Includes are resolved depth-first, so a file's entire include subtree outranks the next
+/// sibling include.
+///
+/// Errors if any file is reached more than once while resolving this file's includes, matching
+/// Cargo. For example, if `a` includes `b` and `c`, and `b` also includes `c`, then `c` is reached
+/// twice and is rejected. Each call tracks reached files independently, so the same file included
+/// from two separate top-level config files (e.g. at different hierarchy levels) is fine.
```
sunshowers, line 382: 'This unfortunately isn't quite a correct reimplementation of Cargo's config detection. In particular, we must first merge configs and then read from them -- this concretely shows up in the case of something like:

in the root file:

```toml
[target.'cfg(unix)']
foo = "bar"
```

in the included file:

```toml
[target.x86_64-unknown-linux-gnu]
foo = "baz"
```

On `x86_64-unknown-linux-gnu`, Cargo would evaluate `foo` to `baz` while nextest would evaluate it to `bar`.

It's possible we may need to rewrite this wholesale to fit Cargo's model better.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -381,7 +419,56 @@ fn load_file(
             error,
         }))
     })?;
-    Ok((CargoConfigSource::File(path), config))
+
+    // The directory containing this config file. Include paths are resolved relative to it.
+    let config_dir = path
+        .parent()
+        .expect("a canonicalized config file path has a parent directory")
+        .to_owned();
```
sunshowers, line 427: Cargo uses the uncanonicalized path, and we'll need to do the same thing here.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -167,20 +174,20 @@ fn parse_cli_configs(
...
 
             let as_path = cwd.join(config_str);
             if as_path.exists() {
-                // Read this config as a file.
+                // Read this config as a file, expanding any files it includes.
                 load_file(as_path)
             } else {
                 let config = parse_cli_config(config_str)?;
-                Ok((CargoConfigSource::CliOption, config))
+                Ok(vec![(CargoConfigSource::CliOption, config)])
```
sunshowers, line 187: This needs to support includes as well.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -354,22 +361,53 @@ fn discover_impl(
...
     let path = path.into();
     let path = path
         .canonicalize_utf8()
         .map_err(|error| CargoConfigError::FailedPathCanonicalization { path, error })?;
 
+    // `seen` is traversal bookkeeping shared across the whole include tree, so it is threaded
+    // through by mutable reference. The loaded configs, in contrast, are returned by each call and
+    // concatenated by the caller.
+    let mut seen = BTreeSet::new();
+    load_file_impl(path, &mut seen)
```
sunshowers, line 395: Like Cargo, this should error out in the case of `--config a.toml --config a.toml`, or two `--config` files that share an include. (i.e., `seen` should be shared across all loaded files).

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -381,7 +419,56 @@ fn load_file(
...
+                // Optional includes silently skip missing files, matching Cargo.
+                continue;
+            }
+            Err(error) => {
+                return Err(CargoConfigError::FailedPathCanonicalization {
+                    path: include_path,
+                    error,
+                });
+            }
+        };
```
sunshowers, line 457: Glob metacharacters `*?[]` and template braces `{}` should be rejected here (Cargo does this).

nextest-runner/src/cargo_config/env.rs
sunshowers, line 54: If the root config contains:

```toml
[env]
FOO = "x"
```

And an included config contains:

```toml
[env]
FOO = { value = "y", force = true }
```

then Cargo rejects this config.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -381,7 +419,56 @@ fn load_file(
...
+            });
+        }
+
+        let include_path = config_dir.join(&include.path);
+        let canonicalized = match include_path.canonicalize_utf8() {
+            Ok(canonicalized) => canonicalized,
+            Err(error) if include.optional && error.kind() == io::ErrorKind::NotFound => {
+                // Optional includes silently skip missing files, matching Cargo.
+                continue;
+            }
```
sunshowers, line 450: Cargo uses `!abs_path.exists()` which is false for any stat failure. I would also log both skips and successful loads at the debug level.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -354,22 +361,53 @@ fn discover_impl(
...
+///
+/// `seen` tracks the paths reached so far across the entire include tree, so that a file reached
+/// more than once (via a cycle or multiple include paths) is rejected.
+fn load_file_impl(
+    path: Utf8PathBuf,
+    seen: &mut BTreeSet<Utf8PathBuf>,
+) -> Result<Vec<(CargoConfigSource, CargoConfig)>, CargoConfigError> {
+    if !seen.insert(path.clone()) {
+        return Err(CargoConfigError::IncludeReachedTwice { path });
+    }
```
sunshowers, line 409: If two includes reach the same file through a symlink, Cargo loads both of them.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -381,7 +419,56 @@ fn load_file(
...
+    // Then each include in reverse listed order: a later include takes precedence over an earlier
+    // one, and higher precedence comes first, so the last include (and its subtree) is appended
+    // before earlier ones. Each include's own subtree is resolved fully before the next sibling.
+    for include in includes.into_iter().rev() {
+        if !include.path.ends_with(".toml") {
+            return Err(CargoConfigError::IncludePathNotToml {
+                path: include.path,
+                included_from: path.clone(),
+            });
+        }
```
sunshowers, line 442: This should reject `.toml`. I'd recommend using `Utf8Path::new(&include.path).extension() != Some("toml")`, or even deserializing as a `Utf8PathBuf` directly.

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -381,7 +419,56 @@ fn load_file(
...
+            Ok(canonicalized) => canonicalized,
+            Err(error) if include.optional && error.kind() == io::ErrorKind::NotFound => {
+                // Optional includes silently skip missing files, matching Cargo.
+                continue;
+            }
+            Err(error) => {
+                return Err(CargoConfigError::FailedPathCanonicalization {
+                    path: include_path,
+                    error,
+                });
```
sunshowers, line 455: This should be wrapped inside `FailedToLoadInclude` as done at line 462, right?

nextest-runner/src/cargo_config/discovery.rs
```diff
@@ -532,4 +691,242 @@ mod tests {
...
+
+        // The rendered chain should mention every include hop.
+        let rendered = format!("{}", DisplayErrorChain::new(&err));
+        for hop in ["a.toml", "b.toml", "missing.toml"] {
+            assert!(
+                rendered.contains(hop),
+                "error chain should mention `{hop}`, got:\n{rendered}"
+            );
+        }
+    }
```
sunshowers, line 931: Can we snapshot the full error message here (maybe after applying some normalization)? Not a huge fan of `.contains` like this.

nextest-runner/src/cargo_config/discovery.rs
celinval: That makes sense. I'll see if I have time later this week to change the implementation and test this corner cases.

celinval:
Sorry, I haven't had time yet to create a second revision yet. I'll convert it to draft until I get it into a better shape.

celinval:
I did talk to the cargo team about this problem, and their config parsing and resolution is not trivial. There's also no formal specification on how it works. I think the best way to match Cargo's exact configuration resolution is to retrieve that information from Cargo itself. However, it doesn't seem like it's a trivial change either, some of them are captured in the cargo config [tracking issue](https://github.com/rust-lang/cargo/issues/9301).

In the short term, for the include case, we could implement the following:
1. Process all includes to generate an expanded `toml` table -- concatenate arrays and deep merge tables
2. Deserialize the expanded `toml` into a `CargoConfig`

This would work for the case you called out and it follows the include spec. I can't, however, guarantee that it's going to match cargo's behavior exactly. Would you prefer if we implement that logic or should I close this PR and we wait for Cargo to expose configuration resolution which I believe is the right long term fix.

sunshowers:
@celinval Thanks, I think that would work and probably be close enough. (Personally I'm a bit sad that Cargo keeps adding these extra bits of complexity to its config without a story for how to expose the information to plugins.)

celinval:
I found another corner case which is environment variable that is set as relative. That won't work either. :disappointed:

sunshowers:
Oh right... so you need provenance tracking for each element to identify which directory it came from. (I guess this is something we currently handle.)

PR #3546, opened 2026-08-14
fix docs code formatting

chirizxc:
before:

изображение

### DIFF
--- a/site/src/docs/ci-features/archiving.md
+++ b/site/src/docs/ci-features/archiving.md
@@ -82,7 +82,7 @@ build process creates artifacts outside of Cargo that are required for Rust test
 
 Use the `profile.<profile-name>.archive.include` configuration option for this. For example:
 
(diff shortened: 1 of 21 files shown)


chirizxc:
изображение

It seems like this isn't working either

chirizxc:
to fix this i run:

```console
rg -l 'title="[^"]*`' site/src/docs | ForEach-Object {
     $content = Get-Content $_ -Raw
     $content = [regex]::Replace($content, '(title="[^"]*)`([^`"]*)`([^"]*")', '$1$2$3')
     Set-Content -Path $_ -Value $content -NoNewline
}
```

sunshowers:
Thanks. Strange, I remember it working in the past. Going to spend a bit of time investigating.

sunshowers:
looks like this was caused by https://github.com/pygments/pygments/pull/3071. sigh

sunshowers (approved):

PR #3551, opened 2026-08-26
[nextest-runner] Add run-wrapper reports

camelid:
Wrappers can now report labels and groups for the tests that they run. For example, a wrapper that implements caching can report a "cached" label for cached tests or a "rerun: code changed" label for tests that were rerun. These labels are printed on the nextest output line corresponding to the test.

Groups are coarser and used in the summary line at the bottom of the nextest output. For example, if 53 tests pass and 43 of them report a "cached" group, the summary line will show "53 passed (43 cached)".

The store format version is bumped to 2.2.

LLM disclosure: I used an LLM to write this code, but I tested the code (by making use of this feature in my wrapper), and I have read the code.


nextest-runner/src/reporter/events.rs
```diff
@@ -1283,6 +1283,21 @@ pub struct OutputErrorSlice {
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(test, derive(test_strategy::Arbitrary))]
+pub struct RunWrapperReport {
+    /// A short label displayed on the per-test status line.
+    pub label: String,
+    /// An optional category used to aggregate counts in the final run summary.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub group: Option<String>,
+}
```
camelid, line 1299: I think requiring the label and making the group optional seems reasonable, but let me know if you'd prefer another way. Note that the label is only required **if** a report is generated (fully backwards compatible with wrappers that don't make reports, of course).

nextest-runner/src/runner/wrapper_report.rs
```diff
@@ -0,0 +1,191 @@
...
+    Parse(#[source] serde_json::Error),
+    #[error(
+        "the label must contain 1 to {MAX_LABEL_LEN} printable ASCII characters, and must start and end with a letter or digit"
+    )]
+    InvalidLabel,
+    #[error(
+        "the group must contain 1 to {MAX_GROUP_LEN} printable ASCII characters, and must start and end with a letter or digit"
+    )]
+    InvalidGroup,
+}
```
camelid, line 40: These restrictions might be too much or even worth skipping altogether. Let me know your thoughts.

nextest-runner/src/runner/wrapper_report.rs
```diff
@@ -0,0 +1,191 @@
...
+
+pub(super) fn new_report_path() -> Result<Utf8TempPath, ChildStartError> {
+    let path = camino_tempfile::Builder::new()
+        .prefix("nextest-run-wrapper-report")
+        .tempfile()
+        .map_err(|error| ChildStartError::TempPath(Arc::new(error)))?
+        .into_temp_path();
+
+    // The absence of a file means that the wrapper ran the test normally.
+    fs::remove_file(&path).map_err(|error| ChildStartError::TempPath(Arc::new(error)))?;
```
camelid, line 50: We remove the new temporary file so that we can tell if the run-wrapper made a report or not. Not strictly necessary, though.

nextest-runner/src/test_command.rs
```diff
@@ -77,11 +79,15 @@ impl TestCommand {
             create_command(program.clone(), args, lctx.double_spawn)
         };
 
+        // Set exactly when the wrapper participates in the command line (see
+        // `TestCommandCli::apply_wrappers`).
+        let has_wrapper = wrapper_env.is_some();
```
camelid, line 84: Is there a better or more robust way to check whether a wrapper is being used here?

camelid:
CI seems to be failing for unrelated reasons (MSRV).

sunshowers:
> CI seems to be failing for unrelated reasons (MSRV).

Thanks -- could you merge/rebase? fixed main.

nextest-runner/src/reporter/displayer/imp.rs
```diff
@@ -3568,6 +3616,63 @@ mod tests {
...
+                        reporter.output.writer_mut().unwrap(),
+                    )
+                    .unwrap();
+            },
+            &mut out,
+        );
+
+        assert_eq!(
+            out,
+            "        PASS [   1.000s] (cached) (  1/100) my-binary-id test_name\n"
```
sunshowers, line 3672: Let's put the label after the counter, not before.

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -602,7 +605,50 @@ pub(super) fn progress_str(
...
+                result: FlakyResult::Pass,
+                ..
+            } => &mut self.passed,
+            ExecutionDescription::Flaky {
+                result: FlakyResult::Fail,
+                ..
+            }
+            | ExecutionDescription::Failure { .. } => &mut self.failed,
+        };
+        *groups.entry(group.clone()).or_default() += 1;
```
sunshowers, line 637: These are wrong for exec-failed and timed-out tests.

<details>

<summary>From Claude</summary>

RunWrapperGroupCounts::record (nextest-runner/src/reporter/displayer/progress.rs:614-636) buckets on ExecutionDescription, which has only Success | Flaky | Failure. But RunStats::on_test_finished routes ExecFail to exec_failed and Timeout { Fail } to failed_timed_out, neither to failed (nextest-runner/src/reporter/events.rs:864-869). write_summary_str renders the failed-group parenthetical only inside if failed > 0 (progress.rs:724-753) and never next to exec failed or timed out.

Consequences:
- Only failures are exec-fails with group = "infra" → N passed, 1 exec failed, 0 skipped; the group is silently dropped. This is the case an infrastructure/caching wrapper most wants to report.
- One Fail + two ExecFail, all "infra" → 1 failed (3 infra), 2 exec failed; the parenthetical exceeds the count it qualifies.

The new unit test wrapper_groups_follow_the_final_test_outcome (progress.rs:1404-1445) hides this by hand-building RunStats { passed: 1, failed: 2 } for statuses that include an ExecFail; on_test_finished would produce failed: 1, exec_failed: 1. Fix both: mirror the four RunStats buckets (or better, add a single ExecutionStatuses::final_outcome() that both on_test_finished and record consume so they cannot drift), and build the test's RunStats by feeding the same statuses through on_test_finished.

</details>

(Did you have an LLM review this locally?)

nextest-runner/src/runner/wrapper_report.rs
sunshowers, line 50: I don't think `TempPath` is correct here, since its message is `#[error("error creating temporary path for setup script")]`.

I think we should probably:

* change `TempPath` to `SetupScriptTempPath` (serialization should probably stay the same)
* add a new variant for this case

nextest-runner/src/runner/wrapper_report.rs
```diff
@@ -0,0 +1,191 @@
...
+pub(super) fn read_report(path: Option<&Utf8TempPath>) -> Option<RunWrapperReport> {
+    let path = path?;
+
+    match try_read_report(path) {
+        Ok(report) => report,
+        Err(error) => {
+            warn!(%error, path = path.as_str(), "failed to read run wrapper report");
+            None
+        }
+    }
```
sunshowers, line 63: This should cause the test to fail.

nextest-runner/src/runner/wrapper_report.rs
```diff
@@ -0,0 +1,191 @@
...
+    )]
+    InvalidGroup,
+}
+
+pub(super) fn new_report_path() -> Result<Utf8TempPath, ChildStartError> {
+    let path = camino_tempfile::Builder::new()
+        .prefix("nextest-run-wrapper-report")
+        .tempfile()
+        .map_err(|error| ChildStartError::TempPath(Arc::new(error)))?
+        .into_temp_path();
```
sunshowers, line 47: Instead of this, I think we should create a temp dir for each test with a `report.json` inside it.

nextest-runner/src/reporter/events.rs
```diff
@@ -1283,6 +1283,21 @@ pub struct OutputErrorSlice {
...
+/// report means the wrapper ran the test normally.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(test, derive(test_strategy::Arbitrary))]
+pub struct RunWrapperReport {
+    /// A short label displayed on the per-test status line.
+    pub label: String,
+    /// An optional category used to aggregate counts in the final run summary.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub group: Option<String>,
```
sunshowers, line 1298: 1. How does this group interact with test groups? We need to find a different name for this.
2. We should have `RunWrapperLabel` and `RunWrapperGroup` newtypes which parse-don't-validate.

nextest-runner/src/test_command.rs
```diff
@@ -65,7 +67,7 @@ impl TestCommand {
         lctx: &LocalExecuteContext<'_>,
         program: String,
         args: &[Cow<'_, str>],
-        env: Option<&ScriptCommandEnvMap>,
+        wrapper_env: Option<&ScriptCommandEnvMap>,
```
sunshowers, line 70: Can you pass in an `Option<&WrapperScriptConfig>` here?

nextest-runner/src/reporter/displayer/progress.rs
```diff
@@ -657,6 +708,15 @@ pub(super) fn write_summary_str(run_stats: &RunStats, styles: &Styles, out: &mut
...
         }
+        if let Some(groups) = passed_groups {
+            for (group, count) in groups {
+                text.push(format!(
+                    "{} {}",
+                    count.style(styles.count),
+                    group.style(styles.skip),
+                ));
+            }
+        }
```
sunshowers, line 719: Some other review notes from Claude:

<details>

<summary>Notes</summary>

- Distinct group cardinality is unbounded; a wrapper that derives the group from the test name produces a summary line thousands of characters long. Cap and fold into "other".
- Group names collide with nextest's own words in the same parenthetical: group = "flaky" yields 12 passed (3 flaky, 5 flaky). The unit test itself uses "flaky" as a group name.
- read_report uses blocking std::fs inside the async executor; parse_env_file (runner/script_helpers.rs:13-56) uses tokio::fs.
- Stress mode: groups print per sub-run and are cleared, but the final RunFinishedStats::Stress summary (imp.rs:1422) omits them. clear() is caller-driven; take() would be structural.
- write_summary_str's Option<&RunWrapperGroupCounts>: None and Some(&default()) render identically, so it encodes provenance, not behaviour. Fine, but document it.
- TempPath error variants use both {0} and #[source], duplicating the cause when chains are printed; the rest of errors.rs uses #[source] alone.
- A valid 256-char label with \uXXXX escapes can exceed 1024 bytes and be rejected as "too large".

</details>

(in the future please do a round of automated review if you don't already)

nextest-runner/src/reporter/events.rs
```diff
@@ -2809,6 +2829,58 @@ mod tests {
         }
     }
 
+    #[test]
```
sunshowers, line 2832: We need to have an end-to-end test for this.

sunshowers (changes requested):

Issue #1469, opened 2024-04-26
[windows] lots of tests running in parallel may cause "leaky" warnings


zaneduffield:
When I run my tests with `datatest-stable` (0.2.9), using `nextest` as the runner, a semi-random number of tests are marked as leaky.

This seems to only depend on how many tests the crate has, and not what the tests are doing.

I have an example repo here: https://github.com/zaneduffield/datatest_leak_issue
That repo has 1000 test files, which are used with a no-op test function.
When I run `cargo nextest run` on it, about 40% of the tests are marked as 'leaky'.
One issue that _may_ be related is how slow this is: ~17s on my machine compared to ~0.2s with `cargo test`.

asomers:
I don't think this is a Windows problem.  I'm seeing it when I run Mockall's test suite (https://github.com/asomers/mockall) with cargo-nextest on FreeBSD 15.0-CURRENT amd64, on a system with 16 hyperthreading cores.  Notably, mockall's test suite does not contain _any_ tests that spawn processes.  I should mention that Mockall's test suite contains 171 binaries.

But if I add `-j1` to the `cargo nextest run` command line, then there are no failures.  In fact, the failure probability is closely related to the value of the `-j` option.  Roughly:
| -j value  | time to test failure |
| ----------- | --------------------------- |
| 4            | never?                       |
| 8            | a few minutes          |
| 16          | a few dozen runs    |
| 32          | 3-5 runs                    |
| 171        | instant                      |

sunshowers:
@asomers interesting. I was wondering whether it was related to not being able to set `CLOEXEC` atomically on pipes, but that's supported on FreeBSD: https://github.com/rust-lang/rust/blob/779e19d8baa3e3625bd4fc5c85cbb2ad47b43155/library/std/src/sys/pal/unix/pipe.rs#L22-L33

asomers:
Correct @sunshowers .  FreeBSD has supported O_CLOEXEC ever since pipe2 was first added to FreeBSD 10.0, which predates Rust itself.

sunshowers:
Thanks. The reason I suspected CLOEXEC was that we used to see leaks on illumos, but those leaks were completely addressed once we added CLOEXEC support to the Rust standard library on illumos (which makes sense, the issue was pipes being leaked in that little window between creation and CLOEXEC being set).

[...]

PR #3553, opened 2026-08-27
🐛 fix(runner): stop concurrent spawns inheriting capture pipes

gaborbernat:
On Unix targets without `pipe2()`, `std` creates a pipe and sets `FD_CLOEXEC` in a second call, so a process spawned from another thread in between inherits both ends ([rust-lang/rust#95584](https://github.com/rust-lang/rust/issues/95584)). Nextest spawns tests and list commands from several threads, so a test can inherit a sibling's capture writer, keep it open after the sibling exits, and get that sibling reported as leaked. I hit this on macOS in [tox-dev/peryx#1629](https://github.com/tox-dev/peryx/issues/1629); sunshowers suspected the same race in [#1469](https://github.com/nextest-rs/nextest/issues/1469).

The fix is a process-wide `RwLock` on the Unix targets outside std's `pipe2` list in [`sys/pipe/unix.rs`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pipe/unix.rs). `create_pipe` takes the write lock around `std::io::pipe()` and `spawn_process` takes the read lock around the spawn, so spawns run in parallel and wait for pipe creation alone. Nextest creates the capture pipes itself instead of using `Stdio::piped()`, which `std` would create inside its spawn. Windows needs no lock because `std` serializes `CreateProcess` there; its split pipes become `CreatePipe` handles, as the combined path has used since [#2665](https://github.com/nextest-rs/nextest/pull/2665), with the default buffer instead of std's 64 KiB.

clap, `cargo nextest run -E 'not test(ui_tests) and not test(example_tests)'`, release builds on a 10-core M-series, mean of alternating runs against main:

| spawn lock | main | branch | change | runs |
|---|---|---|---|---|
| mutex, first version | 1.68s | 2.04s | +21% | 5 |
| read lock, this PR | 1.24s | 1.24s | 0% | 6 |

`std`'s fork-and-exec fallback creates an exec-error pipe of its own inside the spawn; leaked into a concurrent child under a read lock, it would stall that spawn and every pipe creation behind it. Spawns that can fall back take the write lock instead: on Apple a non-absolute program, and on the other locked targets any spawn. Test binaries and the double-spawn stub are absolute, so test spawns keep the read lock.


sunshowers:
How is this specific to macOS? Seems like any platform without atomic CLOEXEC would be hurt by this. I would grab the platform list from the Rust standard library here.

Have you profiled it to figure out how much slower this gets? I would recommend testing against the clap repo by running all tests except ui_tests and example_tests (iirc).

sunshowers:
The specific command I use to test performance with is, in the clap repo:

```
cargo nextest run -E 'not test(ui_tests) and not test(example_tests)'
```

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +348,266 @@ impl ChildOutputMut {
         }
     }
 }
+
+#[cfg(all(test, target_os = "macos"))]
+mod tests {
```
sunshowers: This is an astonishingly complex pair of tests. What is this trying to do? Why is one of these ignored? Could you explain it in your own words (not LLM output)?

gaborbernat: Yeah, I agree. The goal here is to try to simulate how nextest behaves under load, and that each element holds its own writer... but can and will do better, sorry for the noise.

nextest-runner/src/test_command.rs
```diff
@@ -242,22 +245,72 @@ impl TestCommand {
...
+/// Capture pipes must come from [`create_pipe`] rather than `Stdio::piped()`,
+/// which would create them inside the standard library's spawn without the
+/// write lock.
+///
+/// The standard library creates one more pipe of its own when it falls back
+/// from `posix_spawn` to fork and exec: for example, a bare program name with
+/// `PATH` overridden in the environment, or a relative program path combined
+/// with a working directory. That pipe is not covered here. If a concurrent
+/// child inherits its write end, the spawn blocks until that child exits, but
+/// the capture pipes stay unaffected.
```
sunshowers: (FWIW we are very careful about always using posix_spawn as much as possible -- I've contributed multiple fixes to the Rust standard library around this.)

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +238,64 @@ impl TestCommand {
...
+/// is a spawn that blocks until the inheriting child exits, and the capture
+/// pipes stay untouched.
+pub(crate) fn spawn_process<T>(spawn: impl FnOnce() -> std::io::Result<T>) -> std::io::Result<T> {
+    let _guard = SPAWN_INHERITS_PIPES.then(|| {
+        PROCESS_SPAWN_LOCK
+            .read()
+            .unwrap_or_else(PoisonError::into_inner)
+    });
+    spawn()
+}
```
sunshowers, line 276: So... if we ever do fall back to fork/exec, due to the lack of atomic CLOEXEC, then there's a window in which the fork-and-exec pipe gets created but CLOEXEC isn't set on it. In that window, if the pipe gets leaked to a different test, then that spawn will end up being stalled until that test has completed execution. This is really bad.

Specifically, consider the situation in which:

1. Thread A spawns test T1, taking a read lock, then going down the fork/exec branch (our assumption). std creates its exec-error pipe, P1, as part of the fork/exec protocol.
2. Thread B spawns test T2. CLOEXEC hasn't been set on the pipe P1 yet. So P1 is leaked into T2.
3. Thread A waits for the last writer for P1 to close (within std), while still holding a read lock.
4. But the last writer for P1 won't be closed until T2 exits.

This effectively means that _all_ `create_pipe` calls will block until T2 exits. That would destroy nextest's performance.

Now... under which circumstances can we fall back to `posix_spawn`? The conditions that can happen in are:

1. The platform's `posix_spawn` doesn't have a way to set the current directory.
2. On macOS, the path to the binary is not absolute. (This is a fix I wrote to the standard library to work around a macOS libc bug.)

macOS 10.15 and above do have a way to set the current directory, so 1 is not an issue on those platforms. When can 2 happen? This is actually interesting:

* We always execute test binaries as absolute paths.
* But, nextest supports wrapper scripts, which can have relative paths.
* But also, nextest uses double-spawning by default (where it spawns a copy of itself first), which is always invoked as an absolute path.

So for the fork-exec fallback to be hit, double-spawning must be disabled and a wrapper script must be used. This is a relatively rare combination (double-spawning is not a user-controllable config knob) but a real one.

---

How do we fix this? I think the simplest thing would be that if all the following conditions are true:

1. We're on a platform without atomic CLOEXEC.
2. Any of the following conditions are true:
    1. The platform doesn't have posix_spawn.
    2. The platform has posix_spawn but is not Apple.
    3. All of the following conditions are true:
        1. The platform is Apple (which has `posix_spawn`).
        2. The `cmd` program is not absolute (which is a simpler form of the double-spawn/wrapper script check).

then take a write lock.

Or, even simpler:

```
SPAWN_INHERITS_PIPES && (!cfg!(target_vendor = "apple") || !Path::new(cmd.get_program()).is_absolute())
    → write lock
SPAWN_INHERITS_PIPES (i.e. Apple with an absolute program)
    → read lock
otherwise
    → no lock
```

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +238,64 @@ impl TestCommand {
...
+        cmd.stderr(tx);
+        Some(rx)
+    } else {
+        None
+    };
+
+    let mut child = spawn_process(|| tokio::process::Command::from(cmd).spawn())?;
+    child.stdout = stdout_rx.map(pipe_reader_to_child_stdout).transpose()?;
+    child.stderr = stderr_rx.map(pipe_reader_to_child_stderr).transpose()?;
+    Ok(child)
```
sunshowers: If these fail for some reason, then the child gets leaked forever. I would call `start_kill` on the child in case of an error (and refactor this code slightly to make that easier to read).

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +238,64 @@ impl TestCommand {
     }
 }
 
+/// `std::io::pipe()` rather than Tokio's pipes: the standard library tracks
+/// atomic `O_CLOEXEC` per target (mio-pipe 0.1.1 lacks it on illumos), and
+/// Tokio has no anonymous pipes on Windows.
+pub(crate) fn create_pipe() -> std::io::Result<(PipeReader, PipeWriter)> {
```
sunshowers: This doesn't need to be `pub(crate)`.

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +238,64 @@ impl TestCommand {
...
+/// those pipes inside its spawn, outside the write lock.
+///
+/// The standard library's fork-and-exec fallback creates one more pipe that
+/// this lock does not cover. Nextest does not reach it by default:
+/// [`create_command`] spawns the absolute current executable, so `posix_spawn`
+/// applies, and interceptors spawn one process at a time. It stays reachable
+/// without double-spawn and with a relative or bare wrapper program; the cost
+/// is a spawn that blocks until the inheriting child exits, and the capture
+/// pipes stay untouched.
+pub(crate) fn spawn_process<T>(spawn: impl FnOnce() -> std::io::Result<T>) -> std::io::Result<T> {
```
sunshowers: This doesn't need to be `pub(crate)` either.

nextest-runner/src/test_command/unix.rs
```diff
@@ -1,8 +1,17 @@
...
 
-use std::{fs::File, io::PipeReader, os::fd::OwnedFd};
+use std::{fs::File, io, io::PipeReader, os::fd::OwnedFd, process};
+use tokio::process::{ChildStderr, ChildStdout};
 
 pub(super) fn pipe_reader_to_file(rx: PipeReader) -> File {
     File::from(OwnedFd::from(rx))
 }
+
+pub(crate) fn pipe_reader_to_child_stdout(rx: PipeReader) -> io::Result<ChildStdout> {
```
sunshowers: This can be `pub(super)`

nextest-runner/src/test_command/unix.rs
```diff
@@ -1,8 +1,17 @@
...
+    ChildStdout::from_std(process::ChildStdout::from(OwnedFd::from(rx)))
+}
+
+pub(crate) fn pipe_reader_to_child_stderr(rx: PipeReader) -> io::Result<ChildStderr> {
```
sunshowers: This can be `pub(super)`

nextest-runner/src/test_command/windows.rs
```diff
@@ -1,8 +1,17 @@
...
     File::from(OwnedHandle::from(rx))
 }
+
+pub(crate) fn pipe_reader_to_child_stdout(rx: PipeReader) -> io::Result<ChildStdout> {
+    ChildStdout::from_std(process::ChildStdout::from(OwnedHandle::from(rx)))
+}
+
+pub(crate) fn pipe_reader_to_child_stderr(rx: PipeReader) -> io::Result<ChildStderr> {
+    ChildStderr::from_std(process::ChildStderr::from(OwnedHandle::from(rx)))
+}
```
sunshowers: `pub(super)`

sunshowers (commented):
Thanks. Overall looks good but needs a few fixes.

nextest-runner/src/test_command.rs
gaborbernat: Ah, you're right, nice spot! Adjusted to take thisInto acocunt.

gaborbernat:
@sunshowers is this ready to go, or is there anything else outstanding?

sunshowers:
Mostly good, dealing with personal matters which take priority.

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,132 @@ impl ChildOutputMut {
...
+        println!("{CAPTURE_MARKER}");
+        io::stdout().flush().expect("stdout flushes");
+        // SAFETY: `close` has no memory-safety preconditions, and nothing
+        // writes to these descriptors afterwards.
+        unsafe {
+            libc::close(libc::STDOUT_FILENO);
+            libc::close(libc::STDERR_FILENO);
+        }
+        thread::sleep(CHILD_LIFETIME);
+    }
```
sunshowers, line 472: I would prefer something like:

```rust
    fn spawn_lingering_child(
        strategy: CaptureStrategy,
        program: ChildProgram,
    ) -> io::Result<Child> {
        let mut command = match program {
            ChildProgram::Absolute => Command::new("/bin/sh"),
            #[cfg(target_vendor = "apple")]
            ChildProgram::RelativeWithCwd => {
                let mut command = Command::new("./sh");
                command.current_dir("/bin");
                command
            }
        };
        // * `exec >&-` closes the descriptors in the shell.
        // * `exec sleep` replaces the shell.
        command.arg("-c").arg(format!(
            "echo {CAPTURE_MARKER}; exec >&- 2>&-; exec sleep {CHILD_LIFETIME_SECS}"
        ));
        spawn(command, strategy, false)
    }
```

over an ignored test.

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +238,72 @@ impl TestCommand {
...
+}
+
+/// Callers must not use `Stdio::piped()`: the standard library would create
+/// those pipes inside its spawn, outside the write lock.
+///
+/// The read lock is enough only where `posix_spawn` is guaranteed, on Apple
+/// with an absolute program. Elsewhere a fork-and-exec fallback creates an
+/// exec-error pipe inside the spawn; leaked into a concurrent child it would
+/// stall this spawn and every `create_pipe` behind it, so those spawns take
+/// the write lock.
```
sunshowers, line 261: Please rewrite this comment in your own words.

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,132 @@ impl ChildOutputMut {
...
 }
+
+#[cfg(all(test, unix))]
+mod tests {
+    use super::*;
+    use std::{io::Write, path::Path, process::Command, sync::Barrier, thread, time::Duration};
+    use test_case::test_case;
+
+    const SPAWN_CONCURRENCY: usize = 32;
+    const SPAWN_ROUNDS: usize = 16;
```
sunshowers, line 353: How are these numbers picked?

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,132 @@ impl ChildOutputMut {
...
+    const CAPTURE_MARKER: &str = "NEXTEST_CAPTURE_CLOSED";
+    const EOF_TIMEOUT: Duration = Duration::from_secs(30);
+    const CHILD_LIFETIME: Duration = Duration::from_secs(120);
+
+    /// Children close their own capture pipes and linger, so a reader that
+    /// does not reach EOF proves a sibling inherited the writer.
```
sunshowers, line 359: Please rewrite this comment in your own words.

nextest-runner/src/test_command.rs
```diff
@@ -18,14 +18,41 @@ use std::{
...
 mod imp;
+use imp::attach_capture_readers;
 pub(crate) use imp::{Child, ChildAccumulator, ChildFds};
 
+/// Outside the listed targets the standard library has no `pipe2(O_CLOEXEC)`
+/// and sets `FD_CLOEXEC` after `pipe()`, so a child spawned from another
+/// thread in that window keeps the capture writer open and the test shows up
+/// as leaked (rust-lang/rust#95584). Windows serializes `CreateProcess` inside
+/// the standard library. Keep the list in sync with
+/// `library/std/src/sys/pipe/unix.rs`.
```
sunshowers, line 36: Please rewrite this in your own words as well. I think it's worth talking about atomic CLOEXEC here and the consequences of not having it.

grishy:
Adding another data point from our Rust workspace on macOS 26.6.2 (arm64), using nextest 0.9.129 (just for context, as you said non-macos special). I  saw intermittent LEAK reports on synchronous configuration tests that don't spawn processes.

It occurred in one of three workspace repetitions and one of five API-package repetitions aporoximately.
An `lsof` snapshot reported two sibling test processes holding the same pipe endpoint:
```
PID    FD   PIPE
20492  46   0x64532149b05d75a7
20494   2   0x64532149b05d75a7
```
Both had nextest as their parent. One process held the other process's stderr pipe as an extra descriptor.

nextest-runner/src/test_command/imp.rs
gaborbernat: Used your shell command and `ChildProgram` variants, with relative cases gated to Apple. Removed the ignored helper and unsafe descriptor closes.

The relative cases exercise fork/exec but can still pass after a read-lock regression stalls `spawn()`. I left out an elapsed-time assertion because it would depend on runner load.

nextest-runner/src/test_command/imp.rs
gaborbernat: 32 threads oversubscribe both the 10-core development machine and the 3-core macOS CI runner. With the lock disabled, local measurements reached round 16 before detecting an inherited pipe. Since rounds start at zero, the original 16-round budget would miss that sample. I raised it to 32 and added the rationale beside the constants.

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,144 @@ impl ChildOutputMut {
...
+    // the lock, one sample first failed at round 16 (zero-based), so use 32
+    // rounds to leave margin.
+    const SPAWN_CONCURRENCY: usize = 32;
+    const SPAWN_ROUNDS: usize = 32;
+    const CAPTURE_MARKER: &str = "NEXTEST_CAPTURE_CLOSED";
+    const EOF_TIMEOUT: Duration = Duration::from_secs(30);
+    const CHILD_LIFETIME_SECS: u64 = 120;
+
+    /// Keep children alive after closing stdout and stderr so an inherited
+    /// writer in a sibling delays EOF.
```
sunshowers, line 362: I don't think this comment is in the right location.

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,144 @@ impl ChildOutputMut {
...
+
+    #[derive(Clone, Copy)]
+    enum ChildProgram {
+        Absolute,
+        /// The cwd forces Apple's fork/exec fallback for a relative program.
+        #[cfg(target_vendor = "apple")]
+        RelativeWithCwd,
+    }
+
+    /// Kill on panic so inherited writers cannot delay runtime shutdown.
```
sunshowers, line 417: I don't know how this explains `LingeringChildren`.

nextest-runner/src/test_command/imp.rs
```diff
@@ -351,3 +342,144 @@ impl ChildOutputMut {
...
+                runtime.block_on(assert_capture_closes(child_fds, round, index));
+            }
+            runtime.block_on(lingering.kill_all());
+        }
+    }
+
+    #[derive(Clone, Copy)]
+    enum ChildProgram {
+        Absolute,
+        /// The cwd forces Apple's fork/exec fallback for a relative program.
```
sunshowers, line 412: Should this be something like:

```
/// Configure the test to run as a program with a cwd and a relative path.
///
/// In this scenario, Apple platforms fall back to fork/exec.
```

nextest-runner/src/test_command.rs
```diff
@@ -214,6 +237,69 @@ impl TestCommand {
...
+    });
+    std::io::pipe()
+}
+
+/// Create capture pipes with `create_pipe`; `Stdio::piped()` bypasses its lock.
+///
+/// Fork/exec creates an exec-error pipe. A sibling that inherits its writer
+/// stalls the spawn and blocks pipe creation behind its lock. Take the write
+/// lock to prevent this; Apple spawns with absolute paths use `posix_spawn`
+/// and can run under a read lock.
```
sunshowers, line 257: I really don't understand what this is trying to say -- this is incredibly hard to follow.

This is very subtle code, and I would expect a comment here (really on `PROCESS_SPAWN_LOCK`) to be several hundred words (possibly 1000+ words) long and be written in a narrative style with several paragraphs. There is a lot to explain here. Things to cover include:

* The symptoms observed which cause us to have to introduce the spawn lock.
* The protocol we follow. 
* There are two mechanisms Rust uses on POSIX platforms: `posix_spawn` and fork/exec.
* When creating a command with captured output, the Rust standard library must create a pipe for each kind of output that's captured (for both posix_spawn and fork/exec). We do this pipe creation ourselves to control concurrency for the reasons listed below.
* But by default, pipes will end up leaking into child processes. All POSIX platforms support marking a pipe (or other fd) close-on-exec (CLOEXEC).
* But also, there is a window after a pipe is created but before CLOEXEC is set on it that it can leak into another test process.
* So many POSIX platforms support _atomic_ CLOEXEC.
* But some platforms don't, such as Apple ones. So we must do our own synchronization to ensure that no child processes are created in this window.
* With fork/exec, a pipe is also created to communicate status between the parent and the child process. That, too, is susceptible to this race. And the consequences are actually quite a bit worse with this since the parent waits until the write end of the pipe passed to the child is closed (why? because the child writes the errno to that pipe if exec fails, so the parent reads until EOF to learn exec succeeded), and a leaked writer stalls that process spawn until the test it is leaked to completes.
* Rust doesn't support introspection to determine if a particular command invocation will use posix_spawn or fork/exec, so we must simulate what the Rust standard library does here.
* This boils down to rules X Y Z (covered in the earlier review comment), as examined on Rust 1.98. (Also important to cover things like we try our best to use posix_spawn over fork/exec, e.g. not setting uid/gid/`pre_exec` which always forces fork/exec -- and if any of that changes we must revise the rules to account for that).
* We could use a mutex or always do a write lock to be simpler but that has a real measured perf loss. So we try and solve the harder version of the problem and determine where read locks are sufficient and where we have to use a write lock.
* Why this is not an issue on Windows.

Most importantly, you _must_ write this in your own words. You can use an LLM as an editor and proofreader, but I will not accept an LLM-written comment for something this careful and subtle. It doesn't have to be perfect technical English -- I'm happy to edit and polish the comment myself -- but it needs to be much more substantive than we have today.

This isn't required, but if you need tips on how to do technical writing well, I highly recommend [_Style: Lessons in Clarity and Grace_](https://en.wikipedia.org/wiki/Style:_Lessons_in_Clarity_and_Grace).

sunshowers:
Hey @gaborbernat, been thinking about this. I think I want to land this and then write a larger design document about process spawning myself, not as a comment but as a top-level document published to the site. What do you think of cleaning up the comments/shortening them and then making this ready? This seems like an important enough fix that I don't want to block on.

gaborbernat:
Do you have any guidance on how much to shorten?

sunshowers (approved):
Looks great as is -- thanks! When I write the design doc I might update the comments a bit.

gaborbernat:
Would appreciate if you could tag me when you write it for my awareness and information, thanks!

sunshowers:
Absolutely -- will do!

PR #3554, opened 2026-08-27
update repo nextest config to prevent long test timeout on MacOS

k-cross:
Additive repo profile configuration change that adds additional threads to prevent contention and test failures from timeouts. Failures no longer occur for me when running `cargo nextest run` locally.


sunshowers (approved):
Thanks!

